PR #42, opened 2018-05-09
performance - graph, scatterplot and brush select

bkmartinjr:
Performance work:
* improved performance on scatterplot and graph REGL buffer generation
* introduced optimized state storage for the graph (used to be graphMap, now graphVec a typed array)
* optimizations of brush selection
* optimization of linear scaling (replacement for `d3.scaleLinear()`)

Bug fix:
* small bug fix in scatterplot SVG axis generation code - there was a race condition not protected against which would cause a stack crawl periodically.

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,5 +1,6 @@
 // jshint esversion: 6
 import * as globals from "../globals";
+import store from "../reducers";
 import URI from "urijs";
 import _ from "lodash";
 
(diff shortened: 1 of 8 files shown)


src/actions/index.js
```diff
@@ -81,6 +82,30 @@ const initialize = () => {
...
+// (see https://github.com/chanzuckerberg/cellxgene-rest-api/issues/34) but
+// occasionally does.
+//
+function cleanupExpressionResponse(data) {
+  const s = store.getState();
+  const metadata = s.controls.currentCellSelectionMap;
+  let errorOccured = false;
+  const newcells = _.filter(data.data.cells, cell => {
+    const found = metadata[cell.cellname];
+    errorOccured = errorOccured || !found;
```
colinmegill: This is acceptable but reads slightly obfuscated to me.

src/util/scaleLinear.js
```diff
@@ -0,0 +1,18 @@
...
+//     var myScale = scaleLinear([0, 1], [-1, 1]);
+//     myScale(0) === -1
+// this is is equivalent to d3.scaleLinear().domain([0,1]).range([-1,1])
+
+export const scaleLinear = (domain, range) => {
+  const offsetD = domain[0];
+  const scale = (range[1] - range[0]) / (domain[1] - domain[0]);
+  const offsetR = range[0];
+  return function(v) {
+    return (v - offsetD) * scale + offsetR;
```
colinmegill: Acceptable, but generally, prefer fully written out and self commenting variable / param names here to make this read more easily. Even if names end up pretty long.

src/actions/index.js
bkmartinjr: happy to fix.   what is obfuscated?  Is it the inversion of the boolean?   (eg, switch to using `notFound`?)

bkmartinjr: That change would print a console error for every missing cell (seems excessive :-).

How about this:

```js
  let errorLogged = false;
  const newcells = _.filter(data.data.cells, cell => {
    if (!errorLogged && !metadata[cell.cellname]) {
      errorLogged = true;
      console.error(
        "Warning: /expression REST API returned unexpected cell names -- discarding surprises."
      );
      return !metadata[cell.cellname];
    }
  });
```

bkmartinjr: OK, just sent you a PR containing fixes.

PR #71, opened 2018-05-28
introduce new selection state machinery

bkmartinjr:
* added TypedCrossfilter
* refactored redux state to use it
* better state caching in graph component
* general cleanup of state


src/reducers/controls.js
```diff
@@ -1,16 +1,153 @@
 // jshint esversion: 6
 import _ from "lodash";
 import { parseRGB } from "../util/parseRGB";
+var crossfilter = require("../util/typedCrossfilter");
+
+// In the case where the REST server does not implement data schema
+// declaration, we attempt to deduce it by sniffing the data.
+//
+function createSchemaByDataSniffing(ranges) {
```
colinmegill: would move to util / import

PR #80, opened 2018-06-05
render scatterplot on mount, handle switch tabs case

### LINKED ISSUE #62: diff expression scatter plot disappears
To reproduce:
1. select two cell sets and compute differential expression
2. click on Expression tab, and select an X and Y gene.
   -- you should now see diffexp scatterplot --
3. Click Metadata tab
4. Click Expression tab

At this point, the expression tab shows up, missing the scatterplot.   It just doesn't get re-rendered.

colinmegill:
Fixes #62 

screencaptureproject4

### DIFF
--- a/src/components/scatterplot/scatterplot.js
+++ b/src/components/scatterplot/scatterplot.js
@@ -51,6 +51,7 @@ class Scatterplot extends React.Component {
   constructor(props) {
     super(props);
     this.count = 0;
+    this.axes = false;
     this.state = {
       svg: null,
       // ctx: null,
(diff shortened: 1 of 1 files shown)


colinmegill:
On this branch, 

* try switching to the dataset to, ` prefix: "http://pbmc3k.cxg.czi.technology/api/",`
* compute differential expression
* go back and forth between tabs

This demonstrates the bugfix, we now persist the scatter between switching tabs.

colinmegill:
Whereas... another typical path taken: 

* Tabula Muris dataset
* regraph to marrow
* neutrophil vs neutrophil near terminus shows Ngp active along gradient

screencaptureproject5

results in the following error: 

image

PR #82, opened 2018-06-09
fixed brush behavior when zooming + panning

freeman-lab:
Previously, brushing did not select the expected cells when zooming or panning. This fixes that by incorporating the inverse of the camera distance into the conversion from screen coordinates to cell coordinates. See the animation below for the correct behavior provided by this PR.

zoom

### DIFF
--- a/src/components/graph/graph.js
+++ b/src/components/graph/graph.js
@@ -204,7 +204,7 @@ class Graph extends React.Component {
         2 *
           (1 - pin[1] / (this.props.responsive.height - this.graphPaddingTop)) -
         1;
-      const pout = [x + inverse[12], y + inverse[13]];
+      const pout = [x * inverse[14] + inverse[12], y * inverse[14] + inverse[13]];
       return [(pout[0] + 1) / 2, (pout[1] + 1) / 2];
     };
 


bkmartinjr:
LGTM

colinmegill:
@freeman-lab Yay! This is looking great. 

One more edge case to handle before merging: the brush needs to be reset when zooming. I think it's easier for our purposes to procedurally deselect the brush (ie., nuke it using an action, we do this on click outside of brush bounds), rather than trying to keep it in the right place. You do this manually in your gif by clicking off.
 
jun-12-2018 11-42-09

colinmegill:
Perhaps the place to do this is on click on the zoom button?

freeman-lab:
oops i missed this :) you're totally right, good catch, i agree, it should just nuke the selection do we have access to an event when you zoom?

colinmegill:
Fixed.

jun-28-2018 01-15-11

colinmegill:
For posterity, since this isn't all that intuitive, there are a couple pieces to this fix.

First, procedural brush control is necessary. This looks like:

`this.state.svg.select(".graph_brush").call(this.state.brush.move, null);`

Triggered by the onClick handler for activating zoom. But there's a catch. Procedurally moving the brush *still emits an event*. This means the following is also necessary to prevent the brush code from triggering its usual 'user just tried to select some cells' flow with a bunch of null values:

`if (d3.event.sourceEvent !== null) {`

PR #113, opened 2018-07-11
WebGL incremental render to fix battery drain

### LINKED ISSUE #19: excessive GPU utilization
Current webgl code runs the GPU constantly if the application is an active tab.   This will eat laptop batteries for lunch and generally thrashes my laptop.

We should look at ways to only render when necessary.

colinmegill:
Fixes #19


freeman-lab:
confirmed on my box that this seems to be working right, in "zoom" mode battery drain is significant, goes down when switching to draw mode, goes up again while actually drawing!

freeman-lab:
also, reviewed all changes, all looks right and in line with what we discussed, so LGTM!

csweaver:
Merged

PR #116, opened 2018-07-16
Csweaver/packaging

csweaver:


csweaver:
This is actually PR #113. Github was buggy at the time of merge so the changes didn't go through. I made a new pr with the latest from the branch.

PR #117, opened 2018-07-16
Csweaver/tests

csweaver:
Added some tests for the scanpy engine. This requires that you must install the package to run to the tests.

### DIFF
--- a/server/test/test_filter.py
+++ b/server/test/test_filter.py
@@ -1,9 +1,7 @@
 import unittest
-
 from unittest.mock import MagicMock
-import sys
-sys.path.insert(0, "../app")
-from util.filter import _convert_variable, parse_filter
+
+from server.app.util.filter import _convert_variable, parse_filter
 
 
 class UtilTest(unittest.TestCase):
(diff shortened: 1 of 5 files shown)


server/test/test_filter.py
```diff
@@ -72,3 +70,6 @@ def test_filter_contiunous(self):
         filterMock.getlist.return_value = ["0,*"]
         query = parse_filter(filterMock, self.schema)
         assert query == {"n_genes": {"variable_type": "continuous", "value_type": "int", "query": {"min": 0, "max": None}}}
+
+if __name__ == '__main__':
+    unittest.main()
```
ttung: NL after this.

server/test/test_scanpy_engine.py
```diff
@@ -0,0 +1,68 @@
...
+
+from server.app.scanpy_engine.scanpy_engine import ScanpyEngine
+
+
+class UtilTest(unittest.TestCase):
+    def setUp(self):
+        self.data = ScanpyEngine("example-dataset/", schema="data_schema.json")
+
+    def test_init(self):
+        assert self.data.cell_count == 2638
```
ttung: is there a reason this is used instead of `self.assertEqual`?  i think assertEqual will produce nicer messages when they fail (i.e., it'll print out the two values).

csweaver: no reason ¯\_(ツ)_/¯, I'll make that change

PR #119, opened 2018-07-17
Setup travis integration

csweaver:


.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - "8"
+cache:
+  pip: true
+addons:
+  apt:
+    packages:
+    - httpie
+install:
+  - set -eo pipefail
+  - pip install flake8
```
ttung: you could also do pip install httpie here instead of doing it as part of apt.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - ./bin/build-client
+  - python setup.py install
+script:
+  - set -eo pipefail
+  - flake8 server/app/
+  - pytest -s server/test/test_filter.py server/test/test_scanpy_engine.py
+  - cellxgene &
+  - for i in {1..90}; do if http :5005/api/v0.1/initialize > /dev/null; then break; else echo "Waiting for server..."; sleep 1; fi; done
+  - pytest server/test/test_api.py
```
ttung: trailing NL.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 120
```
ttung: NL.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+  - set -eo pipefail
+  - pip install flake8
+  - ./bin/build-client
+  - python setup.py install
+script:
+  - set -eo pipefail
+  - flake8 server/app/
+  - pytest -s server/test/test_filter.py server/test/test_scanpy_engine.py
+  - cellxgene &
+  - for i in {1..90}; do if http :5005/api/v0.1/initialize > /dev/null; then break; else echo "Waiting for server..."; sleep 1; fi; done
```
ttung: You may wish to make this time-based, as that yields a bit more certainty about how long you blow through this loop. :)

.travis.yml
csweaver: 🤦‍♀️ Ugh, what is wrong with me. I'll learn eventually...

.travis.yml
csweaver: that's cleaner, I like that.

.travis.yml
csweaver: Still a loop? Or like sleep 10?

ttung: Oops this should be fine.  I misread this.

PR #121, opened 2018-07-20
Csweaver/cli

csweaver:
Stage 1 of CLI implemented.

Now you can run with:
cellxgene run [engine] [dataset directory]

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -16,23 +16,11 @@ Compress(app)
 CORS(app)
 
 # Config
-CXG_DIR = os.environ.get("CXG_DIRECTORY", default="example-dataset/")
 SECRET_KEY = os.environ.get("CXG_SECRET_KEY", default="SparkleAndShine")
-ENGINE = os.environ.get("CXG_ENGINE", default="scanpy")
-TITLE = os.environ.get("DATASET_TITLE", default="PBMC 3K")
-# TODO remove the 2 when this is prod
-CXG_API_BASE = os.environ.get("CXG_API_BASE2", default="http://0.0.0.0:5005/api/")
 
 app.config.update(
     SECRET_KEY=SECRET_KEY,
-    CXG_API_BASE=CXG_API_BASE,
-    ENGINE=ENGINE,
-    DATA=CXG_DIR,
-    DATASET_TITLE=TITLE
 )
-
-app.config["PROFILE"] = True
-# app.wsgi_app = ProfilerMiddleware(app.wsgi_app, restrictions=[15])
 
 # Application Data
 data = None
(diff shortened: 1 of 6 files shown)


server/app/app.py
```diff
@@ -1,4 +1,5 @@
 import os
+import argparse
```
ttung: keep alphabetized?

server/app/app.py
```diff
@@ -53,5 +40,33 @@
 app.add_url_rule("/", endpoint="index")
 
 
+def run(args):
+    global data
```
ttung: generally not a fan of globals.

consider attaching the data object to the app object (possibly in a separate PR).

consider also instantiating the app object here (i.e., in `run()`), and converting any references to `server.app.app.app` to [flask.current_app](http://flask.pocoo.org/docs/1.0/api/#flask.current_app) in a separate PR.

server/app/app.py
```diff
@@ -53,5 +40,33 @@
...
+    run_subparser.add_argument("--title", "-t", help="Title to display (default = data directory name")
+    run_subparser.add_argument("--port", help="Port to run server on.", type=int, default=5005)
+    run_subparser.add_argument("engine", help="The underlying structure of the data")
+    run_subparser.add_argument("data_directory", metavar="dir", help="Directory containing data and schema")
+    run_subparser.set_defaults(func=run)
+    args = parser.parse_args()
+
+    args.func(args)
+
+    # app.run(host="0.0.0.0", debug=True, port=5005)
```
ttung: remove.

server/app/app.py
```diff
@@ -53,5 +40,33 @@
...
+        data = ScanpyEngine(args.data_directory, schema="data_schema.json")
+    app.run(host="0.0.0.0", debug=True, port=args.port)
+
+
 def main():
-    app.run(host="0.0.0.0", debug=True, port=5005)
+    parser = argparse.ArgumentParser(description="Cellxgene is a tool for exploring single cell expression.")
+    subparsers = parser.add_subparsers(dest="cellxgene_command")
+    run_subparser = subparsers.add_parser("run", help="run cellxgene")
+    run_subparser.add_argument("--title", "-t", help="Title to display (default = data directory name")
```
ttung: I had to dig into the code to figure out what this meant.  Consider clarifying.

server/app/app.py
```diff
@@ -53,5 +40,33 @@
...
+    run_subparser.add_argument("--port", help="Port to run server on.", type=int, default=5005)
+    run_subparser.add_argument("engine", help="The underlying structure of the data")
```
ttung: Is this right?

server/app/app.py
csweaver: sure, I'll fix

server/app/app.py
csweaver: I didn't like it either, I prefer your idea. I wasn't quite sure what the best alternative to be and wanted to keep this PR relatively small. I like your idea, let's discuss later today.

server/app/app.py
csweaver: ok

server/app/app.py
csweaver: Which part? What the run command means?

server/app/app.py
csweaver: Is which right?

server/app/app.py
ttung: Where the title is shown.

csweaver: NP, I'll write some better help text

server/app/app.py
ttung: I thought engine was EM2 vs etc.

csweaver: yeah, I can write a better help text for that. I'm not quite sure how to describe it best in the help text.

ttung: If you make it a [choice](https://docs.python.org/3/library/argparse.html#choices), it might be pretty self-describing.

csweaver: Shifted from `cellxgene run <engine>` to `cellxgene <engine>` I think that make this a little clearer.

ttung:
This now has the changes from https://github.com/chanzuckerberg/cellxgene/pull/138.  i think you may need to rebase.

PR #142, opened 2018-08-03
Csweaver/cli subparsers

csweaver:


client/package.json
```diff
@@ -4,10 +4,8 @@
   "license": "MIT",
   "repository": "https://github.com/chanzuckerberg/cellxgene",
   "scripts": {
-    "build":
```
ttung: this looks like a ws change. does it need to be here?

server/app/app.py
```diff
@@ -64,8 +65,11 @@ def main():
                                               "of the directory from the data_directory arg")
     parser.add_argument("--port", help="Port to run server on.", type=int, default=5005)
     subparsers = parser.add_subparsers(dest="cellxgene_command")
-    scanpy_subparser = subparsers.add_parser("scanpy", help="run cellxgene using the scanpy engine")
-    scanpy_subparser.add_argument("data_directory", metavar="dir", help="Directory containing data and schema file")
-    scanpy_subparser.set_defaults(func=run_scanpy)
+    try:
+        from .scanpy_engine.scanpy_engine import ScanpyEngine
+        ScanpyEngine.add_to_parser(subparsers, run_scanpy)
```
ttung: would move this out of the try-except block.

server/app/app.py
```diff
@@ -64,8 +65,11 @@ def main():
...
+    except ImportError:
+        print("Scanpy engine not available")
```
ttung: https://docs.python.org/2/library/warnings.html ?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -23,6 +23,18 @@ def __init__(self, data, schema=None, graph_method="umap", diffexp_method="ttest
     def _set_cell_names(self):
         self.data.obs["cell_name"] = list(self.data.obs.index)
 
+    @classmethod
+    def add_to_parser(cls, subparsers, invocation_function):
+        scanpy_group = subparsers.add_parser("scanpy", help="run cellxgene using the scanpy engine")
+        # TODO these choices should be generated from the actual available methods
```
ttung: 👍

csweaver:
Sorry about the lack of comments on PR, I fixed that for the other outstanding one.

client/package.json
csweaver: Slipped that in there because the build was breaking. I should be able to get rid of this if I rebase with master, I think Colin fixed the problem.

server/app/app.py
csweaver: I don't understand why? I wrapped those two lines for when we have EM2 and Seurat engines since only one engine needs to be installed for cellxgene to work. So if someone can't import the scanpy engine then those options shouldn't get added to the arg parser, right?

server/app/app.py
csweaver: Yes! that is what I should have used 🤦‍♀️

server/app/app.py
ttung: It should be:

```
try:
    from .scanpy_engine.scanpy_engine import ScanpyEngine
except ImportError:
    warnings.warn("Scanpy engine not available", ImportWarning)
else:
    ScanpyEngine.add_to_parser(subparsers, run_scanpy)
```

The else clause is only run if the code inside the try-block completes without an exception.

Overall, the rationale is to have the least amount of code inside a try-block as possible.

csweaver: Ahh, gotcha. I had misunderstood what you were asking.

PR #155, opened 2018-08-06
Use pip instead of python setup.py

csweaver:


csweaver:
@bkmartinjr this seems to solve the WSL installation problem too, can you confirm by trying to install using

pip install -e <path to cellxgene>

ttung:
By the way, what was the issue you were encountering with `python setup.py install`?

csweaver:
I re-ran it on the fresh branch to see what the error was and the install went smoothly. So looks like that issue solved itself? I can do some archeology to figure out what the actually error was if you are interested.

PR #157, opened 2018-08-07
Added instructions on how to load your own data in scanpy

csweaver:


csweaver:
@mckinsel and @sidneymbell Could you let me know if the instructions are clear enough? I'm not so worried about the config file as that's going to change with the new rest version, but everything else? I want to make sure I'm using the terminology scientists will understand.

sidneymbell:
This looks great!! Thanks so much for writing this up, @csweaver ! I pushed a couple of minor edits, mostly just elaborating on what you already had in a few small places.

Issue #140, opened 2018-08-02
CLI install script failed on Win10/WSL/Ubuntu - could not find HDF5 libs


bkmartinjr:
I followed the instructions in the README to install the local CLI.  It failed with:

```
/usr/bin/ld: cannot find -lhdf5
collect2: error: ld returned 1 exit status
.. ERROR:: Could not find a local HDF5 installation.
   You may need to explicitly state where your local HDF5 headers and
   library can be found by setting the ``HDF5_DIR`` environment
   variable or by using the ``--hdf5`` command-line option.
error: Setup script exited with 1
```

Config:
* Py 3.5.2
* Win10, WSL (Ubuntu 16)

full details

```
(cellxgene) root@woof:/mnt/c/Users/bruce/projects/cellxgene# python3 setup.py install
running install
running bdist_egg
running egg_info
creating cellxgene.egg-info
writing cellxgene.egg-info/PKG-INFO
writing requirements to cellxgene.egg-info/requires.txt
writing entry points to cellxgene.egg-info/entry_points.txt
writing top-level names to cellxgene.egg-info/top_level.txt
writing dependency_links to cellxgene.egg-info/dependency_links.txt
writing manifest file 'cellxgene.egg-info/SOURCES.txt'
reading manifest file 'cellxgene.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'cellxgene.egg-info/SOURCES.txt'
installing library code to build/bdist.linux-x86_64/egg
running install_lib
running build_py
creating build/lib
creating build/lib/server
copying server/run.py -> build/lib/server
copying server/__init__.py -> build/lib/server
creating build/lib/server/app
copying server/app/app.py -> build/lib/server/app
copying server/app/__init__.py -> build/lib/server/app
creating build/lib/server/app/driver
copying server/app/driver/driver.py -> build/lib/server/app/driver
copying server/app/driver/__init__.py -> build/lib/server/app/driver
creating build/lib/server/app/rest_api
copying server/app/rest_api/rest.py -> build/lib/server/app/rest_api
copying server/app/rest_api/__init__.py -> build/lib/server/app/rest_api
creating build/lib/server/app/scanpy_engine
copying server/app/scanpy_engine/scanpy_engine.py -> build/lib/server/app/scanpy_engine
copying server/app/scanpy_engine/__init__.py -> build/lib/server/app/scanpy_engine
creating build/lib/server/app/util
[...]

PR #162, opened 2018-08-07
Trimming out secondary dependencies from requirements

csweaver:


csweaver:
@bkmartinjr I'm wondering if this might work to get install working on your computer. If we let scanpy manage it's own dependencies through pip we might not run into the error where it's missing tkinter

mckinsel:
I still get tkinter problems when I pip install scanpy on some platforms. It's actually a `matplotlib` problem. [This](https://github.com/HumanCellAtlas/table-testing/blob/master/tasks/merge/merge_anndata/Dockerfile#L7) is what I've been doing to get it to work, but it feels somewhat shameful. However you can't easily install tkinter in some environments.

bkmartinjr:
@csweaver - I was able to get install to work without this.   It just required the python-tk install you mentioned in #140 

That said, if we can remove unneeded dependencies, it seems like a win.

csweaver:
@bkmartinjr It is not exactly removing unnecessary dependencies. Most everything is still getting installed. I'm just letting the top level module handle their own dependencies.

Initially, when creating the requirements file I just did a pip freeze and copied that. With this PR, I've gone through and only included the dependencies that are actually imported (with the exception of anndata where I'm using an old version deliberately). That said, I do think this PR should land.

PR #166, opened 2018-08-08
Added clean script for cellxgene

csweaver:
This script allows you to start fresh when installing. Cleans up node modules, builds, and uninstalls cellxgene from python.

### DIFF
--- a/bin/clean
+++ b/bin/clean
@@ -1,0 +1,16 @@
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+CELLXGENE_DIR=$(dirname $DIR)
+
+echo "Uninstalling cellxgene"
+yes | pip uninstall cellxgene
+echo "removing node_modules"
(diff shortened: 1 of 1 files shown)


bin/clean
```diff
@@ -0,0 +1,16 @@
...
+echo "Uninstalling cellxgene"
+yes | pip uninstall cellxgene
+echo "removing node_modules"
+rm -rf $CELLXGENE_DIR/client/node_modules
+echo "removing client_build"
+rm -rf $CELLXGENE_DIR/client/build
+echo "removing egg-info"
+rm -rf $CELLXGENE_DIR/cellxgene.egg-info
+echo "removing static files"
+rm $CELLXGENE_DIR/server/app/web/templates/index.html
```
bkmartinjr: Do you need a -f (force) flag?

csweaver: I didn't include it since it was just a file and not a directory, but probably better to be safe. I'll add it.

PR #179, opened 2018-08-09
Format loaded dataset

csweaver:
Reformatting data to match the spec for the [rest v0.2 ](https://docs.google.com/document/d/1Fxjp1SKtCk7l8QP9-7KAjGXL0eldi_qEnNT0NmlGzXI)


csweaver:
@bkmartinjr yes, you're right. I think this PR needs more work, it's breaking the tests because we are indexing too. And I need to double check about the type conversion errors. I'm going to add in some tests for this and give up for now on having both in parallel.

bkmartinjr:
On the dataframe type coercion errors, I think for now it is fine to simply know what types you can support, and error out if you see anything else (a CLI error and exit - ie, "can't load, unsupported data types").    Is there any typical type in the scanpy data sets?  (ie, are they fp32 natively?)

Because the types are well defined, think we can do this from first principles (ie, don't lose range, OK to lose a bit of precision):
* OK:    int8/16/32/FP32 -> FP32
* not OK:   FP64/int64 -> FP32

The only one of these that is slightly questionable is int32->FP32, as that would lose precision.

@freeman-lab - any opinions on the first principles?   (background: currently, the front-end expects FP32 for expression "values", and the CLI will coerce data into this format.   this proposal is to simply error out if there is no reasonable coercion, where reasonable is defined as preserving *all* of dynamic range of the data, and *most* but not all of the precision.   Hence allowing int32->fp32, but not fp64->fp32)

csweaver:
The latest changes address Bruce's concerns and add tests.
 First I've separated the adding the name and index annotations from validating that the datatypes are correct. I'm also only warning about the matrix datatype. This should be in float32 by default due to scanpy specs, I warn if it's not in that format.

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::UserWarning
```
ttung: nit: nl

ttung: Also, I caution against blanket ignores like this.  https://docs.pytest.org/en/latest/warnings.html#pytest-mark-filterwarnings might work for what you're doing.

csweaver:
Fixed the quotes and limited the warnings to just the one we are interested in. It does mean that I'm importing both pytest and unittest. I think I need to move over to using pytest as the framework and not just the runner.

PR #182, opened 2018-08-09
Set static folder to correct location

csweaver:
Fixed issue where I wasn't serving static files from where they actually lived.


csweaver:
@ttung yes! I'll add a test in.

PR #186, opened 2018-08-13
Cleanup .gitignore

csweaver:
remove lines related to old separate repo version
add ignore for virtual environments and jupyter notebooks (I use them to test)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,52 +1,39 @@
 # dependencies
-node_modules
+node_modules/
 
 # coverage
 coverage
 .nyc_output
 
+.pytest_cache/
+
 # production
-build
+build/
+dist/
+*.egg-info
 
-# misc
-.DS_Store
-npm-debug.log
-.vscode
+# Environments
+venv/
+cellxgene/
 
-data
-
-
-*.idea*
-
-__pycache__
-*.DS_Store*
-
-# Elastic Beanstalk Files
-.elasticbeanstalk/*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,52 +1,39 @@
...
+.pytest_cache/
+
 # production
-build
+build/
+dist/
+*.egg-info
+
+# Environments
+venv/
```
ttung: I wouldn't put this in the .gitignore since it's very user-specific.  I would instead put this in your private .gitignore.

.gitignore
```diff
@@ -1,52 +1,39 @@
...
+# Jupyter Notebook
+.ipynb_checkpoints
+*.ipynb
 
 # misc
 .DS_Store
 npm-debug.log
 .vscode
-
-data
```
colinmegill: I still have a `data` folder - can I safely delete?

csweaver: Does it have EM2 files in it? If so yes. 

`data` was originally in there because that is where I was storing the various datasets built for EM2 in there.

.gitignore
csweaver: Haha! This is actually the whole reason for me improving the `.gitignore` file. Specifically, because @colinmegill didn't like that setting up the cellxgene venv from the quick start in the readme screwed up his git status. I'm happy to delete the venv/ though.

ttung: Would recommend setting up `.git/info/exclude` (yeah, the UX is awful).

PR #192, opened 2018-08-15
Csweaver/filter

csweaver:
Rewriting filtering to match the REST V0.2 spec: https://docs.google.com/document/d/1Fxjp1SKtCk7l8QP9-7KAjGXL0eldi_qEnNT0NmlGzXI/edit#heading=h.8qc9q57amldx

Adding filtering for both genes and cells (thus renamed filter_cells-> filter_dataframe)
I've spun off the functions for filter index and filter by annotations since they are doing different things at the scanpy level. 

Scanpy (really anndata) still has a few strange things that I had to work around
- the indexes can't be integers -> made them categories
- can't index by both var and obs at the same time -> used temp variable to index by one at a time.

### DIFF
--- a/server/app/driver/driver.py
+++ b/server/app/driver/driver.py
@@ -21,15 +21,13 @@ class CXGDriver(metaclass=ABCMeta):
     @abstractmethod
     def filter_dataframe(self, filter):
         """
-        Filter cells from data and return a subset of the data
-        A filter is a dictionary where the key is a metadatata category
-        Value is dictionary
-            value_type: int, float, string
-            variable_type: continuous, categorical
-            query: filter value, for categorical [val1, val2], for continuous {min: x, max:y}
-        Filters are combined with the and operator
-        :param filter:
-        :return: filtered dataframe
+         Filter cells from data and return a subset of the data. They can operate on both obs and var dimension with
+         indexing and filtering by annotation value. Filters are combined with the and operator.
+         See REST specs for info on filter format:
+         https://docs.google.com/document/d/1Fxjp1SKtCk7l8QP9-7KAjGXL0eldi_qEnNT0NmlGzXI/edit#heading=h.8qc9q57amldx
+
+        :param filter: dictionary with filter parames
+        :return: View into scanpy object with cells/genes filtered
         """
         pass
 
--- a/server/app/driver/driver.py
+++ b/server/app/driver/driver.py
@@ -19,7 +19,7 @@ class CXGDriver(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def filter_cells(self, filter):
+    def filter_dataframe(self, filter):
         """
         Filter cells from data and return a subset of the data
         A filter is a dictionary where the key is a metadatata category
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -118,7 +118,7 @@ class ScanpyEngine(CXGDriver):
         """
         d_axis = getattr(self.data, axis.value)
         for v in filter:
-            if d_axis[v["name"]].dtype.name in ["category", "string"]:
+            if d_axis[v["name"]].dtype.name in ["boolean", "category", "string"]:
                 key_idx = np.in1d(getattr(d_axis, v["name"]), v["values"])
                 index = np.logical_and(index, key_idx)
             else:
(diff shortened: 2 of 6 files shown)


server/app/driver/driver.py
```diff
@@ -19,17 +19,15 @@ def genes(self):
...
-            value_type: int, float, string
-            variable_type: continuous, categorical
-            query: filter value, for categorical [val1, val2], for continuous {min: x, max:y}
-        Filters are combined with the and operator
-        :param filter:
-        :return: filtered dataframe
+         Filter cells from data and return a subset of the data. They can operate on both obs and var dimension with
+         indexing and filtering by annotation value. Filters are combined with the and operator.
+         See REST specs for info on filter format:
+         https://docs.google.com/document/d/1Fxjp1SKtCk7l8QP9-7KAjGXL0eldi_qEnNT0NmlGzXI/edit#heading=h.8qc9q57amldx
```
ttung: I'm curious why this isn't just a pointer to the swagger spec you have in `server/app/rest_api/rest.py`

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -60,33 +60,75 @@ def genes(self):
...
+            if "index" in filter["obs"]:
+                cells_idx = self._filter_index(filter["obs"]["index"], cells_idx, "obs")
+            if "annotation_value" in filter["obs"]:
+                cells_idx = self._filter_annotation(filter["obs"]["annotation_value"], cells_idx, "obs")
+        if "var" in filter:
+            if "index" in filter["var"]:
+                genes_idx = self._filter_index(filter["var"]["index"], genes_idx, "var")
+            if "annotation_value" in filter["var"]:
+                genes_idx = self._filter_annotation(filter["var"]["annotation_value"], genes_idx, "var")
+        # Due to anndata issues we can't index into cells and genes at the same time
```
ttung: if that's the case, do you want to verify that people aren't trying to do that?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -60,33 +60,75 @@ def genes(self):
...
+         indexing and filtering by annotation value. Filters are combined with the and operator.
+         See REST specs for info on filter format:
+         https://docs.google.com/document/d/1Fxjp1SKtCk7l8QP9-7KAjGXL0eldi_qEnNT0NmlGzXI/edit#heading=h.8qc9q57amldx
+
+        :param filter: dictionary with filter parames
+        :return: View into scanpy object with cells/genes filtered
+        """
+        cells_idx = np.ones((self.cell_count,), dtype=bool)
+        genes_idx = np.ones((self.gene_count,), dtype=bool)
+        if "obs" in filter:
```
ttung: recommend making "obs" a constant or better yet an enum.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -60,33 +60,75 @@ def genes(self):
...
+                genes_idx = self._filter_annotation(filter["var"]["annotation_value"], genes_idx, "var")
+        # Due to anndata issues we can't index into cells and genes at the same time
+        data = self.data[cells_idx, :]
+        return data[:, genes_idx]
+
+    def _filter_index(self, filter, index, axis):
+        """
+        Filter data based on index. ex. [1, 3, [111:200]]
+        :param filter: subset of filter dict for obs/var:index
+        :param index: np logical vector containing true for passing false for failing filter
```
ttung: This is kind of a mouthful.  How about: np logical vector for whether the data passes the filter?

server/app/driver/driver.py
csweaver: This is a new spec. rest.py is obsolete at this point in the v2.0 refactor. I'll make a note to update that when it is up to date

server/app/scanpy_engine/scanpy_engine.py
csweaver: That comment refers to how I'm indexing into data in 2 lines instead of in one. A user can filter by both obs and by var but I have to index them one at a time.

server/app/scanpy_engine/scanpy_engine.py
csweaver: good idea. I'll add that in

server/app/scanpy_engine/scanpy_engine.py
csweaver: haha, yes thanks for translating that into english

csweaver:
@ttung can you do a quick double check and make sure I'm using enums in a sane way? Thanks!

server/app/scanpy_engine/scanpy_engine.py
ttung: You can use Axis.OBS.value in place of "obs" now.

Alternatively, you can use the AugmentedEnum hack we did https://github.com/spacetx/starfish/blob/master/starfish/constants.py which would allow you to just do `if AXIS.OBS in filter:`

PR #197, opened 2018-08-22
(EASY) Separate development requirements

csweaver:
add a separate file for python requirements for testing

### DIFF
--- a/server/requirements-dev.txt
+++ b/server/requirements-dev.txt
@@ -1,0 +1,2 @@
+pytest
+requests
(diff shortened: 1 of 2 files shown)


server/requirements-dev.txt
```diff
@@ -0,0 +1,2 @@
+pytest
+requests
```
shanaxel42: I think you might want:
 -r requirements.txt
at the end of this file as well so that travis installs all the requirements, not just these two.

ttung: `pip install -e .` will bring in the requirements of the package, but I think it's preferable to manage it with with `-r requirements.txt`, as it helps people who are just developing.

csweaver: thanks, added

PR #202, opened 2018-08-27
Csweaver/config

csweaver:
Add config route

> Configuration information to assist in front-end adaptation to underlying engine, available functionality, interactive time limits, etc.

I didn't put the feature detection directly in the engine because it depends on the CLI options selected instead of depending directly on the the methods availabe with the engine. Not 100% confident in that decision though.

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -52,22 +52,6 @@ def run_scanpy(args):
         DATASET_TITLE=title,
         CXG_API_BASE=api_base
     )
-    app.features = {
-        "cluster": {"available": False},
-        "layout": {
-            "obs": {"available": False},
-            "var": {"available": False},
-        },
-        "diffexp": {"available": False}
-    }
-    # TODO - Interactive limit should be generated from the actual available methods see GH issue #94
-    if "layout" in args and args.layout:
-        # TODO handle "var" when gene layout becomes available
-        app.features["layout"]["obs"] = {"available": True, "interactiveLimit": 15000}
-    if "diffexp" in args and args.diffexp:
-        app.features["diffexp"] = {"available": True, "interactiveLimit": 5000}
-    if "cluster" in args and args.cluster:
-        app.features["cluster"] = {"available": True, "interactiveLimit": 45000}
 
     from .scanpy_engine.scanpy_engine import ScanpyEngine
     app.data = ScanpyEngine(args.data_directory, graph_method=args.layout, diffexp_method=args.diffexp)
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -52,6 +52,22 @@ def run_scanpy(args):
         DATASET_TITLE=title,
         CXG_API_BASE=api_base
     )
+    app.features = {
+        "cluster": {"available": False},
+        "layout": {
+            "obs": {"available": False},
+            "var": {"available": False},
+        },
+        "diffexp": {"available": False}
+    }
+    # TODO - Interactive limit should be generated from the actual available methods see GH issue #94
+    if "layout" in args and args.layout:
+        # TODO handle "var" when gene layout becomes available
+        app.features["layout"]["obs"] = {"available": True, "interactiveLimit": 15000}
+    if "diffexp" in args and args.diffexp:
+        app.features["diffexp"] = {"available": True, "interactiveLimit": 5000}
+    if "cluster" in args and args.cluster:
+        app.features["cluster"] = {"available": True, "interactiveLimit": 45000}
 
     from .scanpy_engine.scanpy_engine import ScanpyEngine
     app.data = ScanpyEngine(args.data_directory, graph_method=args.layout, diffexp_method=args.diffexp)
--- a/server/app/driver/driver.py
+++ b/server/app/driver/driver.py
@@ -4,10 +4,13 @@ from abc import ABCMeta, abstractmethod
 class CXGDriver(metaclass=ABCMeta):
     def __init__(self, data, graph_method=None, diffexp_method=None):
         self.data = self._load_data(data)
-        self._set_features(graph_method=graph_method, diffexp_method=diffexp_method, cluster=None)
+        self.graph_method = graph_method
+        self.diffexp_method = diffexp_method
+        self.cluster = None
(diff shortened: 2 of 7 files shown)


server/app/app.py
```diff
@@ -52,6 +52,22 @@ def run_scanpy(args):
         DATASET_TITLE=title,
         CXG_API_BASE=api_base
     )
+    app.features = {
```
ttung: I think a `features()` method might be a good thing for `CXGDriver` classes to provide.

server/app/rest_api/rest.py
```diff
@@ -49,8 +49,60 @@ def get(self):
...
+        config = {
+            "config": {
+                "features": [
+                    {"method": "POST", "path": "/cluster/", **current_app.features["cluster"]},
+                    {"method": "POST", "path": "/layout/obs", **current_app.features["layout"]["obs"]},
+                    {"method": "POST", "path": "/layout/var", **current_app.features["layout"]["var"]},
+                    {"method": "POST", "path": "/diffexp/", **current_app.features["diffexp"]},
+                ],
+                "displayNames": {
+                    "engine": f"ScanPy Engine version {current_app.data.__version__}",
```
ttung: I think this is potentially confusing because is this the scanpy version or the scanpy cellxgene interface?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -13,6 +13,8 @@
 
 class ScanpyEngine(CXGDriver):
 
+    __version__ = "0.0.1"
```
ttung: Is there a compelling reason to version each individual `CXGDriver`?

server/app/rest_api/rest.py
csweaver: Do you mean I should change the verbiage? Or is the concept of versioning the engine itself confusing?

server/app/scanpy_engine/scanpy_engine.py
csweaver: Unfortunately, yes. If someone is running analysis through cellxgene then reproducibility is dependent on the version of the engine they used.

server/app/app.py
csweaver: ✅

server/app/rest_api/rest.py
ttung: Yeah.  cellxgene scanpy engine?  I dunno.  I'm terrible with naming things. :)

server/app/scanpy_engine/scanpy_engine.py
ttung: Shouldn't cellxgene itself just be versioned?

csweaver: Maybe? But the engines are so independent, someone using the (future) EM2 engine wouldn't care about updates to the scanpy engine. That was my reasoning as far as versioning at the scanpy engine level.

ttung: I guess, but it's two separate version numbers to keep track of now (for cellxgene and for the engines).

server/app/driver/driver.py
```diff
@@ -2,8 +2,27 @@
 
 
 class CXGDriver(metaclass=ABCMeta):
-    def __init__(self, data, schema=None, graph_method=None, diffexp_method=None):
+    def __init__(self, data, graph_method=None, diffexp_method=None):
         self.data = self._load_data(data)
+        self._set_features(graph_method=graph_method, diffexp_method=diffexp_method, cluster=None)
```
ttung: rather than do this, i would just make a `features` `@property` that calculates this and returns it.

rationale: keep your constructor lightweight, and this postpones the work until it's needed.

server/app/rest_api/rest.py
```diff
@@ -49,8 +49,60 @@ def get(self):
         return make_response(jsonify({"schema": current_app.data.schema}), 200)
 
 
+class ConfigAPI(Resource):
+    @swagger.doc({
+        "summary": "Configuration information to assist in front-end adaptation"
+                   " to underlying engine, available functionality, interactive time limits, etc",
+        "tags": ["initialize"],
+        "parameters": [],
+        "responses": {
```
ttung: this response has a pretty well-defined schema.  is it possible to define it here?

ttung: Can do a follow up diff if it helps unblock you.

server/test/test_scanpy_engine.py
```diff
@@ -98,5 +98,9 @@ def test_schema_produces_error(self):
         with pytest.raises(TypeError):
             self.data._create_schema()
 
+    def test_config(self):
+        self.assertEqual(self.data.features["layout"]["obs"], {'available': True, 'interactiveLimit': 15000})
```
ttung: this magic value is in the code twice. consider making it (and all the other interactiveLimit values) constants.

server/app/scanpy_engine/scanpy_engine.py
csweaver: Definitely would prefer a solution that has me updating versions less. You are suggesting we only version cellxgene instead of either the driver or the engines, right?

ttung: Exactly.

server/app/driver/driver.py
csweaver: Cool, can do!

server/app/rest_api/rest.py
csweaver: Sure, I'm going to postpone this to a new PR though. I'll make a ticket

server/app/scanpy_engine/scanpy_engine.py
csweaver: Ok, moved it to the cxg version. I'll just bump that faster than I ordinarily would.

server/test/test_scanpy_engine.py
csweaver: I'm going to do that when I do the plugin architecture and the limits are actually real. Right now they are mostly just placeholders

PR #204, opened 2018-08-28
Fixed bug where names were NaNs

csweaver:
The names for obs and var were being set to NaNs since the indices of obs and var start as names instead of sequential integers. Had to add the dataframe's index to the series.


csweaver:
Good idea

PR #215, opened 2018-09-06
parse_filter parses v0.2 GET filters

csweaver:
This is basically a rewrite of the parse_filter function to parse the GET version of complex filters.

GET /data?obs:tissue=lung&obs:tissue=heart&obs:num_reads=1000,*

### DIFF
--- a/server/app/util/filter.py
+++ b/server/app/util/filter.py
@@ -60,38 +60,36 @@ def parse_filter(filter, schema):
             raise QueryStringError(f"Error: bad axis in query string {axis}")
         if axis not in query:
             query[axis] = {}
-        if annotation == "index":
-            query[axis]["index"] = filter.getlist(key)
+
+        if "annotation_value" not in query[axis]:
+            query[axis]["annotation_value"] = []
+        ann_filter = {"name": annotation}
+        for ann in schema[axis]:
+            if ann["name"] == annotation:
+                dtype = ann["type"]
+                break
         else:
-            if "annotation_value" not in query[axis]:
-                query[axis]["annotation_value"] = []
-            ann_filter = {"name": annotation}
-            for ann in schema[axis]:
-                if ann["name"] == annotation:
-                    dtype = ann["type"]
-                    break
-            else:
-                raise QueryStringError(f"Error: {annotation} not a valid annotation type")
-            if dtype in ["string", "categorical", "boolean"]:
-                ann_filter["value"] = [_convert_variable(dtype, i) for i in filter.getlist(key)]
-            else:
-                value = filter.get(key)
-                try:
-                    min_, max_ = value.split(",")
-                except ValueError:
-                    raise QueryStringError(f"Error: min,max format required for range for {annotation}, got {value}")
-                if min_ == "*":
-                    min_ = None
-                if max_ == "*":
-                    max_ = None
-                try:
-                    ann_filter["value"] = {
-                        "min": _convert_variable(dtype, min_),
-                        "max": _convert_variable(dtype, max_)
-                    }
-                except ValueError:
-                    raise QueryStringError(
-                        f"Error: expected type {dtype} for key {annotation}, got {value}"
-                    )
-            query[axis]["annotation_value"].append(ann_filter)
+            raise QueryStringError(f"Error: {annotation} not a valid annotation type")
+        if dtype in ["string", "categorical", "boolean"]:
+            ann_filter["value"] = [_convert_variable(dtype, i) for i in filter.getlist(key)]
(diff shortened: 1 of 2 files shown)


server/app/util/filter.py
```diff
@@ -9,18 +14,27 @@ def _convert_variable(datatype, variable):
...
+    assert datatype in ["boolean", "categorical", "float32", "int32", "string"]
+    if variable is None:
         return variable
+    try:
+        if datatype == "int32":
+            variable = int32(variable)
+        elif datatype == "float32":
+            variable = float32(variable)
+        elif datatype == "boolean":
+            if variable == "false":
```
ttung: `json.loads(variable)` ?

server/app/util/filter.py
```diff
@@ -9,18 +14,27 @@ def _convert_variable(datatype, variable):
...
+                variable = False
+            elif variable == "true":
+                variable = True
+            else:
+                raise ValueError
```
ttung: When you raise an exception, you should raise an instance, i.e., `raise ValueError()`.

bonus points for adding an exception message.

server/app/util/filter.py
```diff
@@ -9,18 +14,27 @@ def _convert_variable(datatype, variable):
...
     except ValueError:
```
ttung: I think you can remove the entire try: except: construct, and just return variable.

server/app/util/filter.py
```diff
@@ -35,36 +49,47 @@ def parse_filter(filter, schema):
...
-        query[key] = {
-            "variable_type": schema[key]["variabletype"],
-            "value_type": schema[key]["type"]
-        }
-        if query[key]["variable_type"] == "categorical":
-            query[key]["query"] = [_convert_variable(query[key]["value_type"], v) for v in value]
-        elif query[key]["variable_type"] == "continuous":
-            value = value[0]
+        axis, annotation = key.split(":", 1)
+        if axis not in [ax for ax in Axis]:
```
ttung: recommend:

```
try:
    Axis(axis)
except ValueError:
    raise QueryStringError(f"Error: bad axis in query string {axis}")
```

enumtype(value) will look up an enum by value.

server/app/util/filter.py
```diff
@@ -35,36 +49,47 @@ def parse_filter(filter, schema):
...
+            raise QueryStringError(f"Error: bad axis in query string {axis}")
+        if axis not in query:
```
ttung: consider [collections.defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict) (i would not necessarily adopt it, but you can look at it and see if you think it's right for you)

server/app/util/filter.py
```diff
@@ -35,36 +49,47 @@ def parse_filter(filter, schema):
...
+            query[axis] = {}
+
+        if "annotation_value" not in query[axis]:
```
ttung: could also use collections.defaultdict here if you want.

server/app/util/filter.py
```diff
@@ -35,36 +49,47 @@ def parse_filter(filter, schema):
     The results is a matrix with the cells the pass the filter and at this point all the genes
     :param filter: flask's request.args
     :param schema: dictionary schema
+    :raises QueryStringError
     :return:
     """
     query = {}
+
     for key in filter:
```
ttung: fyi, filter is a python keyword.  may be confusing to use it as a variable name.

ttung: also, what's the type of filter?  can you extract the key + value at the same time like a dict?

server/app/util/filter.py
csweaver: Cool! Didn't realize loads didn't have to take an object/array

server/app/util/filter.py
csweaver: It's ImmutableMultiDict type. The reason I'm not getting the values is that I need to use `getlist` method if it is a categorical type and `get` if it is a numerical type.

server/app/util/filter.py
csweaver: Added that. Good catch, I had meant to use default dict and forgot.

server/app/util/filter.py
csweaver: Done. I had some trouble figuring out the best way to check that, thanks

server/app/util/filter.py
csweaver: Fixed.

server/app/util/filter.py
csweaver: good point. moot now, though

PR #216, opened 2018-09-06
Cleanup Tasks

csweaver:
Minor cleanup tasks
- add the API tests back in (😨 can't believe they weren't already in...)
- Use a custom JSON encoder for responses so I don't have to do weird stuff when making JSON encoded responses with the API. This eliminates the need for a custom response making function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,6 @@ script:
   - set -eo pipefail
   - flake8 server/app/
   - pytest -s server/test/test_filter.py server/test/test_scanpy_engine.py
+  - cellxgene scanpy example-dataset/ &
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -15,3 +15,6 @@ script:
   - set -eo pipefail
   - flake8 server/app/
   - pytest -s server/test/test_filter.py server/test/test_scanpy_engine.py
+  - cellxgene scanpy example-dataset/ &
+  - for i in {1..90}; do if http :5005/api/v0.1/initialize > /dev/null; then break; else echo "Waiting for server..."; sleep 1; fi; done
+  - pytest server/test/test_api.py
```
ttung: recommend that you just add the server initialization to the test setup, and then you can just put the whole thing in pytest.

PR #220, opened 2018-09-14
CLI improvements

flying-sheep:
This adds a `__main__.py` that still works with werkzeug’s bugged reloading behavior. This way one can run cellxgene via `python -m server` from the working directory without installing the package.

This PR also improves the error reporting of the CLI:

1. There‘s now a helpful error if you don’t provide a subcommand (like when not providing any arguments).
2. Python disables ImportWarnings by default, so I reenabled them them to see the scanpy warning!
3. If none of the engines could be loaded, it now fails with a nicer error message.

### DIFF
--- a/server/__main__.py
+++ b/server/__main__.py
@@ -1,0 +1,12 @@
+# Work around bug https://github.com/pallets/werkzeug/issues/461
+if __package__ is None:
+    import sys
+    from pathlib import Path
+    PKG_PATH = Path(__file__).parent
+    sys.path.insert(0, str(PKG_PATH.parent))
+    import server
+    __package__ = PKG_PATH.name
+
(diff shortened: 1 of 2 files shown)


csweaver:
Hi @flying-sheep and welcome to cellxgene! Thank you for your PR -- good idea. I'd like to welcome you to join the cellxgene slack channel on CZIScience slack org as well! https://cziscience.slack.com/messages/CCTA8DF1T

Can I request that you resubmit your PR against the restv2 branch? We are in the middle of some refactoring right now and we are trying to keep the updates to the server folder in restv2 while we develop on the client folder in master.

flying-sheep:
OK, sure! Will create a new PR

PR #221, opened 2018-09-15
CLI improvements

flying-sheep:
This adds a `__main__.py` that still works with werkzeug’s bugged reloading behavior. This way one can run cellxgene via `python -m server` from the working directory without installing the package.

This PR also improves the error reporting of the CLI:

1. There‘s now a helpful error if you don’t provide a subcommand (like when not providing any arguments).
2. Python disables ImportWarnings by default, so I reenabled them them to see the scanpy warning!
3. If none of the engines could be loaded, it now fails with a nicer error message.


csweaver:
Actually, the line 68 where you add the parameter required=True is causing the failure. It's strange since this is a valid param for add_subparsers. Especially since the following works:

    subparsers = parser.add_subparsers(dest="engine")
    subparsers.required = True

Do you want to just go with the above?

flying-sheep:
Hey! what do you think?

bkmartinjr:
@ttung - ping....   Let us know if you don't have time to review as I'd like to get this merged.  Tnx!

ttung:
> @ttung - ping.... Let us know if you don't have time to review as I'd like to get this merged. Tnx!

Oh, I thought we were operating on single approval.

I personally don't see much value in supporting the use case of running the server without installing it first.  End users _should_ be installing it, we might use pkg_resources to find data, and tests also don't work without installing the package.

The rest look good to me though.

Issue #184, opened 2018-08-11
add ability to serve HTTP requests over interfaces other than loopback


bkmartinjr:
Currently, the CLI will only serve HTTP requests via loopback.   This is good for a default.  However, we need a CLI option to serve on other interfaces so that the CLI can be used to serve data to browsers running on other machines.

Possible UX: add a --host option, which defaults to loopback.

PR #225, opened 2018-09-16
Option to bind to all interfaces

ttung:
app.run("0.0.0.0") instead of app.run("127.0.0.1") binds to all interfaces.

Note: There are comments on the internet that says that the flask server is not up to the task of production serving.  I don't think that such scalability concerns apply here, but I was able to get cellxgene working with twistd relatively easily, and we could switch to that if there are scalability concerns.

Test plan: browsed to <ip>:5005/api/v0.2/config on a different host.

Fixes https://github.com/chanzuckerberg/cellxgene/issues/184

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -57,7 +57,11 @@ def run_scanpy(args):
 
     from .scanpy_engine.scanpy_engine import ScanpyEngine
     app.data = ScanpyEngine(args.data_directory, layout_method=args.layout, diffexp_method=args.diffexp)
-    app.run(host="127.0.0.1", debug=True, port=args.port)
+    if args.bind_all:
+        host = "0.0.0.0"
+    else:
+        host = "127.0.0.1"
+    app.run(host=host, debug=True, port=args.port)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


server/app/app.py
```diff
@@ -57,14 +57,22 @@ def run_scanpy(args):
...
+        host = "127.0.0.1"
+    app.run(host=host, debug=True, port=args.port)
 
 
 def main():
     parser = argparse.ArgumentParser(description="Cellxgene is a tool for exploring single cell expression.")
     parser.add_argument("--title", "-t", help="Title to display -- if this is omitted the title will be the name "
                                               "of the directory from the data_directory arg")
     parser.add_argument("--port", help="Port to run server on.", type=int, default=5005)
+    parser.add_argument(
```
bkmartinjr: What is the relative merit of --bind-all (ie, 0,0.0.0) vs. having a --bind-address option that simply takes an address (could be 0.0.0.0 or any other)?   I was imagining that there would be deployment scenarios where it would be beneficial to find to a specific interface, rather than all.

I don't feel strongly about this - just wanted to float the concern in case it actually is preferable.

ttung: I think `--bind-address` is likely confusing to those who are not super savvy about running ifconfig/ipconfig and finding their ip addresses.  It's also pretty unlikely that someone will be running multiple non-localhost interfaces and not want to bind to all of them.  Having said that, it's likely someone who's trying to deploy this will likely know how to do so.

Lastly, this is not properly ipv6 dual-stacked, but I guess we can cross that bridge when someone cares...? :)

bkmartinjr: I agree that localhost should be the default, and I'm not opposed to --bind-all as a shortcut.  I'm just thinking about our intent to support web-facing deployments in the near future - to let people stand up a cellxgene backend and serve data to remote users.

For a single interface solution, is there any security benefit to binding to a single interface, rather than 0.0.0.0?

bkmartinjr: we have decided --bind-all is enough

ttung:
I don't have write access so someone else will have to merge this. :)

bkmartinjr:
@ttung I would add you, but apparently I don't have permissions either.

I'll do the merge.

PR #235, opened 2018-09-17
PUT /annotations

csweaver:
Add PUT route for annotations. This allows a user to submit a complex filter in the body.

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -170,7 +170,7 @@ class AnnotationsObsAPI(Resource):
             "in": "query",
             "name": "annotation-name",
             "type": "string",
-            "description": "comma-separated annotation keys, ex: num_genes,percent_mito"
+            "description": "list of 1 or more annotation names"
         }],
         "responses": {
             "200": {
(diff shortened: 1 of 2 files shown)


server/app/rest_api/rest.py
```diff
@@ -201,6 +201,53 @@ def get(self):
...
 
+    @swagger.doc({
+        "summary": "Fetch annotations (metadata) for filtered subset.",
+        "tags": ["annotations"],
+        "parameters": [
+            {
+                "in": "query",
+                "name": "annotation-name",
+                "type": "string",
+                "description": "comma-separated annotation keys, ex: num_genes,percent_mito"
```
bkmartinjr: Description is obsolete - we now take a normal query string, rather than a comma-separated list.

server/app/rest_api/rest.py
```diff
@@ -201,6 +201,53 @@ def get(self):
...
+        }
+    })
+    def put(self):
+        fields = request.args.getlist("annotation-name", None)
+        df = current_app.data.filter_dataframe(request.get_json()["filter"])
+        try:
+            annotation_response = current_app.data.annotation(df, fields)
+        except KeyError:
+            return make_response(f"Error bad key in {fields}", 404)
+        else:
```
ttung: technically, since you return on all the exception handling paths, you don't need the else.

csweaver: fixed

server/app/rest_api/rest.py
csweaver: fixed

PR #248, opened 2018-09-20
Annotation var

csweaver:
Add var route for annotations to REST api

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -134,10 +134,10 @@ class LayoutObsAPI(Resource):
         "tags": ["layout"],
         "parameters": [
             {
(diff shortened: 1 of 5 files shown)


server/app/rest_api/rest.py
```diff
@@ -224,13 +223,96 @@ def get(self):
...
+                            [1, "RER1", None],
+                            [49, "S100B", 6]
+                        ]
+                    }
+
+                }
+            }
+        }
+    })
+    def get(self):
```
ttung: is this not the same as the put request?  i understand this is because some requests might exceed the 2k or so limit for GET requests, but maybe you can have them call the same code.

PR #262, opened 2018-09-24
Add Dockerfile and make it run in Docker

flying-sheep:
Not sending `client/node_modules` to the build context is a good practice, so I’m doing that via `.dockerignore`.

Not installing deterministic versions isn’t, however I don’t know what specific problem with your package-lock.json makes it all trip up, so I’m not including it in the build context via `.dockerignore`.

The packages I exclude from `requirements.txt` are already in the base docker image, so I could verify that it works with the versions in the image as well. Not installing different versions saves a lot of time.


csweaver:
Hey @flying-sheep! Great idea. Thanks for contributing! Would you mind joining the slack channel (https://cziscience.slack.com/messages/CCTA8DF1T)? 

Unfortunately I'm at the Grace Hopper Conference all this week so I won't have a chance to properly reivew this until next week.

flying-sheep:
Hi! I would, but it says I need an @chanzuckerberg mail address

bkmartinjr:
@flying-sheep - just dropped you an email...

flying-sheep:
It’s a much easier setup, see the added README entry. It will always work, independently of system dependencies. You can use `docker-compose` to easily fire up and control one or multiple instances as part of a server fleet.

server/app/app.py
```diff
@@ -44,10 +44,11 @@
 
 
 def run_scanpy(args):
+    host = "0.0.0.0"  # Necessary for redirection
```
csweaver: Can you pull the changes from master for app.py? I think one of our other PRs have solved this problem.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+COPY [ "bin/",    "/cellxgene/bin/" ]
+COPY [ "server/", "/cellxgene/server/" ]
+COPY [ "client/", "/cellxgene/client/" ]
+COPY [ "setup.py", "MANIFEST.in", "README.md", "/cellxgene/" ]
+
+WORKDIR /cellxgene
+
+RUN \
+    apk --no-cache add --virtual .builddeps bash nodejs nodejs-npm && \
```
csweaver: Add :
```
bash bin/clean && \ 
```
and this will eliminate the need for the .gitignore file

flying-sheep:
hi! the problem here is that installing the dependencies for scanpy is pretty complex, specifically python-igraph.

i feel uncomfortable duplicating the whole work going on in [here](https://github.com/FASTGenomics/base_image_alpine_scanpy/blob/master/build.sh). It’s fragile enough to maintain the whole shebang in one place, let alone two.

what out do you suggest?

csweaver:
Thanks for clarifying, we didn’t quite appreciate the complexity of scanty packaging, though I’ve been there in the past… :) At the moment we are working on getting a release out and working on some other installation-related issues which may help clarify which version of python libraries we are going to support. We really like this idea, but we want to support it long term which will require a bit more work.  We’d would like to put this on hold for a few weeks and then revisit as part of a broader installation refactoring. Can leave this open until then so we remember to sort it out?

flying-sheep:
Oh, i just realized I didn’t reply. Of course!

csweaver:
Hi @flying-sheep! Now that we have a release, and can pip-install the cellxgene package, it makes the docker container creation a little more straightforward than what you had to do. I created a PR [(#505 )](https://github.com/chanzuckerberg/cellxgene/pull/505) to add a docker file.  I'm hoping you'll take a look and let me know if it fits your use case and if you have any suggestions. Thanks again for your contributions, we really appreciate it.

flying-sheep:
OK, sounds good!

PR #265, opened 2018-09-24
Docs

bkmartinjr:
Create docs directory.  Populate with first commit of the REST 0.2 API doc in MD (supersedes gdoc version)

### DIFF
--- a/docs/REST_API.md
+++ b/docs/REST_API.md
@@ -68,11 +68,11 @@ There are several common filters supported by various routes. Route descriptions
 
 Filter by the value of specific observation or variable annotation values. For example, the `/data` routes support sub-selection by the value of the annotations, which might be used to filter by some specific metadata (eg, `tissue_type: lung`).
 
-Annotation value filters can be specified as a URL query parameter (GET), or a JSON object in a request body (POST)[^1].
+Annotation value filters can be specified as a URL query parameter (using GET), or a JSON object in a request body (using PUT or POST)<sup>[1](#footnote-1)</sup>.
 
 For a GET URL query parameter:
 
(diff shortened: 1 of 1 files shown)


docs/REST_API.md
```diff
@@ -0,0 +1,679 @@
...
+
+## Filters
+
+There are several common filters supported by various routes. Route descriptions supporting these filters will reference the filter types.
+
+### Annotation _Value_ Filters
+
+Filter by the value of specific observation or variable annotation values. For example, the `/data` routes support sub-selection by the value of the annotations, which might be used to filter by some specific metadata (eg, `tissue_type: lung`).
+
+Annotation value filters can be specified as a URL query parameter (GET), or a JSON object in a request body (POST)[^1].
```
csweaver: Can we replace POST with POST/PUT? Most of the urls are using PUT instead of POST so I think this is a bit confusing. There are a few instances of this.

docs/REST_API.md
```diff
@@ -0,0 +1,679 @@
...
+
+For a GET URL query parameter:
+
+- Annotation name is encoded as `obs:name` or `var:name` [^2].
```
csweaver: Could we use a different footnote annotation method? I like https://stackoverflow.com/a/29384216/1181324 as it adds a link. The carat format is unfamiliar to me.

docs/REST_API.md
bkmartinjr: yes, will do

docs/REST_API.md
bkmartinjr: yes, will do.

PR #268, opened 2018-09-24
Fix jest test suites

### LINKED ISSUE #229: Make Jest tests work with babel
Pre-ship, revitalize the Jest tests

This will require babelizing

bkmartinjr:
Resolves #229 

Changes:
* all test files now use LF, rather than CRLF as a line terminator
* enable babel transformation when running jest:
    * devDependency addition: babel-jest
    * added a test-only babel config to package.json (rather than a .babelrc)
* fix TWO (!) latent bugs the revitalized test suites found.

@colinmegill - when you review, diff ignoring white space (so you can see past the CRLF edits)


bkmartinjr:
Committing without review.

PR #270, opened 2018-09-24
Diffexp/obs

csweaver:
This is the quick and dirty version of /diffexp/obs. Anything not absolutely necessary for getting the FE running has been ignored.


csweaver:
I'm going to close this and open a new one with the combined scanpy engine and rest api

PR #273, opened 2018-09-25
Csweaver/rest diffexp

csweaver:
added t-test diffexp for scanpy engine. This is the quick and dirty version of /diffexp/obs. Anything not absolutely necessary for getting the FE running has been ignored.

a little unsure about Bonferroni correction (@freeman-lab can you please check that the stats make sense?) Bonferroni not at right sigfig currently for sure.
@ttung wasn't sure if the _top_sort method should be in the engine or utils. Went with engine since I didn't imagine using it elsewhere

### DIFF
--- a/server/test/test_api.py
+++ b/server/test/test_api.py
@@ -294,5 +294,4 @@ class EndPoints(unittest.TestCase):
         file = "js/service-worker.js"
         url = f"{LOCAL_URL}{endpoint}/{file}"
         result = self.session.get(url)
-        print(result.status_code)
         self.assertEqual(result.status_code, 200)
--- a/server/test/test_api.py
+++ b/server/test/test_api.py
@@ -105,7 +105,6 @@ class EndPoints(unittest.TestCase):
         result = self.session.get(url)
         self.assertEqual(result.status_code, 404)
 
-
     def test_put_annotations(self):
         endpoint = "annotations/obs"
         url = f"{URL_BASE}{endpoint}"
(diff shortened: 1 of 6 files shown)


server/test/test_api.py
```diff
@@ -147,6 +147,34 @@ def test_filter_put_annotations_obs(self):
         self.assertEqual(len(result_data["data"][0]), 3)
         self.assertEqual(len(result_data["data"]), 15)
 
+    def test_diff_exp(self):
+        endpoint = "diffexp/obs"
+        url = f"{URL_BASE}{endpoint}"
+        params = {
```
bkmartinjr: This test does not conform to the spec.    mode should be an object and contain "count" and "name", eg,

{  mode:  { count: 7, name: "topN" } }

server/test/test_api.py
```diff
@@ -147,6 +147,34 @@ def test_filter_put_annotations_obs(self):
...
+            "mode": "topN",
+            "set1": {
+                "filter": {
+                    "obs": {"annotation_value": [
+                        {"name": "louvain", "values": ["NK cells"]}
+                    ]
+                    }
+                }
+            },
+            "set2": {
```
bkmartinjr: super minor point - the current f/e usage is filtering by index, not annotation value.   Might want to add a test for that?

server/test/test_api.py
bkmartinjr: spec is wrong (inconsistent) - will fix it.

server/test/test_api.py
csweaver: got it

PR #284, opened 2018-09-28
Restv2 feature branch merge to master

bkmartinjr:
Initial PR containing the port to the new REST 0.2 API.   This includes partial route implementation on the back-end, sufficient to support the current front-end requirements.   Front-end should be functionally equivalent to the current REST 0.1 implementation (notably, relayout on regraph is not implemented - yet).


client/src/actions/index.js
```diff
@@ -17,44 +15,44 @@ function catchErrorsWrap(fn) {
...
-      })
-    }
-  );
+/*
+Bootstrap application with the initial data loading.
+  * /config - application configuration
+  * /schema - schema of dataframe
+  * /annotations/obs - all metadata annotation
+*/
+const doJsonRequest = async url => {
```
colinmegill: The name implies to me that this is general / generic and will be consumed by multiple other actions. If we intend on using this as a standalone ever, I would suggest renaming to something more specific.

colinmegill: Ie., this could almost be broken out into a 'util' file as it does not look like a proper action

client/src/actions/index.js
```diff
@@ -131,93 +128,81 @@ const resetGraph = () => (dispatch, getState) =>
...
+        dispatch,
+        getState,
+        [gene]
+      );
       dispatch({
         type: "color by expression",
         gene,
         data: {
-          [gene]: getState().controls.world.varDataCache[gene]
+          [gene]: expressionData[gene]
```
colinmegill: That's nice!

client/src/util/stateManager/universe.js
```diff
@@ -4,138 +4,53 @@ import _ from "lodash";
...
-aka all of the var/obs data and annotations.
-
-These functions are used exclusively by the actions and reducers to
-build an internal POJO for use by the rendering components.
+Private helper function - create and return a template Universe
 */
+function templateUniverse() {
+  /* default universe template */
+  const VarDataCacheLowWatermark = 32;
+  const VarDataCacheTTLMs = 1000;
```
colinmegill: Would appreciate clarifying comment on this var name & magic number

client/src/actions/index.js
bkmartinjr: actually fixed a bug :-)

client/src/actions/index.js
bkmartinjr: I'm not up to speed on proper style in action creators.  Should there be (literally) no helper code in these files?   If so, I can move both this and `catchErrorsWrap` into a utils helper file.

PR #287, opened 2018-10-01
add Jest tests to travis

bkmartinjr:


.travis.yml
```diff
@@ -15,3 +15,4 @@ script:
   - set -eo pipefail
   - flake8 server/app/
   - pytest -s server/test
+  - npm run test
```
csweaver: We need to prefix this to the client directory or it won't be able to find the package file:
`npm run --prefix client/ test`

bkmartinjr: thank you!

PR #293, opened 2018-10-02
Guarantee sort order for data return by index

csweaver:


bkmartinjr:
the argument for server side is that often sorting is a noop if done on the server (because the server can choose indexing schemes to make that so).  The client can't - it actually has to sort.

PR #295, opened 2018-10-03
/data/var

csweaver:
Able to access data from var axis. Also fixed a bug where accessing a single var or obs would cause failure.

@ttung to head off what I already know you will say, I'm going to do a batch cleanup of duplicate documentation once I've taken care of other high priority development, it's coming.

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -478,7 +478,7 @@ class DataObsAPI(Resource):
             return make_response(e.message, HTTPStatus.BAD_REQUEST)
         df = current_app.data.filter_dataframe(filter_, include_uns=False)
         if accept_type and accept_type[0] == "application/json":
-            return make_response((jsonify(current_app.data.data_frame(df))))
+            return make_response((jsonify(current_app.data.data_frame(df, axis=Axis.OBS))))
         # TODO support CSV
         else:
             return make_response(f"Unsupported accept-type: {accept_type}", HTTPStatus.NOT_ACCEPTABLE)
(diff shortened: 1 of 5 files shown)


server/test/test_api.py
```diff
@@ -265,71 +265,81 @@ def test_filter_put_annotations_var(self):
...
-        query = "accept-type=application/json"
-        url = f"{URL_BASE}{endpoint}?{query}"
-        result = self.session.get(url)
-        self.assertEqual(result.status_code, 200)
-        result_data = result.json()
-        self.assertEqual(len(result_data["obs"]), 2638)
+        for axis in ["obs", "var"]:
+            endpoint = f"data/{axis}"
+            query = "accept-type=application/json"
+            url = f"{URL_BASE}{endpoint}?{query}"
```
bkmartinjr: also need to test without a query param to validate that the default content-type is returned

server/test/test_api.py
```diff
@@ -265,71 +265,81 @@ def test_filter_put_annotations_var(self):
...
-                    ]
+        for axis in ["obs", "var"]:
+            endpoint = f"data/{axis}"
+            url = f"{URL_BASE}{endpoint}"
+            header = {"Accept": "application/json"}
+            var_filter = {
+                "filter": {
+                    "var": {
+                        "annotation_value": [
+                            {"name": "name", "values": ["RER1"]},
```
bkmartinjr: also worth doing multiple values, as the app currently uses both cases

server/app/rest_api/rest.py
```diff
@@ -520,7 +520,103 @@ def put(self):
...
+            "400": {
+                "description": "Malformed filter"
+            },
+            "406": {
+                "description": "Unacceptable MIME type"
+            },
+        }
+    })
+    def put(self):
+        if not request.accept_mimetypes.best_match(["application/json", "text/csv"]):
```
bkmartinjr: to confirm: does this correctly handle the default case (where no Accept) is specified, and JSON should be returned?

server/app/rest_api/rest.py
```diff
@@ -520,7 +520,103 @@ def put(self):
...
+    def get(self):
+        # request.args is immutable
+        args = dict(request.args)
+        accept_type = args.pop("accept-type", None)
+        try:
+            filter_ = parse_filter(ImmutableMultiDict(args), current_app.data.schema['annotations'])
+        except QueryStringError as e:
+            return make_response(e.message, HTTPStatus.BAD_REQUEST)
+        df = current_app.data.filter_dataframe(filter_, include_uns=False)
+        if accept_type and accept_type[0] == "application/json":
```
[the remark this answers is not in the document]
csweaver: agreed, I've still got that on my list todo, I was going to knock it out in /data/var and data/obs at the same time

csweaver:
I was going to fix the content negotiation in a separate PR since I needed to do that for a few routes. Is that OK with you?

bkmartinjr:
Yes, I'm fine with this.  Will approve.

PR #296, opened 2018-10-04
Add arbitrary genes as continuous crossfilters

colinmegill:


client/src/reducers/controls.js
```diff
@@ -35,6 +35,7 @@ const Controls = (
     categoricalAsBooleansMap: null,
     crossfilter: null,
     dimensionMap: null,
+    dimensionMapVar: {},
```
bkmartinjr: not sure this gets used - extraneous?

client/src/reducers/controls.js
```diff
@@ -128,6 +134,16 @@ const Controls = (
           );
         }
       });
+
+      _.forEach(action.expressionData, (val, key) => {
+        dimensionMap[key] = World.createVarDimension(
```
bkmartinjr: Note that the current implementation will 'lose' all of these dimensions whenever you regraph (set world to the current selection).   The gene expression histograms will still be displayed (because you are displaying a histogram for anything in world.varDataCache), so if the user brushes the histogram, they will get a hard error.

General rule of thumb:
- dimensionMap needs to be kept in sync with world
- it also needs to be kept in sync with whatever state you use to store the list of genes to display a histogram for (currently, world.varDataCache).

Currently, the only actions that mess with this are:
* `reset World to eq Universe`
* `set World to current selection`

These rebuild the obs annotation dimensionMap when world changes.   If you want to keep things consistent, you need to rebuild the var data dimensions as well.

client/src/components/geneExpression/index.js
```diff
@@ -0,0 +1,183 @@
...
+              display: "inline-block",
+              marginLeft: 7,
+              color: globals.brightBlue,
+              fontSize: 22
+            }}
+          />{" "}
+          add gene{" "}
+        </button>
+        {this.props.world
+          ? _.map(this.props.world.varDataCache, (value, key) => {
```
bkmartinjr: As we discussed, I think a better approach is to:
* store a list of genes in state.control, eg, state.control.geneHistograms
* generate these histograms from that list
* anytime you add a gene to this list, also call the action handler to fetch the expression data and populate the varDataCache
* limit the size of this geneHistogram list, and ensure that the varDataCache lowWatermark config is larger than this limit
* make sure you keep state.controls.dimensionMap updated wiht any changes to this gene list, especially when world changes (see my notes on reducers/controls.js)

colinmegill:
Diffexp, regraph, brush implemented as of b80b74e, by recreating dimensions on the two actions mentioned by @bkmartinjr's comment above.

screen shot 2018-10-09 at 6 34 35 pm

client/src/reducers/controls.js
```diff
@@ -75,6 +78,39 @@ const Controls = (
       const categoricalAsBooleansMap = createCategoricalAsBooleansMap(world);
       const crossfilter = Crossfilter(world.obsAnnotations);
       const dimensionMap = World.createObsDimensionMap(crossfilter, world);
+
+      let worldVarDataCache = world.varDataCache;
```
bkmartinjr: use `const`, not `let`

client/src/reducers/controls.js
```diff
@@ -35,6 +35,8 @@ const Controls = (
     categoricalAsBooleansMap: null,
     crossfilter: null,
     dimensionMap: null,
+    userDefinedGenes: [],
```
bkmartinjr: I realize this issue is not new, but I suggest we start a comment on line 23 that documents the purpose/semantics of each field.   Eg, 

/*
  dimensionMap:  object containing all dimensions created on the current crossfilter.  Obs annotations are keyed by annotation name, layout keyed by 'X' and 'Y', and varData cached by var name.
*/

client/src/reducers/controls.js
```diff
@@ -75,6 +78,39 @@ const Controls = (
...
+
+      /* var dimensions */
+      if (userDefinedGenes.length > 0) {
+        /*
+          verbose & slightly confusing that we also access this as an object
+          in controls rather than an array, should be abstracted into
+          util ie., createDimensionsFromBothListsOfGenes(userGenes, diffExp)
+        */
+        _.forEach(userDefinedGenes, (gene, index) => {
+          dimensionMap[gene] = World.createVarDimension(
```
bkmartinjr: dimensionMap keyed with a flat namespace that allows for collisions between users (obs annotations, var data names aka gene names, and layout X & Y)

client/src/reducers/controls.js
```diff
@@ -75,6 +78,39 @@ const Controls = (
...
+            /* "__var__" + */
+            world,
+            worldVarDataCache,
+            crossfilter,
+            gene
+          );
+        });
+      }
+
+      if (diffexpGenes.length > 0) {
```
bkmartinjr: so what happens if the same gene is in both userDefinedGenes and diffexpGenes?  Do you really want two dimensions created?  Currently, the first one is leaked (memory leak because dispose is never called when it is over-written).

I suspect you want to create a single (unique) dimension for each 'gene', so you probably should it on the union of userDefinedGenes and diffexpGenes

client/src/reducers/controls.js
```diff
@@ -101,6 +139,39 @@ const Controls = (
...
```
bkmartinjr: see above note about leaking / clobbering dimensions in the case where you have the same gene name in both userDefinedGenes and diffexpGenes.

client/src/reducers/controls.js
```diff
@@ -140,30 +222,93 @@ const Controls = (
...
+      });
       return {
         ...state,
-        loading: false,
-        error: action.error
+        diffexpGenes: _diffexpGenes
       };
     }
+    case "clear differential expression":
+      const { world, universe, crossfilter, dimensionMap } = state;
```
bkmartinjr: illegal declaration in a lexical scope (lint) - use curly braces...

client/src/reducers/controls.js
```diff
@@ -140,30 +222,93 @@ const Controls = (
...
 
-    /*******************************
-             User Events
-     *******************************/
-    case "parallel coordinates axes have been drawn": {
+      _.forEach(action.diffExp, values => {
+        const name = world.varAnnotations[values[0]].name;
+        // clean up crossfilter dimensions
+        const filterID = dimensionMap[name];
+        const dimension = crossfilter.filters.find(d => {
```
bkmartinjr: this will destroy the dimension for any user defined gene

(same name space collision noted above)

client/src/reducers/controls.js
```diff
@@ -140,30 +222,93 @@ const Controls = (
...
+    case "user defined gene": {
+      /* this could also live in expression success with a conditional, but that handles diffexp also */
+      const newUserDefinedGenes = state.userDefinedGenes.slice();
+      newUserDefinedGenes.push(action.data);
+      return {
+        ...state,
+        userDefinedGenes: newUserDefinedGenes
+      };
+    }
+    case "clear user defined gene": {
```
bkmartinjr: shouldn't this also remove the dimension associated with this gene?

client/src/actions/index.js
```diff
@@ -265,5 +283,6 @@ export default {
   requestSingleGeneExpressionCountsForColoringPOST,
   requestGeneExpressionCountsPOST,
```
bkmartinjr: `requestGeneExpressionCountsPOST` is no longer used outside this module (ie, it is now a module-internal helper).  It probably should not be exported.

bkmartinjr: digging in more - it looks to me like `requestGeneExpressionCountsPOST()` can be delete, as it is a NOOP wrapper around `_doRequestExpressionData()`.   While `requestGeneExpressionCountsPOST()` does emit some actions, they are not used, and are identical in semantics to those emitted by `_doRequestExpressionData()`

Suggest removing this extraneous layer.

client/src/reducers/controls.js
```diff
@@ -222,37 +229,63 @@ const Controls = (
         }
       };
     }
+    case "request user defined gene success": {
```
bkmartinjr: missing a handler for "request user defined gene error"

client/src/util/nameCreators.js
```diff
@@ -0,0 +1,43 @@
+/*
```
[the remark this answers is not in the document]
colinmegill: Yes, will start with nameCreators

PR #299, opened 2018-10-04
REST Error handling

csweaver:
Adding documentation, logic, and tests to handle errors specified by the spec document (added in the  ones that fell through the cracks on the first pass)

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -206,7 +206,7 @@ class AnnotationsObsAPI(Resource):
 
                 }
             },
-            "404": {
+            "400": {
                 "description": "one or more of the annotation-name identifiers were not associated with an "
                                "annotation name"
             }
@@ -217,7 +217,7 @@ class AnnotationsObsAPI(Resource):
         try:
             annotation_response = current_app.data.annotation(current_app.data.data, "obs", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
+            return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         return make_response(jsonify(annotation_response), HTTPStatus.OK)
 
     @swagger.doc({
@@ -256,11 +256,8 @@ class AnnotationsObsAPI(Resource):
                 }
             },
             "400": {
-                "description": "malformed filter"
-            },
-            "404": {
-                "description": "one or more of the annotation-name identifiers were not associated with an"
-                               " annotation name"
+                "description": "malformed filter or one or more of the annotation-name identifiers were"
+                               "not associated with an annotation name"
             }
         }
     })
@@ -273,7 +270,7 @@ class AnnotationsObsAPI(Resource):
         try:
             annotation_response = current_app.data.annotation(df, "obs", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
+            return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         return make_response(jsonify(annotation_response), HTTPStatus.OK)
 
 
@@ -304,7 +301,7 @@ class AnnotationsVarAPI(Resource):
 
                 }
             },
-            "404": {
+            "400": {
                 "description": "one or more of the annotation-name identifiers were not associated with an"
                                " annotation name"
             }
@@ -315,7 +312,7 @@ class AnnotationsVarAPI(Resource):
         try:
             annotation_response = current_app.data.annotation(current_app.data.data, "var", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
+            return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         return make_response(jsonify(annotation_response), HTTPStatus.OK)
 
     @swagger.doc({
@@ -352,11 +349,8 @@ class AnnotationsVarAPI(Resource):
                 }
             },
             "400": {
-                "description": "malformed filter"
-            },
-            "404": {
-                "description": "one or more of the annotation-name identifiers were not associated"
-                               " with an annotation name"
+                "description": "malformed filter or one or more of the annotation-name identifiers were"
+                               "not associated with an annotation name"
             }
         }
     })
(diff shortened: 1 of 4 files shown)


server/app/rest_api/rest.py
```diff
@@ -392,24 +441,24 @@ def post(self):
         try:
```
bkmartinjr: Shouldn't these optional routes check to see if they are implemented by the engine, and return a 501 if not?

I would assume the engine would fill out the "features" implemented (as defined by driver), and that would then let the REST layer know when to immediately return a 501...

Not a blocker for the current release, but it seems like this file "assumes" that diffexp is *always* implemented.

Put another way - the engine should declare what it implements, and the REST layer should comply with that.

server/app/util/utils.py
```diff
@@ -10,3 +10,24 @@ def default(self, obj):
...
+
+
+def get_mime_type(default="application/json", acceptable_types=["application/json", "text/csv"], query_param=None,
+                  header=None):
+    mime_type = default
+    if query_param:
+        if query_param in acceptable_types:
+            mime_type = query_param
+        else:
+            raise MimeTypeError(f"Unsupported mime type {query_param}")
```
bkmartinjr: slightly ambiguous error message -- maybe "Unsupported mime type {query_param} specified in query parameter 'accept-type'"?

server/app/util/utils.py
```diff
@@ -10,3 +10,24 @@ def default(self, obj):
...
+    elif len(header):
+        mime_type = header.best_match(acceptable_types)
+        if not mime_type:
+            raise MimeTypeError(f"Unsupported mime type(s) {header}")
```
bkmartinjr: see above.

Maybe "... in content negotiation." or "in HTTP Accept header"?  Or?

server/app/rest_api/rest.py
```diff
@@ -201,8 +217,8 @@ def get(self):
...
             annotation_response = current_app.data.annotation(current_app.data.data, "obs", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", 404)
-        return make_response(jsonify(annotation_response))
+            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
```
ttung: I think this should probably be `BAD_REQUEST`

server/app/rest_api/rest.py
```diff
@@ -238,17 +254,27 @@ def get(self):
...
+        try:
+            df = current_app.data.filter_dataframe(request.get_json()["filter"], include_uns=False)
+        except KeyError:
+            return make_response("Malformed filter", HTTPStatus.BAD_REQUEST)
         try:
             annotation_response = current_app.data.annotation(df, "obs", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", 404)
-        return make_response(jsonify(annotation_response))
+            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
```
ttung: BAD_REQUEST

server/app/rest_api/rest.py
```diff
@@ -285,8 +315,8 @@ def get(self):
         try:
             annotation_response = current_app.data.annotation(current_app.data.data, "var", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", 404)
-        return make_response(jsonify(annotation_response))
+            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
```
ttung: BAD_REQUEST

server/app/rest_api/rest.py
```diff
@@ -320,17 +350,27 @@ def get(self):
...
+        try:
+            df = current_app.data.filter_dataframe(request.get_json()["filter"], include_uns=False)
+        except KeyError:
+            return make_response("Malformed filter", HTTPStatus.BAD_REQUEST)
         try:
             annotation_response = current_app.data.annotation(df, "var", fields)
         except KeyError:
-            return make_response(f"Error bad key in {fields}", 404)
-        return make_response(jsonify(annotation_response))
+            return make_response(f"Error bad key in {fields}", HTTPStatus.NOT_FOUND)
```
ttung: BAD_REQUEST

server/app/rest_api/rest.py
csweaver: I'm postponing this until I do the plugin architecture for the algorithms. It seemed like it wasn't going to be necessary for the initial release, and that it would be super clear how to do this with plugin architecture -- If there are no options for an algorithm then return 501.

server/app/rest_api/rest.py
```diff
@@ -158,12 +159,23 @@ def get(self):
...
+            },
+            "400": {
+                "description": "Malformed filter"
+            },
+            "403": {
+                "description": "Non-interactive request"
+            },
         }
     })
     def put(self):
```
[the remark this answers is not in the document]
csweaver: Ok, yes totally agree. But... I think that this error is outside the scope of this PR. I do a separate PR for that. 

Essentially this was an inconsistency introduced way earlier and I'd like to have a specific commit to point to when it is fixed rather than one muddled by other "central theses" (as Tony would say)

server/app/util/utils.py
csweaver: Will fix

server/app/util/utils.py
csweaver: will fix

server/app/rest_api/rest.py
csweaver: @bkmartinjr NOT_FOUND is in the spec, but I'm with @ttung on this one. NOT_FOUND has the implication that the resource doesn't exist, not just that a user supplied bad values.

bkmartinjr: Makes sense to me.  Please change the spec as well as the code.

server/app/rest_api/rest.py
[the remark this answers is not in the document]
bkmartinjr: Sounds good to me.  Lets just not forget.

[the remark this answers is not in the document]
csweaver: Working on it now

server/app/rest_api/rest.py
csweaver: will do

server/app/rest_api/rest.py
bkmartinjr: filed as issue #302

PR #301, opened 2018-10-08
/layout put-> post

csweaver:
Fixed layout's HTTP verbs to match spec


csweaver:
@bkmartinjr Do we need to pair this with the FE changes too?

bkmartinjr:
@csweaver - f/e does not yet use this route, so go ahead and make the change.

PR #304, opened 2018-10-09
Cache fix

csweaver:
Caching was broken because the method is calling the cached version if the df is any view of the same original data. Fixed by adding the var and obs indices as parameters to the memoized functions.

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -10,7 +10,7 @@ from werkzeug.datastructures import ImmutableMultiDict
 from server.app.util.constants import Axis, DiffExpMode
 from server.app.util.filter import parse_filter, QueryStringError
 from server.app.util.models import FilterModel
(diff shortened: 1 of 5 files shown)


bkmartinjr:
I just looked at the flask-cache memoization docs, and I don't think it does what we expect for short-lived objects like these filtered dataframes.

See https://flask-caching.readthedocs.io/en/latest/#memoization - in particular the section that describes how it keys the memo cache, and the big red Warning box.   TL;DR - it uses the (short lived, reused) object id() as the cache key.   We need a value that will represent the request or response (for idempotent requests).

Passing the var & obs indices does the trick because it is equivalent to the filtered dataframe 'value'.  But it is very expensive for large requests.  I wonder if there is a cheaper way to do it, eg, just pass the original REST filter as a string?  Lets talk f2f.

csweaver:
Summary of offline conversation: We will be using the filter as the param for memoization. I'm closing this and PR and starting over.

PR #307, opened 2018-10-10
Filter in engine

csweaver:
Move the filter responsibility to the engine instead of the rest command. This is in anticipation of caching the engine methods by filter instead of dataframe. There are two additional changes this implies (1) need to catch filter errors at the engine level, and (2) need to deal with interactive limits at the engine level.


server/app/rest_api/rest.py
```diff
@@ -598,18 +592,15 @@ def post(self):
...
-            return make_response("Non-interactive request", HTTPStatus.FORBIDDEN)
         # mode
         count = args.get("count", None)
         try:
-            diffexp = current_app.data.diffexp(df1, df2, count)
-        except ValueError as ve:
-            return make_response(ve.message, HTTPStatus.BAD_REQUEST)
+            diffexp = current_app.data.diffexp(set1_filter, set2_filter, count,
+                                               current_app.data.features["diffexp"]["interactiveLimit"])
+        except ValueError as e:
```
bkmartinjr: missing FilterError catch?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -114,18 +115,20 @@ def genes(self):
         return self.data.var.index.tolist()
 
     # Can't seem to cache a view of a dataframe, need to investigate why
```
bkmartinjr: Obsolete comment?

csweaver: yep

server/app/rest_api/rest.py
csweaver: good catch

server/app/driver/driver.py
```diff
@@ -65,10 +65,10 @@ def filter_dataframe(self, filter):
         pass
 
     @abstractmethod
-    def annotation(self, df, axis, fields=None):
+    def annotation(self, filter, axis, fields=None):
         """
         Gets annotation value for each observation
-        :param df: from filter_cells, dataframe
+        :param filter: filter: dictionary with filter params
```
ttung: probably should be:
`:param filter: dictionary with filter params`

also, maybe consider rewording to: dictionary with XXX as keys and YYY as the values.

server/app/driver/driver.py
```diff
@@ -77,36 +77,38 @@ def annotation(self, df, axis, fields=None):
...
         :param axis: string obs or var
         :return: {
             "var": list of variable ids,
             "obs": [cellid, var1 expression, var2 expression, ...],
         }
         """
-        pass
+    pass
```
ttung: intentional?

server/app/rest_api/rest.py
ttung: You can catch two exceptions at once like:

```
except (ValueError, FilterError) as ex:
    return make_response(ex.message, HTTPStatus.BAD_REQUEST)
```

server/app/rest_api/rest.py
```diff
@@ -248,7 +246,7 @@ class AnnotationsVarAPI(Resource):
     def get(self):
         fields = request.args.getlist("annotation-name", None)
         try:
-            annotation_response = current_app.data.annotation(current_app.data.data, "var", fields)
+            annotation_response = current_app.data.annotation(None, "var", fields)
```
ttung: You're passing in a `None` when your docblock calls this a dict.  Expected?

server/app/rest_api/rest.py
```diff
@@ -643,7 +636,7 @@ class LayoutObsAPI(Resource):
         }
     })
     def get(self):
-        return make_response((jsonify({"layout": current_app.data.layout(current_app.data.data)})), HTTPStatus.OK)
+        return make_response((jsonify({"layout": current_app.data.layout(None)})), HTTPStatus.OK)
```
ttung: You're passing in a `None` when your docblock calls this a dict.  Expected?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -200,15 +202,19 @@ def _filter_annotation(self, filter, index, axis):
...
+        :param filter: filter: dictionary with filter params
         :param axis: string obs or var
         :param fields: list of keys for annotation to return, returns all annotation values if not set.
         :return: dict: names - list of fields in order, data - list of lists or metadata
         [observation ids, val1, val2...]
         """
+        try:
+            df = self.filter_dataframe(filter)
+        except KeyError as e:
+            raise FilterError(f"Error parsing filter: {e}")
```
ttung: Use [exception chaining](https://www.python.org/dev/peps/pep-3134/)

i.e., `raise FilterError(f"Error parsing filter") from e`

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -219,16 +225,20 @@ def annotation(self, df, axis, fields=None):
...
         :param axis: string obs or var
         :return: {
             "var": list of variable ids,
             "obs": [cellid, var1 expression, var2 expression, ...],
         }
         """
+        try:
+            df = self.filter_dataframe(filter)
+        except KeyError as e:
+            raise FilterError(f"Error parsing filter: {e}")
```
ttung: chain exceptions.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -252,20 +262,31 @@ def data_frame(self, df, axis):
...
+        :param filter1: filter: dictionary with filter params for first set of observations
+        :param filter2: filter: dictionary with filter params for second set of observations
         :param top_n: Limit results to top N (Top var mode only)
+        :param interactive_limit: -- don't compute if total # genes in dataframes are larger than this
         :return: top genes, stats and expression values for variables
         """
+        try:
+            df1 = self.filter_dataframe(filter1)
+        except KeyError as e:
+            raise FilterError(f"Error parsing filter for set 1: {e}")
```
ttung: chain exceptions.

server/app/rest_api/rest.py
ttung: That is a valid approach.  It's also valid to just update the documentation to say that `None` == `{}`.

server/app/scanpy_engine/scanpy_engine.py
csweaver: will fix

server/app/driver/driver.py
csweaver: hmm, this is a little more complicated. I think I'm going to punt on this and add a ticket to make a filter object to simplify documentation in the long term.

PR #312, opened 2018-10-11
Add caching back in

csweaver:
Works now against filter instead of dataframe view

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -672,15 +672,13 @@ class LayoutObsAPI(Resource):
     })
     def post(self):
         try:
-            return make_response(
-                jsonify({
-                    "layout": current_app.data.layout(
-                        request.get_json()["filter"],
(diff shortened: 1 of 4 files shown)


server/app/rest_api/rest.py
```diff
@@ -674,15 +674,13 @@ def get(self):
...
-                        request.get_json()["filter"],
-                        interactive_limit=current_app.data.features["layout"]["obs"]["interactiveLimit"]
-                    )
-                }),
-                HTTPStatus.OK
-            )
+            filter = request.get_json()["filter"]
+            interactive_limit = current_app.data.features["layout"]["obs"]["interactiveLimit"]
+            layout = current_app.data.layout(filter, interactive_limit=interactive_limit)
+            # don't memoize layout with filter as is not idempotent
```
ttung: If that's the case, why not remove the @cache.memoize annotation?

PR #325, opened 2018-10-12
Launch browser to cellxgene

csweaver:
Also added  --no-launch command line parameter

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -1,6 +1,7 @@
 import argparse
 import os
 import warnings
+import webbrowser
 
 from flask import Flask
 from flask_caching import Cache
@@ -61,6 +62,8 @@ def run_scanpy(args):
         host = "0.0.0.0"
     else:
         host = "127.0.0.1"
+    if not args.no_launch:
+        webbrowser.open(f"http://{host}:{args.port}")
     app.run(host=host, debug=True, port=args.port)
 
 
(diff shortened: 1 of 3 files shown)


csweaver:
@freeman-lab What param name would you recommend for not launching the browser?

colinmegill:
A common test runner phrasing is 'headless'

csweaver:
@colinmegill I don't know if "headless" fits this scenario. I imagine it might be people trying out different parameters and not wanting the browser to launch every time. Though they could be running something just against the REST API.

@freeman-lab suggested not to open by default in the doc, thoughts?
> perhaps not by default, but when a `--open` or `--live` flag is set to true.

csweaver:
From Jeremy [open, no-open]. Default to open

server/app/app.py
```diff
@@ -69,6 +72,7 @@ def main():
     parser.add_argument("--title", "-t", help="Title to display -- if this is omitted the title will be the name "
                                               "of the directory from the data_directory arg")
     parser.add_argument("--port", help="Port to run server on.", type=int, default=5005)
+    parser.add_argument("--no-open", help="Do not launch the webbrowser.", action='store_true')
```
ttung: Often times, I find it helpful to reason parameters like this as:

`parser.add_argument("--no-open", help="Do not launch the webbrowser.", action="store_false", dest="open_browser")`

then your code would be:

```
if args.open_browser:
    ...
```

Rationale: you write the crufty inverted logic once for `parser.add_argument`, and the rest of code is much easier to read.  Since you only use `no_open` once, it's nbd.

I would fix the inconsistent string quote style though. :)

csweaver: That is much cleaner. 🤦‍♀️ on the string quote again...

PR #350, opened 2018-10-18
handle case of excessively large number of options in categorical met…

### LINKED ISSUE #349: Out of memory crash due to large metadata cardinality
If a data set contains an annotation (metadata) field with large cardinality, the application will attempt to display *every* possible category in the Categorical Metadata selection UI.   This may cause the browser to crash (out of memory) as a very large number of DOM elements are created.   Best case, it also hangs the UI for an unacceptably large period of time while the DOM is populated.

A better solution would be to cap the number of categories displayed for any given categorical metadata field (say ~30). 

^ @colinmegill - please add any thoughts and I can fix this once we agree on approach.

bkmartinjr:
This change caps the maximum number of categorical metadata values displayed in the categorical metadata selection UI (pick list of values).  It picks up to 100 values, and selects those that have the maximum number of cells annotated with that category value.    When the number of options is capped, an "... truncated list ..." annotation is added to the bottom of the selection list.

The '100' value is hard-coded, but could be parameterized if there is future need.

Fixes #349

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -5,18 +5,27 @@ import { connect } from "react-redux";
 
 import Category from "./category";
 
-@connect(state => {
-  const ranges = _.get(state.controls.world, "summary.obs", null);
-  return {
-    ranges
-  };
-})
-class Categories extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {};
+/* Cap the max number of displayed categories */
+const truncateCategories = options => {
+  const maxOptionsToDisplay = 100; // could be a property if helpful....
+  const numOptions = _.size(options);
+  if (numOptions <= maxOptionsToDisplay) {
+    return options;
   }
+  return _(options)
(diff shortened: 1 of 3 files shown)


client/src/components/categorical/categorical.js
```diff
@@ -5,18 +5,27 @@ import { connect } from "react-redux";
...
-    ranges
-  };
-})
-class Categories extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {};
+/* Cap the max number of displayed categories */
+const truncateCategories = options => {
+  const maxOptionsToDisplay = 100; // could be a property if helpful....
```
colinmegill: prefer this as a constant in globals, several others of the same kind are there

bkmartinjr: I put it in the component as it is very specific to the component (and not a "global" constant).    Globals should be, you know, global :-)

That said, I don't feel strongly and will make the change.

PR #356, opened 2018-10-21
categorical metadata performance work

bkmartinjr:
Changes are related to performance of categorical metadata selection UI:

* added memoization of the category option counting used to handle category-wide selection state (all selected, some selected, none selected)
* split annotation summarization into separate file within util/stateManager
* summarize Universe annotations - removes need to perform this operation on every world reset
* added summarization of variable annotations (and made corresponding changes to the tests
* add suppression of categorical metadata picker UI whenever total number of metadata options exceeds a configurable limit.  Configuration currently defaulted in front end to 1000, but plan is to pass this from the CLI so the user can control it.

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -23,11 +23,16 @@ const truncateCategories = options => {
 };
 
 @connect(state => ({
-  ranges: _.get(state.controls.world, "summary.obs", null)
+  ranges: _.get(state.controls.world, "summary.obs", null),
+  categorySelectionLimit: _.get(
+    state.config,
+    "parameters.category-selection-limit",
+    globals.configDefaults.parameters["category-selection-limit"]
+  )
 }))
 class Categories extends React.Component {
   render() {
-    const { ranges } = this.props;
+    const { ranges, categorySelectionLimit } = this.props;
     if (!ranges) return null;
 
     return (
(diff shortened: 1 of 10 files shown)


client/src/components/categorical/categorical.js
```diff
@@ -23,11 +23,16 @@ const truncateCategories = options => {
 };
 
 @connect(state => ({
-  ranges: _.get(state.controls.world, "summary.obs", null)
+  ranges: _.get(state.controls.world, "summary.obs", null),
+  categorySelectionLimit: _.get(
+    state.config,
+    "parameters.category-selection-limit",
+    globals.configDefaults.parameters["category-selection-limit"]
+  )
```
colinmegill: Recommend `categoryDisplayLimit` for clarity / ease of reading

client/src/components/categorical/categorical.js
```diff
@@ -42,7 +47,13 @@ class Categories extends React.Component {
         <p> Categorical Metadata </p>
         {_.map(ranges, (value, key) => {
           const isColorField = key.includes("color") || key.includes("Color");
-          if (value.options && !isColorField && key !== "name") {
+          const isSelectableCategory =
+            value.options &&
+            !isColorField &&
+            key !== "name" &&
+            value.numOptions < categorySelectionLimit;
```
colinmegill: `value.numOptions < categoryDisplayLimit`

client/src/components/categorical/categorical.js
bkmartinjr: the design intent is to have the same string as the CLI parameter, and command line params are dashed (see the CLI param proposal).   I think it will be far more confusing to add a mapping layer between the two namespaces.

client/src/components/categorical/categorical.js
bkmartinjr: I strongly prefer a model where we use the same name for the same concept.   Re-aliasing the command line param seems destined to confuse future readers of the code.

client/src/components/categorical/categorical.js
colinmegill: To clarify, no complaint about doing `category-display-limit`, only replacing the word `selection` with `display`, since that better describes the intention of the constant.

bkmartinjr:
For posterity, the variable name is a placeholder.   It will be replaced with the CLI param name once that is defined.   Issue #360.

PR #357, opened 2018-10-21
Docs:  add parameters to /config

bkmartinjr:
Proposed REST changes to add parameter communication from server to client.

### DIFF
--- a/docs/REST_API.md
+++ b/docs/REST_API.md
@@ -210,7 +210,7 @@ Information will include:
   - Limitations are optional, and specified as an additional parameter on a feature, encoded as a number value with key `interactiveLimit`. By convention, this is the maximum number of inputs at which point the request is likely to return an error indicating non-interactive compute request. For example, this might be the maximum number of observations specified in the request to `PUT /layout/obs`.
   - mandatory routes do not need to be specified
 - `displayNames` - names will include `engine` and `dataset` display names, as a JSON string. These should be relatively short strings (eg, suitable for window titles or equivalent)
-- `parameters` - system configuration parameters, specified as a key/value pair in a simple object.
+- `parameters` - system configuration parameters, specified as a key/value pair in a simple object. See [Currently Defined Parameters](#currently-defined-parameters) for additional information.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/REST_API.md
```diff
@@ -230,6 +233,12 @@ GET /config
       "engine": "ScanPy version 1.33",
       "dataset": "/home/joe/mouse/blorth.csv"
     },
+    "parameters": {
+      "verbose": false,
+      "log-level": "info",
+      "minimumThreshold", 99
```
csweaver: why is this minimum and not maximum?

docs/REST_API.md
```diff
@@ -230,6 +233,12 @@ GET /config
       "engine": "ScanPy version 1.33",
       "dataset": "/home/joe/mouse/blorth.csv"
     },
+    "parameters": {
+      "verbose": false,
+      "log-level": "info",
```
csweaver: Where are verbose and log-level coming from? Is this the `--debug` param? If so, what are our log-levels? We should probably define them.

bkmartinjr: they are just examples - intended to show what _might_ be.

you are right that when we add our first parameter, we need to document it.  Probably need to add a section that describes  each.   Given that we do not yet have any, I made up some examples.

docs/REST_API.md
bkmartinjr: just another random example...could be anything.

docs/REST_API.md
bkmartinjr: BTW, we could hold this PR (and add to it) once we have our first set of CLI params defined.

csweaver: Yeah, let's do that. The doc looks good outside of the example params. Technically they are fine, but I'm worried they would be confusing to a new contributor who thinks they are required.

csweaver:
If there are no parameters to pass through, is there still an empty "parameters" key? 
```
                "parameters" : { }              
```

bkmartinjr:
Yes, it should be an empty parameters object if no params

PR #364, opened 2018-10-23
add CLI tool for dataset preparation using scanpy

freeman-lab:
This PR adds a `prepare` CLI tool that prepares and preprocesses datasets for `cellxgene` using `scanpy`. It does some basic checks, formatting, and computation, but tries to be relatively unopinionated. For now, it's inside the `prepare` subfolder, but eventually we can merge it into a subcommand of the main CLI.

Here's an example call

```
python prepare/cli.py pbmc3k/raw/ --output=pbmc3k_processed.h5ad --layout=umap
```

Here's the CLI output for that call

``` 
[cellxgene] Starting CLI...
[cellxgene] Loading data from pbmc3k/raw/, please wait...
[cellxgene] Beginning preprocessing...
[cellxgene] Progress  [####################################]  100%
[cellxgene] Saving results to pbmc3k_processed.h5ad...
[cellxgene] Success!
```
 
And here's the help showing the full set of options

``` 
Usage: cli.py [OPTIONS] <dataset: file or path to data>

  preprocesses data for use with cellxgene

Options:
  --layout [umap|tsne|umap+tsne]  layout algorithm  [default: umap]
  --recipe [none|seurat|zheng17]  preprocessing to run  [default: none]
  --output TEXT                   save a new file to filename
  --set-obs-names TEXT            named field to set as index for obs
  --set-var-names TEXT            named field to set as index for var
  --make-obs-names-unique         ensure obs index is unique  [default: True]
  --make-var-names-unique         ensure var index is unique  [default: True]
  --sparse                        whether to force sparsity  [default: False]
  --overwriting                   allow file overwriting  [default: False]
  --plotting                      whether to generate plots  [default: False]
  --help                          Show this message and exit.
```

I used the library `click` and did a few things to make the UX nice (at least, nice to me!) that we might want to consider propagating to the main CLI (e.g. status messages, progress bar). But opinions may differ.

Comments / suggested changes welcome.


prepare/cli.py
```diff
@@ -0,0 +1,160 @@
+import click
```
csweaver: add click to requirements.txt

prepare/cli.py
```diff
@@ -0,0 +1,160 @@
...
+from numpy import unique, ndarray
+from scipy.sparse.csc import csc_matrix
+from os.path import isfile, isdir, splitext, expanduser
+
+settings = dict(help_option_names=['-h', '--help'])
+
+
+@click.command()
+@click.argument('dataset', nargs=1, metavar='<dataset: file or path to data>', required=True)
+@click.option('--layout', default='umap', type=click.Choice(['umap', 'tsne', 'umap+tsne']),
```
csweaver: could layout take more than one option instead of the umap+tsne? Then we could loop though? This solves a possible permutation issue if we have more than two layout options.

prepare/cli.py
```diff
@@ -0,0 +1,160 @@
...
+              help='layout algorithm', show_default=True)
+@click.option('--recipe', default='none', type=click.Choice(['none', 'seurat', 'zheng17']),
+              help='preprocessing to run', show_default=True)
+@click.option('--output', default='', help='save a new file to filename')
+@click.option('--set-obs-names', default='', help='named field to set as index for obs')
+@click.option('--set-var-names', default='', help='named field to set as index for var')
+@click.option('--make-obs-names-unique', default=True, is_flag=True, help='ensure obs index is unique', show_default=True)
+@click.option('--make-var-names-unique', default=True, is_flag=True, help='ensure var index is unique', show_default=True)
+@click.option('--sparse', default=False, is_flag=True, help='whether to force sparsity', show_default=True)
+@click.option('--overwriting', default=False, is_flag=True, help='allow file overwriting', show_default=True)
```
csweaver: `help=whether to allow file overwriting`  to match other flags

prepare/cli.py
```diff
@@ -0,0 +1,160 @@
...
+    sc.settings.verbosity = 0
+    sc.settings.autosave = True
+
+    # check args
+    if sparse and not recipe == 'none':
+        raise click.UsageError('Cannot use a recipe when forcing sparsity')
+
+    output = expanduser(output)
+    if isfile(output) and not overwrite:
+        raise click.UsageError('Cannot overwwrite existing file %s, try using the flag --overwrite' % output)
```
csweaver: overwwrite -> overwrite

prepare/cli.py
```diff
@@ -0,0 +1,160 @@
...
+            name, extension = splitext(dataset)
+            if extension == '.h5ad':
+                adata = sc.read_h5ad(dataset)
+            elif extension == '.loom':
+                adata = sc.read_loom(dataset)
+            else:
+                raise click.FileError(dataset, hint='does not have a valid extension [.h5ad | .loom]')
+        elif isdir(dataset):
+            if not dataset.endswith('/'):
+                raise click.FileError(dataset, hint='a path must end with a /')
```
csweaver: Why are we forcing this?

prepare/cli.py
```diff
@@ -0,0 +1,160 @@
...
+            sc.tl.umap(adata)
+            if plotting:
+                sc.pl.umap(adata, color='louvain', palette=palette, save='_louvain')
+
+        if layout == 'tsne' or layout == 'umap+tsne':
+            sc.tl.tsne(adata)
+            if plotting:
+                sc.pl.tsne(adata, color='louvain', palette=palette, save='_louvain')
+
+    def show_step(item):
```
csweaver: I like this!

prepare/cli.py
freeman-lab: will fix by automatically appending a separator

prepare/cli.py
freeman-lab: fixed

prepare/cli.py
freeman-lab: fixed

prepare/cli.py
freeman-lab: great catch, changed the argument type to handle this

prepare/cli.py
freeman-lab: done

prepare/cli.py
freeman-lab: thanks!

Issue #355, opened 2018-10-21
CLI: /data/obs fetch of expression data is very slow on immune_bone_marrow dataset


bkmartinjr:
A simple fetch of a expression data via the REST API is very slow on the immune_bone_marrow dataset.   The bottleneck seems to be in the CLI processing - it is definitely not in the data transfer.

If I make a request twice, it takes 90 seconds to fetch the result.  Second attempt takes 2 seconds (essentially the data transfer time).

Given the trivial nature of the data filtering (eg, return an entire column for the dataframe of 378K rows), this seems extremely slow.

```
# curl -X PUT http://localhost:5005/api/v0.2/data/obs -H "Content-Type: application/json" -d '{"filter":{"var":{"annotation_value":[{"name":"gene_names","values":["RP11-34P13.7"]}]}}}' >
 /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 14.6M  100 14.6M  100    89   290k      1  0:01:29  0:00:51  0:00:38 3496k
root@woof:/mnt/c/Users/bruce/projects/cellxgene/client# curl -X PUT http://localhost:5005/api/v0.2/data/obs -H "Content-Type: application/json" -d '{"filter":{"var":{"annotation_value":[{"name":"gene_names","values":["RP11-34P13.7"]}]}}}' > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 14.6M  100 14.6M  100    89  6919k     40  0:00:02  0:00:02 --:--:-- 6922k
```

I get a very similar result when using the GET interface, eg., `/data/obs?var:name=3`

Digging into the CLI code a bit, it looks like almost all of the time is spent in `scanpy_engine.py::filter_dataframe()`, in these two lines (near bottom of function):

```
        cell_data = self.data[cells_idx, :]
        data = cell_data[:, genes_idx]
```

I'm not sure if this is due to the way we are slicing, or the nature of the data, but it causes quite a noticeable delay in the UI.

@freeman-lab - the above REST request is simply looking for the observation-major data, for all observations, for the column where the annotation gene_names=='RP11-34P1.7'.

Issue #348, opened 2018-10-18
Make diffexp calculation sparse matrix safe


csweaver:
Scipy does not work properly for stats calculations with sparse matrices. Scanpy is supposed to have some diffexp calcualtion tools available, we should use those instead.

Issue #347, opened 2018-10-18
pbmc3k test data set fails diffexp calculation


bkmartinjr:
Load the latest pbmc3k "prepared" data set, ask for diffexp calculation from the f/e, and you get back bogus JSON from the CLI (contains NaN's which are not legal in JSON).

This looks like a bug in the diffexp CLI calculation.

```
[
  [
    22, 
    0.0, 
    NaN, 
    NaN, 
    -0.03223249316215515, 
    -0.03223249316215515
  ], 
  [
    59, 
    3.725290298461914e-09, 
    1.5152738737504437e-91, 
    0.0, 
    -0.03832865506410599, 
    -0.038328658789396286
  ], 
  [
    79, 
    2.177830457687378, 
    2.1666501748703266e-165, 
    0.0, 
    1.6735427379608154, 
    -0.5042877197265625
  ], 
  [
    242, 
    3.2382142543792725, 
    3.0608739607110622e-263, 
    0.0, 
    2.9571640491485596, 
    -0.2810501456260681
  ], 
  [
    283, 
    2.637721061706543, 
    1.900322634622438e-299, 
    0.0, 
    2.224553108215332, 
    -0.41316789388656616
  ], 
  [
    523, 
    2.9486725330352783, 
    5.912974536818484e-257, 
    0.0, 
    2.600834846496582, 
    -0.3478377163410187
  ], 
  [
    666, 
    2.5547847747802734, 
    2.679752079590864e-155, 
    0.0, 
    2.289745807647705, 
    -0.26503893733024597
  ], 
  [
    701, 
    0.0, 
    NaN, 
    NaN, 
    -0.0792982429265976, 
    -0.0792982429265976
  ], 
  [
    802, 
    2.4982500076293945, 
    1.1483218184552997e-120, 
    0.0, 
    2.3370797634124756, 
    -0.16117022931575775
  ], 
  [
    884, 
    2.120547294616699, 
    2.907435060589997e-184, 
    0.0, 
    1.6300982236862183, 
    -0.4904491603374481
  ]
]
```

bkmartinjr:
More info - the CLI stdout/stderr contains:

```
# cellxgene scanpy ../cellxgene-datasets/cellxgene-prepare/pbmc3
k/
[...]

PR #371, opened 2018-10-24
add --obs-names and --var-names CLI params

### LINKED ISSUE #309: De-hack uns for single gene
data.uns = cell_data.uns fails if there is a single gene. Need to fix.

bkmartinjr:
Changes:
* add CLI params: --obs-names, --var-names
* name aliasing in scanpy engine wrapper, in support of --obs-names and --var-names
* reworked annotation indexing to improve startup performance
* implement slicing that is aware of col vs row performance preferences
* remove redundant need to `reset_index()` when creating REST response, by using to_records()
* remove dead code (genes(), cells()) in the driver.py framework
* updated scanpy_engine tests to allow for either list or tuple return values
* correctly handle sparse data types for AnnData::X (must call toarray() to get an NP array)
* update required AnnData version to incorporate fix for https://github.com/theislab/scanpy/issues/332
* Change driver framework __init__ args to just pass entire args dict

Fixes #309 #355 #347 #348

### DIFF
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -23,13 +23,41 @@ Sort order for methods
 
 class ScanpyEngine(CXGDriver):
 
-    def __init__(self, data, layout_method=None, diffexp_method=None):
+    def __init__(self, data, layout_method=None, diffexp_method=None, obs_names=None, var_names=None):
         super().__init__(data, layout_method=layout_method, diffexp_method=diffexp_method)
-        self._validatate_data_types()
-        self._add_mandatory_annotations()
+        self._alias_annotation_names(obs_names, var_names)
+        self._validate_data_types()
         self.cell_count = self.data.shape[0]
         self.gene_count = self.data.shape[1]
         self._create_schema()
+
+    def _alias_annotation_names(self, obs_names, var_names):
+        """
+        Do all user-specified annotaiton aliasing.
+        As a *critical* side-effect, ensure the indices are simple number ranges
+        """
+        aliased_names = {'obs': obs_names, 'var': var_names}
+        for ax in Axis:
+            name = aliased_names[ax]
+            if name == 'name':
+                continue
+
+            ax_name = str(ax)
+            df_axis = getattr(self.data, ax_name)
+            if name is None:
+                # reset index to simple range; alias 'name' to point at the
+                # previously specified index.
+                df_axis = df_axis.reset_index().rename(columns={'index': 'name'})
+            elif name in df_axis.columns:
+                if name not in df_axis.columns:
+                    raise KeyError(f"Annotation name {name}, specified in --{ax_name}-name does not exist.")
+                if not df_axis[name].is_unique:
+                    raise KeyError(f"Values in -{ax_name}-name must be unique. Please prepare data to contain unique values.")
+                # reset index to simple range; alias user-specified annotation to 'name'
+                df_axis = df_axis.reset_index(drop=True).rename(columns={name: 'name'})
+            else:
+                raise KeyError(f"Annotation name {name}, specified in --{ax_name}_name does not exist.")
+            setattr(self.data, ax_name, df_axis)
 
     def _create_schema(self):
         self.schema = {
@@ -71,6 +99,9 @@ class ScanpyEngine(CXGDriver):
                                   help="Algorithm to use for graph layout")
         scanpy_group.add_argument("-d", "--diffexp", choices=["ttest"], default="ttest",
                                   help="Algorithm to used to calculate differential expression")
+        scanpy_group.add_argument("--obs-names",
+                                  help="Annotation name to use as unique, human-readable observation name")
+        scanpy_group.add_argument("--var-names", help="Annotation name to use as unique, human-readable variable name")
         scanpy_group.add_argument("data_directory", metavar="dir", help="Directory containing data and schema file")
         scanpy_group.set_defaults(func=invocation_function)
         return scanpy_group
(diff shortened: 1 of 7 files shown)


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -93,15 +124,7 @@ def _nan_to_one(values):
...
-    def _add_mandatory_annotations(self):
-        # ensure gene
-        self.data.var["name"] = Series(list(self.data.var.index), dtype="unicode_", index=self.data.var.index)
-        self.data.var.index = Series(list(range(self.data.var.shape[0])), dtype="category")
-        # ensure cell name
-        self.data.obs["name"] = Series(list(self.data.obs.index), dtype="unicode_", index=self.data.obs.index)
-        self.data.obs.index = Series(list(range(self.data.obs.shape[0])), dtype="category")
-
-    def _validatate_data_types(self):
+    def _validate_data_types(self):
```
csweaver: 🤦‍♀️

freeman-lab:
looked over the differential expression calculations, after reviewing behavior on some example data, i have two proposed changes:

- sort on the t-test `statistic` (from large to small) rather than the `pval` (from small to large), so, replace `sort_order = np.argsort(np.abs(pval))` with  `sort_order = np.argsort(np.abs(diffexp_result.statistic))[::-1]`

- add a method to replace all `NaN`s in `diffexp_result.statistic` with 0 (otherwise the above may have undesirable behavior when there are many `NaN`s)

Rationale: although the `pval` and the t-test `statistic` are monotonically related (via the CDF of the t-distribution), we may encounter cases where a large number of the lowest p-values are exactly 0, so the sort order becomes arbitrary (for the top N for small N), whereas sorting by the `statistic` will be meaningful

based on several examples, this shouldn't make a dramatic difference, but feels more robust

and we can reevaluate all of this once we are using `scanpy` built-in differential expression methods

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -202,6 +215,34 @@ def _filter_annotation(self, filter, index, axis):
                     index = np.logical_and(index, key_idx)
         return index
 
+    def _slice(self, data, obs_selector=None, vars_selector=None):
```
csweaver: This can be a static method

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -298,14 +331,17 @@ def diffexp(self, filter1, filter2, top_n=None, interactive_limit=None):
...
         pval = self._nan_to_one(diffexp_result.pvalue)
         bonferroni_pval = 1 - (1 - pval) ** self.gene_count
-        ave_exp_set1 = np.mean(df1.X, axis=0)
-        ave_exp_set2 = np.mean(df2.X, axis=0)
+        ave_exp_set1 = np.mean(X1, axis=0)
+        ave_exp_set2 = np.mean(X2, axis=0)
         ave_diff = ave_exp_set1 - ave_exp_set2
         if mode == DiffExpMode.TOP_N:
-            sort_order = np.argsort(np.abs(diffexp_result.statistic))[::-1]
+            sort_order = np.argsort(np.abs(pval))
```
csweaver: Why sort by p-val instead of statistic?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -219,11 +260,11 @@ def annotation(self, filter, axis, fields=None):
         df_axis = getattr(df, axis)
         if not fields:
             fields = df_axis.columns.tolist()
-        annotations = DataFrame(df_axis[fields], index=df_axis.index)
-        return {
+        result = {
             "names": fields,
-            "data": annotations.reset_index().values.tolist()
+            "data": DataFrame(df_axis[fields]).to_records(index=True).tolist()
```
csweaver: Nice find, this actually handles the tiny floats correctly

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -237,28 +278,20 @@ def data_frame(self, filter, axis):
...
-        var_idx = df.var.index.tolist()
-        obs_idx = df.obs.index.tolist()
-        values = df.X
-        df_shape = df.shape
-        if df_shape[0] == 1:
-            values = values[None, :]
-        elif df_shape[1] == 1:
-            values = values[:, None]
+        # convert sparse slice to dense
+        X = slice._X.toarray() if sparse.issparse(slice._X) else slice._X
```
csweaver: I tend to avoid the one line if statements in python, but that's probably just me.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -22,17 +22,48 @@
...
-        self._add_mandatory_annotations()
+        self._alias_annotation_names(obs_names, var_names)
+        self._validate_data_types()
         self.cell_count = self.data.shape[0]
         self.gene_count = self.data.shape[1]
         self.layout_options = ["umap", "tsne"]
         self.diffexp_options = ["ttest"]
         self._create_schema()
 
+    def _alias_annotation_names(self, obs_names, var_names):
```
csweaver: As discussed, change the signature to make what the name params are clearer.

csweaver:
Thank you, this is fantastic. The static method, refactoring the _alias_annotation_names signature, and the changes Jeremy requested are the only blockers. 

I can't believe you found a reasonable way to convert the scanpy data to JSON. 🎆 🍾

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: done

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: it was a request from @freeman-lab, but he has since reverted that to use t-statistic.   I fixed the issue with NaNs in that stat.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: Unless you feel strong, I think this is actually preferable in this case, and follows the ScanPy/AnnData examples.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: done

PR #377, opened 2018-10-24
max categories

csweaver:
Hooked up the --max-categories flag and passing it through to config.


server/app/app.py
```diff
@@ -105,6 +105,7 @@ def create_cli():
                               dest="open_browser")
     launch_group.add_argument(
         "--max-categories",
+        type=int,
```
bkmartinjr: must be greater than zero

PR #385, opened 2018-10-26
CLI renaming and phrasing

csweaver:
Picked the low hanging fruit from the CLI UX review
* category-selection-limit -> max-category-items
* Indicate load may taking a long time
* program -> command (for launch, prepare)
* debug -> verbose
* flask-debug -> debug
    * made debug into developer mode +verbose -open
* move examples for running launch command up higher in help

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -58,7 +58,7 @@ description:
 cellxgene is a local web application for exploring single cell expression.
     """
     parser.add_argument("-V", "--version", help="show version and exit")
-    subparsers = parser.add_subparsers(dest="program")
+    subparsers = parser.add_subparsers(dest="command")
     subparsers.required = True
     launch_group = subparsers.add_parser("launch", help="launch web application",
                                          formatter_class=argparse.RawTextHelpFormatter)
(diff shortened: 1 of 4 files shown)


server/app/app.py
```diff
@@ -100,16 +99,16 @@ def create_cli():
...
                               help="more verbose output, including outputting warnings and every REST request")
-    launch_group.add_argument("--flask-debug", action="store_true", help=argparse.SUPPRESS)
+    launch_group.add_argument("--debug", action="store_true", help=argparse.SUPPRESS)
     launch_group.add_argument("--no-open", help="do not launch the webbrowser", action="store_false",
                               dest="open_browser")
     launch_group.add_argument(
-        "--category-selection-limit",
+        "--max-category-items",
         type=whole_number,
         help="maximum number of categories to display on the front-end. "
```
bkmartinjr: this is not accurate - it does not control the number of categories.

server/app/app.py
```diff
@@ -136,24 +135,29 @@ def run_scanpy(args):
...
         CXG_API_BASE=api_base
     )
-    if not args.debug:
+
+    if not args.verbose:
         log = logging.getLogger('werkzeug')
         log.setLevel(logging.ERROR)
     from .scanpy_engine.scanpy_engine import ScanpyEngine
-    print(f"Loading data from {args.data}")
+    print(f"Loading data from {args.data} (this may take a while for large datasets)")
```
bkmartinjr: I would just say "this may take a while", as some users have slow computers and "large datasets" is a very relative thing....

server/app/app.py
csweaver: Should I use the verbiage from the doc? 
> specifies a numeric limit for the cardinality of a categorical annotation, beyond which the annotation will not be available for user selection in the front-end

server/app/app.py
csweaver: 👍

PR #386, opened 2018-10-26
Speed up help

csweaver:
I moved the computational args back to app.py. While these are scanpy specific we probably shouldn't require importing scanpy and all the packages that come along with it just to figure out which layout options we have. This speeds up time to get help immensely.

@bkmartinjr you can put your obs/var-name options directly in app.py for now


freeman-lab:
looks good to me! nice that it was such an easy change

csweaver:
@bkmartinjr merging now to let @freeman-lab develop over the weekend. If you want changes you can still do a review and we can perform some git surgery on Monday

PR #401, opened 2018-11-02
integrate prepare cli

freeman-lab:
This PR just moves the `prepare` CLI into the main CLI as a subcommand, alongside `launch`. I also fixed a few formatting issues and ensured consistency with `launch`.


csweaver:
Love to have some tests for the prepare command at some point

PR #416, opened 2018-11-07
Prepare fixes

csweaver:
fixed minor issues I encountered when running the prepare command

1. `cellxgene prepare <file>` does nothing because no file is saved. I added an error for this case
2. There are some hidden requirements that I added to the requirements.txt file -- louvain and python-igraph

### DIFF
--- a/server/cli/prepare.py
+++ b/server/cli/prepare.py
@@ -1,9 +1,4 @@
-<<<<<<< HEAD
 from os.path import expanduser, isdir, isfile, sep, splitext
-=======
-import click
-import warnings
->>>>>>> 0395f679... no saved output error -> warning
 
 import click
 from numpy import ndarray, unique
--- a/server/cli/prepare.py
+++ b/server/cli/prepare.py
@@ -118,7 +118,12 @@ def prepare(data, layout, recipe, output, plotting, sparse, overwrite,
         sc.pp.neighbors(adata)
 
     def run_louvain(adata):
-        sc.tl.louvain(adata)
+        try:
+            sc.tl.louvain(adata)
+        except ModuleNotFoundError:
+            click.echo("\nWarning: louvain module is not installed, no clusters will be calculated. "
+                       "To fix this please install cellxgene with the optional feature louvain enabled: "
+                       "`pip install cellxgene[louvain]`")
 
     def run_layout(adata):
         if len(unique(adata.obs["louvain"].values)) < 10:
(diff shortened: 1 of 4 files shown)


freeman-lab:
cool @csweaver ! so the "error" you're flagging was actually intended behavior :) a common use case for me, especially when comparing preprocessing choices, was to save plots (with the `--plotting` option) but not bother saving a file

i'd be fine with either
1) change this error you've added into a warning
2) making the _default_ for output be writing a file with the same name as the original with `prepared` appended, and still let you specify `--no-output` to prevent file writing

in other words, i want to preserve the ability to run `prepare` and not write a file, but either of the above would seem to preserve that ability and address your concern (running and accidentally forgetting to specify an output)

i think my preference is (1)

csweaver:
@freeman-lab a warning is fine! I just was really confused when I ran it and I kept getting the same errors about cells on the frontend. I had to dig into the code to figure out I just wasn't writing a file at all.

freeman-lab:
totally get that! let's do a warning, perfect

server/cli/prepare.py
```diff
@@ -112,7 +117,12 @@ def run_neighbors(adata):
         sc.pp.neighbors(adata)
 
     def run_louvain(adata):
-        sc.tl.louvain(adata)
+        try:
+            sc.tl.louvain(adata)
+        except ModuleNotFoundError:
+            click.echo("\nWarning: louvain module is not installed, no clusters will be calculated. "
```
csweaver: This interrupts the click progress bar so I included a new line so that the warning is on its own line and not after the progress bar

csweaver:
Added one more fix before we merge this puppy. --overwrite was not hooked up. I added the code to overwrite the original file as well as a warning if the overwrite flag was selected on loom data.

freeman-lab:
just to capture an offline chat with @csweaver  re: the overwrite changes here: the _intended_ behavior was for `--overwrite` to act as a flag that allows overwriting if you specify an output file with the same name as the original file, not to actually trigger overwriting. 

Rationale: I erred on the side of making overwriting quite difficult. You always need to specify an output  file if you want to save, and if it's going to overwrite, you need to explicitly specify that. My worry about the behavior proposed here is that it makes overwriting too easy.

Maybe it'd be better to rename the flag to `--allow-overwrite`?

Open to other opinions! Could also keep the current behavior for now and wait till we do a proper UX review.

PR #417, opened 2018-11-07
update package dependencies

bkmartinjr:
update all client-side package dependencies _except_ for d3.

note: we still have the merge security vulnerability, but it is only an issue for dev dependencies, and will be resolved when jest fixes their dependencies.


colinmegill:
Ran the install, QA, nothing broken, seems reasonable to me!

PR #421, opened 2018-11-09
fix anaconda build

csweaver:


freeman-lab:
This looks right to me! Let's just remember that this might resurface if we for some other reason in the future we need to import scanpy prior to this point.

server/cli/launch.py
```diff
@@ -82,6 +82,9 @@ def launch(data, layout, diffexp, title, verbose, debug, obs_names, var_names,
 
     click.echo(f"[cellxgene] Loading data from {basename(data)}, this may take awhile...")
 
+    # Fix for anaconda python. Allows python to not be installed as a framework
```
bkmartinjr: fix looks OK, but comment is a bit opaque (to me)

PR #426, opened 2018-11-09
bump-update

csweaver:
Add regexs to ensure we only update our own package number, not dependencies.


csweaver:
Merging without review so I can work on build process. Will happily take post-merge feedback

PR #449, opened 2018-11-16
Change color scales to rainbow

colinmegill:
Rainbow color scales for categorical, and warm rainbow range for continuous

continuous

image

categorical 

image

### DIFF
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -1,11 +1,13 @@
 // jshint esversion: 6
 import { connect } from "react-redux";
 import React from "react";
+import _ from "lodash";
 
 @connect(state => ({
   categoricalAsBooleansMap: state.controls.categoricalAsBooleansMap,
   colorScale: state.controls.colorScale,
-  colorAccessor: state.controls.colorAccessor
+  colorAccessor: state.controls.colorAccessor,
+  schema: _.get(state.controls.world, "schema", null)
 }))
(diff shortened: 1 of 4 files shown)


client/src/middleware/updateCellColors.js
```diff
@@ -59,11 +65,19 @@ const updateCellColorsMiddleware = store => next => action => {
   */
 
   if (action.type === "color by categorical metadata") {
-    colorScale = d3.scaleOrdinal().range(globals.ordinalColors);
+    // colorScale = d3.scaleOrdinal().range(globals.ordinalColors);
```
bkmartinjr: cleanup?

client/src/components/categorical/value.js
```diff
@@ -34,14 +36,23 @@ class CategoryValue extends React.Component {
...
 
     if (!categoricalAsBooleansMap) return null;
 
     const selected = categoricalAsBooleansMap[metadataField][value];
     /* this is the color scale, so add swatches below */
     const c = metadataField === colorAccessor;
+    let categories = null;
+
+    if (c && schema) {
+      /* lint wants [categories] which seems less clear, prefer [0] over array destructuring */
```
bkmartinjr: we should configure eslint to what we want and then roll it out globally (rather than letting it be ad hoc).  I personally am OK with either, with a very slight preference for destructuring.  But I'm OK if you want to standardize on not using it.

PR #453, opened 2018-11-16
Use externally-reachable image urls

mckinsel:
This is needed for the images to show up in pypi.


freeman-lab:
lgtm! super excited to have this fixed!

PR #465, opened 2018-11-21
improve type handling for non-string annotation data

### LINKED ISSUE #457: Issue with launching cellxgene when anndata object has boolean metadata
cellxgene_error

Issue with launching cellxgene when anndata object has boolean metadata

bkmartinjr:
Fixes #457 

Several improvements to annotation/metadata handling:
* revised front-end categorical metadata (components and reducers) to not assume all annotation is composed of string data.   It will now safely handle native types (bool, string, number, etc).
* improved the type inferencing in the ScanPy engine so that it correctly handles bools, and does not mistakenly coerce between incompatible numpy types.

While this fixes #457, there is still residual issues with how we summarize metadata in the front-end (specfically, `world.summary` assumes all annotation/metadata values can be safely converted to Object lables/strings without losing information).   There will be another PR to fix that.

### DIFF
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -2,29 +2,15 @@ import React from "react";
 import _ from "lodash";
 import { connect } from "react-redux";
 import { FaChevronRight, FaChevronDown } from "react-icons/fa";
-import memoize from "memoize-one";
-import { Button, Tooltip, Position } from "@blueprintjs/core";
+import { Button, Tooltip } from "@blueprintjs/core";
 
 import * as globals from "../../globals";
 import Value from "./value";
 import alphabeticallySortedValues from "./util";
 
-const countCategories = (values, optsAsBools) =>
-  _.reduce(
-    values,
-    (r, v, k) => {
-      r.total += 1;
-      if (optsAsBools[k]) {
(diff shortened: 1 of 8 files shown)


client/src/components/categorical/category.js
```diff
@@ -33,17 +19,15 @@ class Category extends React.Component {
...
-    const { categoricalAsBooleansMap, metadataField, values } = this.props;
-    const categoryCount = this.countCategories(
-      values,
-      categoricalAsBooleansMap[metadataField]
-    );
+    const { categoricalSelectionState, metadataField } = this.props;
+    const cat = categoricalSelectionState[metadataField];
+    const categoryCount = {
+      total: cat.numOptions,
+      on: _.reduce(cat.optionSelected, (res, cond) => (cond ? res + 1 : res), 0)
```
colinmegill: Hmm, feeling `on` might be a little bit obfuscated. Does it mean 'currentlySelected', or is it related to a handler for an event? Room for a better name or a comment here. Obvious what it's doing here after spending a minute reading the ternary, so maybe it's fine.

bkmartinjr: will improve the clarity

PR #467, opened 2018-11-26
Fix readme images on PyPi

mckinsel:
But this time actually do that


freeman-lab:
i think this works! went to one of the test URLs in a new window and the gif loads correctly, and paths for all  of them look right to me

csweaver:
I can make a new test.pypi release for this if you'd like

mckinsel:
Yeah that'd be great.

PR #469, opened 2018-11-26
Add keyring to dev requirements

csweaver:
Allows you to store pypi user/pass securely so you don't have to look it up every time you release


csweaver:
I think @freeman-lab is right. I'm going to close this.

PR #481, opened 2018-11-29
External host option

### LINKED ISSUE #440: Aplication loads properly but no data is shown in the screen
I have two installations of cellxgene. One works properly and the other loads the data and starts the server but does not show it. All layout elements are loaded properly but no dots, histograms, etc are shown.

I tried to add the verbose argument but I do not get any error or warning message. 

Do you know how to debug this?

csweaver:
Since we are sending the front-end the api base, simply listening to all ports on the backend (0.0.0.0) wasn't working for hosting cellxgene externally. I've replaced the --listen-all option with a --host option. And then send flask either localhost if --host isn't specified or 0.0.0.0 if it is. The host option works with an IP or with a host url.

Fixes #440

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -24,14 +24,13 @@ from server.app.util.errors import ScanpyFileError
 @click.option("--port", "-p", help="Port to run server on.", metavar="", default=5005, show_default=True)
 @click.option("--obs-names", default=None, metavar="", help="Name of annotation field to use for observations.")
 @click.option("--var-names", default=None, metavar="", help="Name of annotation to use for variables.")
-@click.option("--listen-all", is_flag=True, default=False, show_default=True,
-              help="Bind to all interfaces (this makes the server accessible beyond this computer).")
+@click.option("--host", default=None, help="Host name or ip (for serving externally accessible instance).")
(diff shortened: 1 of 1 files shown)


csweaver:
I've tested this on an EC2 instance -- it's pretty cool!

server/cli/launch.py
```diff
@@ -65,10 +64,12 @@ def launch(data, layout, diffexp, title, verbose, debug, obs_names, var_names,
         file_parts = splitext(basename(data))
         title = file_parts[0]
 
-    if listen_all:
-        host = "0.0.0.0"
+    if host:
+        host = host
+        flask_host = "0.0.0.0"
```
ttung: if the user is expected to provide the local ip address, you might as well bind only to that ip address (caveat: NAT is the exception, but if the user is behind a NAT, they have a lot of other hoops to jump through).

bkmartinjr: agree...

bkmartinjr:
@ttung - it would be ideal if we can preserve the ability to serve the JS at a different base URL than the data REST API.    At least in a debugging mode, as we use it quite extensively for front-end development.

csweaver:
Limited to only host ip and was able to remove some code in the process ^_^ by having the default for the --host be localhost.

Issue #483, opened 2018-11-29
Show summary stats for each gene in dataset


bkmartinjr:
The new fuzzy gene name search attempts to access a variable annotation called `n_counts`, and assumes that this field contains some sort of  metric related to the gene (count == cells with expression above some baseline?).

Currently, the meaning and semantics of any annotation field (other than `name`) is completely dependent on the data set, and should not be used in this way.   Ie, the current server API has no concept of summary metrics for genes.

I've commented out the code and am filing this issue so @colinmegill is aware and can take any necessary actions.

Code in question, from `components/geneExpression/index.js`
```
    <MenuItem
      active={modifiers.active}
      disabled={modifiers.disabled}
      label={gene.n_counts}
      key={gene.name}
      onClick={g => {
        /* this fires when user clicks a menu item */
        handleClick(g);
      }}
      text={text}
    />
```

bkmartinjr:
@neuromusic @colinmegill - this issue was "FYI" in case Colin wanted to fix it in some other way.  If not, we should close it.

colinmegill:
It's been a long running thread to allow users to explore genes / see the
shape of that space at a high level. Probably the biggest omission in the
interface re: allowing user to see shape of dataset, and one I'd like to
eventually correct. It would make sense to send summary information ie.,
`[[gene_name: count], ...] `. If we feel like that's an API we want to
support, we could use that info to fill this in.

I'm for adding this to initialize, I think it's a good step.

bkmartinjr:
To add this feature, we would need to come up with a way to count obs/cells with the gene, which requires having an opinion about what constitutes expressing the gene.  That is a bit of a step into having an opinion about how to interpret the data (similar to the current half-baked differential expression feature).

@neuromusic - maybe this need should be included in your efforts to think through marker gene exploration.

@GenevieveHaliburton - also curious on your take, as this is starting to get into "doing science" rather than just presenting data.  Is there a consensus way to allocate cells to genes (ie, what Colin asks for above), without making an opinionated decision?

Issue #480, opened 2018-11-28
annotations with +inf, -inf or NaN floats will generate error 


bkmartinjr:
Currently, all numeric data is serialized to the web front-end via a JSON serializer.   JSON has no standardized support for +Inf, -Inf or NaN floats.  The Python JSON serializer converts them to special tokens (+Infinity, -Infinity and NaN respectively - note that these are tokens, NOT JSON strings).   The Chrome/V8 JSON deserializer does not recognize these tokens, and throws an error.

end result is that any AnnData containing these special float values in annotations will see a browser error when loading.

Note: this issue affects all AnnData floats (not just annotations).

PR #484, opened 2018-11-29
issue #480 workaround

bkmartinjr:
Work-around to handle non-finite floating point values in datasets.   This does not completely fix the issue, but should provide a temporary fix.   Any NaN or Infinite values in floating point data are set to zero.

PR includes two other related but small issues:
* stop loading error-prone var annotations (until we have legit use case, the are just a speed & memory hit)
* comment out erroneous use of var annotation (see issue #483 for the details) in recently added fuzzy search for gene names.

### DIFF
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -195,7 +195,7 @@ class ScanpyEngine(CXGDriver):
             for ann in curr_axis:
                 dtype = curr_axis[ann].dtype
                 if dtype.kind == 'f':
-                    not_finite = np.isfinite(curr_axis[ann]) == False
+                    not_finite = np.isfinite(curr_axis[ann]) == False   # noqa: E712
                     if np.count_nonzero(not_finite) > 0:
                         warnings.warn(
                             f"{str(ax).title()} annotation '{ann}' contains floating point NaN or Infinities. "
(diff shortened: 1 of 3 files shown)


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, data, args):
         self.diffexp_options = ["ttest"]
         self._create_schema()
 
+        # TODO: temporary work-arounds
+        self._IEEE754_X_warning_issued = False
```
csweaver: We don't seem to be using this variable, recommend deleting it

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -172,6 +176,50 @@ def _validate_data_calculations(self):
...
+        Where non-finite floating point is present in obs, var or X:
+        * issue a warning to the user that these values will be treated as zeros.
+        * set the value to zero within the in-memory data (self.data)
+        """
+        for ax in Axis:
+            curr_axis = getattr(self.data, str(ax))
+            for ann in curr_axis:
+                dtype = curr_axis[ann].dtype
+                if dtype.kind == 'f':
+                    not_finite = np.isfinite(curr_axis[ann]) == False   # noqa: E712
```
csweaver: np.isfinite checks for both infinity and nans? cool!

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: oops - this is because I forgot to use it.   Great catch!

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: yes indeed!

PR #503, opened 2018-12-07
[EASY] improve FAQ formatting

freeman-lab:
randomly learned today about the use of the `<details>` tag to make collapsible lists in github readmes, so i tried adding it to make the FAQ easier to browse. simple short-term fix for a critique that @bkmartinjr raised ("can't find anything in the FAQ"), and will let us expand the FAQ without the readme becoming unwieldy.

details

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,12 @@ pip install cellxgene
 
 ## FAQ
 
+<details>
+
+<summary> questions about data formatting </summary>
+  
+<hr>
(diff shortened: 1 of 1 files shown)


freeman-lab:
@bkmartinjr totally agree, this was meant as a short-term improvement :) a proper site will be probably needed for what you're describing, starts to stretch the limits of a readme

PR #505, opened 2018-12-10
Add Docker file

csweaver:
Now that cellxgene can be installed via pip, the docker container creation is a little more straightforward. I've updated @flying-sheep's request but was able to start from an ubuntu image instead of having to use the scanpy image. 

Adding the container build to the tests takes approx 1min 15s. I'm not sure if it is worth it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:bionic
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+EXPOSE 5005:5005
+
+RUN apt-get update
+RUN apt-get install -y build-essential libxml2-dev python3-dev python3-pip zlib1g-dev
+RUN pip3 install cellxgene
+
+ENTRYPOINT ["cellxgene"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,8 @@ FROM ubuntu:bionic
 ENV LC_ALL=C.UTF-8
 ENV LANG=C.UTF-8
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:bionic
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+EXPOSE 5005:5005
+
+RUN apt-get update
+RUN apt-get install -y build-essential libxml2-dev python3-dev python3-pip zlib1g-dev
```
mckinsel: I believe best practice is to combine these and remove the cache at the end:
```dockerfile
RUN apt-get update && \
    apt-get install -y ... && \
    rm -rf /var/lib/apt/lists/*
```
see https://docs.docker.com/v17.12/develop/develop-images/dockerfile_best-practices/#run

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:bionic
+
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
+EXPOSE 5005:5005
```
mckinsel: I'm not sure this has any effect? I don't really know for sure, but I think the `-p 5005:5005` is doing all the work.

Dockerfile
csweaver: Good catch, I had this in to debug, but forgot to re-"&& \"-ify it. WIll fix.

Dockerfile
csweaver: removed

PR #511, opened 2018-12-13
Move jsonification to engine level

csweaver:


csweaver:
Goal: be able use the same routes for flatbuffer and json data depending on what the client asks for

I've switched from flask's jsonify to json.dumps (keeps the flask logic out of the engine) This required adding the content type headers to the make response. I created a constant for content type and we can add the binary type with that PR.

server/app/rest_api/rest.py
```diff
@@ -147,13 +192,16 @@ class AnnotationsObsAPI(Resource):
...
+        content_type = JSON_MIMETYPE
         fields = request.args.getlist("annotation-name", None)
         try:
             annotation_response = current_app.data.annotation({}, "obs", fields)
         except KeyError:
             return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         try:
-            return make_response(jsonify(annotation_response), HTTPStatus.OK)
+            return make_response(
+                annotation_response, HTTPStatus.OK, {"Content-Type": content_type}
```
bkmartinjr: I don't find this style (use of `content_type`) to be helpful.   If we are going to move away from strings, it should be to names that signify intent (eg, `HTTPStatus.BAD_REQUEST`).

So I would either leave this as a string ("application/json"), or I would use `JSON_MIMETYPE` directly.

server/app/rest_api/rest.py
```diff
@@ -147,13 +192,16 @@ class AnnotationsObsAPI(Resource):
...
+            )
         except ValueError as e:
```
bkmartinjr: this restructuring will fail to capture JSON failures (which throw a ValueError).   

I suggest moving the make_response into the first try block, and just catch all of the errors for one try.

server/app/rest_api/rest.py
```diff
@@ -194,15 +247,20 @@ def get(self):
...
         except KeyError:
             return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         except FilterError as e:
             return make_response(e.message, HTTPStatus.BAD_REQUEST)
         try:
-            return make_response(jsonify(annotation_response), HTTPStatus.OK)
+            return make_response(
+                annotation_response, HTTPStatus.OK, {"Content-Type": content_type}
+            )
         except ValueError as e:
```
bkmartinjr: see previous comment about ValueError

server/app/rest_api/rest.py
```diff
@@ -240,13 +302,16 @@ class AnnotationsVarAPI(Resource):
...
         try:
             annotation_response = current_app.data.annotation({}, "var", fields)
         except KeyError:
             return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         try:
-            return make_response(jsonify(annotation_response), HTTPStatus.OK)
+            return make_response(
+                annotation_response, HTTPStatus.OK, {"Content-Type": content_type}
+            )
         except ValueError as e:
```
bkmartinjr: see previous comment about ValueError

server/app/rest_api/rest.py
```diff
@@ -283,15 +357,20 @@ def get(self):
...
         except KeyError:
             return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
         except FilterError:
             return make_response("Malformed filter", HTTPStatus.BAD_REQUEST)
         try:
-            return make_response(jsonify(annotation_response), HTTPStatus.OK)
+            return make_response(
+                annotation_response, HTTPStatus.OK, {"Content-Type": content_type}
+            )
         except ValueError as e:
```
bkmartinjr: see previous comment about ValueError

server/app/rest_api/rest.py
```diff
@@ -549,14 +737,17 @@ def post(self):
...
+                current_app.data.features["diffexp"]["interactiveLimit"],
             )
         except (ValueError, FilterError) as e:
             return make_response(e.message, HTTPStatus.BAD_REQUEST)
         except InteractiveError:
             return make_response("Non-interactive request", HTTPStatus.FORBIDDEN)
         try:
-            return make_response(jsonify(diffexp), HTTPStatus.OK)
+            return make_response(diffexp, HTTPStatus.OK, {"Content-Type": content_type})
         except ValueError as e:
```
bkmartinjr: see previous comment about ValueError

server/app/rest_api/rest.py
csweaver: I was anticipating your additions where content_type would be set to BINARY_MIMETYPE earlier. Happy to make your changes if you had a different way of handling this.

server/app/rest_api/rest.py
csweaver: will do

server/app/rest_api/rest.py
bkmartinjr: Ah, I see.  I thought it would be clearer to have two separate branches which each called make_response.  That would be more stylistically aligned with the rest of the code.  

In other words, something link:
```
   if accept_type == "application/octet-stream":
       make_ressponse(current_app.data.doitReturningFBS({}, "obs", fields), HTTPStatus.OK, {"Content-Type": "application/octet-stream"})
   elif accept_type == "application/json":
       make_ressponse(current_app.data.doItReturningJSON({}, "obs", fields), HTTPStatus.OK, {"Content-Type": "application/json"})
   else:
       make_response("Unsupported MIME types", HTTPStatus.NOT_ACCEPTABLE)
```

that said, I'm not feeling dogmatic!

csweaver: I made the change, I was mis-predicting what you would be doing when combing the routes for the flatbuffer PR.

server/app/rest_api/rest.py
csweaver: fixed all

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -352,13 +394,19 @@ def annotation(self, filter, axis, fields=None):
...
             var = self.data.var[var_selector]
             if not fields:
                 fields = var.columns.tolist()
-            result = {"names": fields, "data": DataFrame(var[fields]).to_records(index=True).tolist()}
-        return result
+            result = {
+                "names": fields,
+                "data": DataFrame(var[fields]).to_records(index=True).tolist(),
+            }
+        return json.dumps(result, cls=Float32JSONEncoder)
```
bkmartinjr: Have you confirmed this will correctly throw an error if NaN is present in the data?   I am not clear if we need to specify `allow-nan=False` directly in the dumps() args, or if specifying it in the encoder is sufficient.   dumps() and jsonify() work a bit differently, so we will need to confirm so there is not a feature regression.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -382,27 +430,37 @@ def data_frame(self, filter, axis):
...
         else:
             result = {
                 "obs": obs_index_sliced.tolist(),
-                "var": DataFrame(_X.T, index=var_index_sliced).to_records(index=True).tolist(),
+                "var": DataFrame(_X.T, index=var_index_sliced)
+                .to_records(index=True)
+                .tolist(),
             }
-        return result
+        return json.dumps(result, cls=Float32JSONEncoder)
```
bkmartinjr: See previous comment about ensuring this will trap NaNs/Infinities

server/app/rest_api/rest.py
```diff
@@ -150,10 +195,11 @@ def get(self):
...
         try:
             annotation_response = current_app.data.annotation({}, "obs", fields)
+            return make_response(
+                annotation_response, HTTPStatus.OK, {"Content-Type": JSON_MIMETYPE}
+            )
         except KeyError:
             return make_response(f"Error bad key in {fields}", HTTPStatus.BAD_REQUEST)
-        try:
-            return make_response(jsonify(annotation_response), HTTPStatus.OK)
         except ValueError as e:
```
bkmartinjr: One concern I have with handling ValueError at this level: it is a widely used Python error, and in the future could potentially be generated by many different code paths, thereby confusing JSON encoding errors with other conditions.

I think it would be better if this specific exception handling was wrapped directly around the call to JSON.dumps(), and then rethrown as an app-specific error (eg, JSONEncodingError)

server/app/rest_api/rest.py
```diff
@@ -555,10 +568,15 @@ def get(self):
             )
         except FilterError as e:
             return make_response(e.message, HTTPStatus.BAD_REQUEST)
-        except ValueError as e:
+        except JSONEncodingValueError as e:
+            # JSON encoding failure, usually due to bad data
+            return make_response(str(e), HTTPStatus.INTERNAL_SERVER_ERROR)
+        except JSONEncodingValueError as e:
```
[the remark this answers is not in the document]
csweaver: fixed 🤦‍♀️

csweaver:
I think I've covered your two concerns and improved the testing for nans so we won't break in the future

PR #512, opened 2018-12-13
Clarify that we do not support python 3.7

### LINKED ISSUE #510: Clarify that python 3.7 is not supported
Clarify in the readme that 3.6 is the only python version that is supported.

csweaver:
Added warning that we do not support python 3.7 in readme, and an import error to installation when installing with anything but python 3.6. 

I also added some more classifiers while I was under the hood in setup.py

Fixes #510

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Or you can create a virtual environment by using
 
 ```
 ENV_NAME=cellxgene
-python3 -m venv ${ENV_NAME}
+python3.6 -m venv ${ENV_NAME}
 source ${ENV_NAME}/bin/activate
 pip install cellxgene
 ```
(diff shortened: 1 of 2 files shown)


csweaver:
@bkmartinjr yep, I had intended to change the venv section to ensure python 3.6 but it slipped by. I've modified that. 

Where in the FAQ would you suggest putting that we will support 3.7 eventually comment? I agree that we should, but it is not obvious which section.

bkmartinjr:
I wasn't suggesting a commitment, but rather rewording "is not supported" to "is known to not work" or something that does not imply a decision to not support...

csweaver:
Ahh, got it! Yeah, that's a good idea; I'll change the language.

PR #513, opened 2018-12-13
Cluster Occupancy (n dim cube)

### LINKED ISSUE #462: n-dim summary cube
On startup and subset to current selection, summarize how categorical metadata is broken down by other categorical metadata, add the results of that calculation to world in the following format:

image

Use case:

image

colinmegill:
image

Fixes #462

### DIFF
--- a/client/src/components/categorical/occupancy.js
+++ b/client/src/components/categorical/occupancy.js
@@ -1,29 +1,65 @@
 // jshint esversion: 6
 import React from "react";
 import _ from "lodash";
+import { connect } from "react-redux";
+import * as d3 from "d3";
 
 @connect()
 class Occupancy extends React.Component {
   render() {
     const {
-      categoricalAsBooleansMap,
-      metadataField,
-      count,
-      value,
-      colorAccessor,
+      occupancy,
       colorScale,
+      metadataField,
+      colorAccessor,
       schema
     } = this.props;
+    const width = 100;
+    const height = 11;
+
+    const categories = _.filter(schema.annotations.obs, {
+      name: colorAccessor
+    })[0].categories;
+
+    const x = d3
+      .scaleLinear()
+      /* get all the keys d[1] as an array, then find the sum */
+      .domain([0, d3.sum(Array.from(occupancy, d => d[1]))])
+      .range([0, width]);
+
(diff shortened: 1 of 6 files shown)


client/src/components/categorical/value.js
```diff
@@ -52,13 +57,22 @@ class CategoryValue extends React.Component {
     /* this is the color scale, so add swatches below */
     const c = metadataField === colorAccessor;
```
bkmartinjr: variable name could be more descriptive

client/src/components/categorical/occupancy.js
```diff
@@ -0,0 +1,88 @@
...
+    //   stack.push({
+    //     key,
+    //     value,
+    //     rectWidth: scaledValue,
+    //     offset: currentOffset,
+    //     fill: colorScale(categories.indexOf(key))
+    //   });
+    //   currentOffset += scaledValue;
+    // });
+
```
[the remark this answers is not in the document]
bkmartinjr: I agree!

client/src/components/categorical/occupancy.js
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <svg
+        style={{
+          marginRight: 5,
+          width,
+          height
+        }}
+      >
+        {_.map(stack, d => (
```
bkmartinjr: I like the fact that you are no longer using lodash as much (eg, above array.forEach()).   Why not use the built-in map as well?

client/src/components/categorical/occupancy.js
```diff
@@ -0,0 +1,85 @@
...
+    //     rectWidth: scaledValue,
+    //     offset: currentOffset,
+    //     fill: colorScale(categories.indexOf(key))
+    //   });
+    //   currentOffset += scaledValue;
+    // });
+
+    const cat = categoricalSelectionState[colorAccessor];
+
+    cat.categoryValues.forEach(d => {
```
bkmartinjr: this could be rewritten as `cat.categoryValues.map(...)`, and get rid of the manual stack.push().  Style comment only, so at your discretion.  I tend to prefer map() when that is the only side effect.

PR #520, opened 2018-12-18
Use url_root from Flask rather than hardcoding hostname.

jxtx:
Currently the base URL for pull requests hardcodes the hostname.

This was preventing me from running in a remote host's Docker container
or behind a proxy without some hostname hacks.

This changes the url_base for API requests to be determined dynamically
from the request, so it will work behind proxies, etc.


bkmartinjr:
@jxtx - thank you for the PR!   We are going to run it through our normal test & release process, and will reach out if there are any questions.  Feel free to join us on the CZI Science slack (details in the repo index file).

PR #535, opened 2018-12-27
performance improvements in color-by feature

### LINKED ISSUE #534: switching between colors with large number of cells is slow
The color-by feature has gotten slow for large number of cells (1M+).   This is largely due to how colors are updated in updateCellColors.js, and can be optimized.

bkmartinjr:
Fixes #534 

Color by metadata has become quite slow for large collections (multiple seconds to switch color-by choices with 1MM+ cell count).  This PR contains several changes:
* fixes accidentally memoization defeat in histogram creation for brushableHistogram component (was triggering a full histogram rebuild on each color change).
* the unused "colorName" array has been removed from the `controls` reducer.   It previously held a string representation of each cell color (eg, rgb(0,0,0,1)).
* code in `updateCellColors.js` previously set all cell colors by color name, and then parsed into RGB space (ie, 1M+ string parse operations).  This has been modified to compute a predefined number of discrete color bins, and assign these to each cell.

The last change has the effect of discretizing color assignment.  For categorical metadata this is identical to previous behavior - just much faster.   For continuous metadata (either annotations or expression), the code creates a constant number of color bins.  The number of bins can be tuned until it is a reasonable compromise between speed and aesthetics.   It is currently set to 100 color bins, which looks the same to my eye.

@colinmegill - please advise if you wish a larger or smaller number of color bins for continuous metadata.  The current number is based upon my own casual observation, which is known deficient :-)

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -26,7 +26,7 @@ import actions from "../../actions";
   obsAnnotations: _.get(state.controls.world, "obsAnnotations", null)
 }))
 class HistogramBrush extends React.Component {
-  calcHistogramCache = memoize((obsAnnotations, field, ranges) => {
+  calcHistogramCache = memoize((obsAnnotations, field, rangeMin, rangeMax) => {
     const { world } = this.props;
     const histogramCache = {};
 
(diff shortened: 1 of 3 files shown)


colinmegill:
:) this looks fine to my eye. 100 bins looks the same to me.

PR #536, opened 2019-01-02
Landing page faq comments

GenevieveHaliburton:
Hi @neuromusic @fionagriffin 
I made a few updates to the landing page `index` and `faq`. The main changes are to add one question about what are the acceptable file formats and the briefest overview of what `cellxgene prepare` does. I also changed some words in the titles from `monospace` to _italics_ because the way they rendered with the current theme made it a little hard to read. 
In the new question I added, i say that `cellxgene prepare` supports any file format currently supported in `scanpy` but I'm not actually sure if that's true. We can definitely fix that if I overstated there.
Thanks!

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -5,7 +5,7 @@ description: Frequently Asked Questions
 ---
 
 
-# data formatting
+# Data formatting
 
 #### What file formats can I use with _cellxgene_?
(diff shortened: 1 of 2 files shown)


neuromusic:
resolved conflicts with master.

can @colinmegill give this a review, too?

colinmegill:
Made some changes to ordering and readability, looks good to me at this point

PR #543, opened 2019-01-07
Async on long running operations

colinmegill:
This pull requests implements async feedback for user defined genes and differential expression

async

### DIFF
--- a/client/src/reducers/controls.js
+++ b/client/src/reducers/controls.js
@@ -323,6 +323,18 @@ const Controls = (
         }
       };
     }
+    case "request user defined gene started": {
+      return {
+        ...state,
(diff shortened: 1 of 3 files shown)


client/src/reducers/controls.js
```diff
@@ -323,6 +323,18 @@ const Controls = (
         }
       };
     }
+    case "request user defined gene started": {
+      return {
+        ...state,
+        userDefinedGenesLoading: true
```
bkmartinjr: shouldn't this also be defined in the intial state value (lines 102-134), as we do with other state values?

colinmegill: Yup! Nice catch

PR #548, opened 2019-01-10
remove --nan-to-num CLI parameter

### LINKED ISSUE #527: remove `--nan-to-num`
Once #509 is merged, cellxgene can handle non-finite floats (NaNs & +/-Infs) in the raw data (both annotations and gene expression).

This appears to make `--nan-to-num` unnecessary, as it was explicitly added (in #490) to address this issue (#480).

If there are no strong use-cases for continuing to provide this feature, we should remove it.

bkmartinjr:
with the move to binary wire format, non-finite numbers can be serialized.  This removes the need for the `--nan-to-num` CLI parameter.

Changes:
* removed --nan-to-num from code, docs & tests
* lint/pretty printing of a few MD files that were modified for the above 
* fixed bug in dataframe filtering for FBS route (did not correctly handle a filter of None)

Fixes #527


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -480,7 +411,9 @@ def data_frame_to_fbs_matrix(self, filter, axis):
             raise FilterError("filtering on obs unsupported")
 
         # Currently only handles VAR dimension
-        X = self.data._X[:, var_selector]
+        X = self.data._X
+        if var_selector is not None:
+            X = X[:, var_selector]
```
csweaver: Is this related to the nan-to-num or is this just an improvement?

server/test/test_nan_scanpy_engine.py
```diff
@@ -16,69 +19,52 @@ def setUp(self):
...
-        self.assertEqual(annotations["names"], ["name", "n_cells", "var_with_nans"])
-        self.assertEqual(len(annotations["data"]), 100)
+        self.assertEqual(annotations["n_rows"], 100)
+        self.assertTrue(math.isnan(annotations["columns"][2][0]))
+
+        annotations = decode_fbs.decode_matrix_FBS(self.data.annotation_to_fbs_matrix("var"))
+        self.assertEqual(annotations["col_idx"], ["name", "n_cells", "var_with_nans"])
+        self.assertEqual(annotations["n_rows"], 100)
+        self.assertTrue(math.isnan(annotations["columns"][2][0]))
+
```
csweaver: Thanks for reworking the tests! They look great!

server/test/test_nan_scanpy_engine.py
```diff
@@ -16,69 +19,52 @@ def setUp(self):
...
-        self.assertEqual(len(data_frame_var["var"]), 100)
-        self.assertEqual(len(data_frame_var["obs"]), 100)
+        data_frame_var = decode_fbs.decode_matrix_FBS(self.data.data_frame_to_fbs_matrix(None, "var"))
+        self.assertIsNotNone(data_frame_var)
+        self.assertEqual(data_frame_var["n_rows"], 100)
+        self.assertEqual(data_frame_var["n_cols"], 100)
+        self.assertTrue(math.isnan(data_frame_var["columns"][3][3]))
+
+        with self.assertRaises(ValueError) as cm:
+            data_frame_obs = decode_fbs.decode_matrix_FBS(self.data.data_frame_to_fbs_matrix(None, "obs"))
```
csweaver: causing failing flake8. Just need to remove `data_frame_obs = ` assignment since the variable isn't used again.

server/app/util/constants.py
```diff
@@ -28,5 +28,5 @@ class DiffExpMode(AugmentedEnum):
 
 
 JSON_NaN_to_num_warning_msg = (
-    "JSON encoding failure - suggest trying --nan-to-num command line option"
+    "JSON encoding failure - please verify all data are finite values (no NaN or Infinities)"
```
csweaver: Could we add a comment here that when working with datasets including non-finite values to use flatbuffer routes instead of JSON? Low priority, I'm mainly thinking about external developers here.

server/test/test_nan_scanpy_engine.py
csweaver: this is testing that we don't support dataframes by "obs" with the flatbuffer method, right? I think that might be better as a separate test.  `test_dataframe_obs_not_implemented` or something similar

README.md
```diff
@@ -8,10 +8,9 @@
 
 ## getting started
 
-You'll need **python 3.6** and **Google Chrome**. (*Warning*: Python 3.7 is **not** supported at this time)
+You'll need **python 3.6** and **Google Chrome**. (_Warning_: Python 3.7 is **not** supported at this time)
```
csweaver: Could we make the doc clean up a separate PR instead of including them in the --nan-to-num change? I can do that since I know the git clients for linux for windows don't have the best GUI for extracting line level changes.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: latent bug required to support the test cases (not used by the front-end, but should be supported).     This was noted in the PR comments:  "fixed bug in dataframe filtering for FBS route (did not correctly handle a filter of None)"

server/app/util/constants.py
bkmartinjr: well, in theory this error will never occur again, as the front-end doesn't call those routes (and nobody else uses the back-end).    I simply changed this error as prelude to removing the whole set of unused JSON routes...

README.md
bkmartinjr: OK with me.  I understand you are doing this, right?

csweaver: yep!

server/app/scanpy_engine/scanpy_engine.py
csweaver: Ahh, I understand now.

server/app/util/constants.py
csweaver: yeah, that's fine

server/test/test_nan_scanpy_engine.py
bkmartinjr: fixed

bkmartinjr: will fix

bkmartinjr: note - the same error occurred in all of the exception tests, but wasn't noted by flake8 due to a reused variable.  I'll fix it everywhere.

server/app/util/constants.py
bkmartinjr: cleanup reminder filed as issue #550

PR #551, opened 2019-01-10
home page fix

neuromusic:
This PR makes two changes:
- adds the homepage URL to the Jekyll config so that `{{ site.url }}` works (see https://jekyllrb.com/docs/variables/#site-variables)
- adds an scss file that will allow overriding the site theme with custom css or sass (see https://help.github.com/articles/customizing-css-and-html-in-your-jekyll-theme/#customizing-your-jekyll-themes-css)

note: I'm not sure how to check the first item without merging into master since this variable is overridden when running Jekyll locally

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,10 +1,7 @@
 theme: jekyll-theme-cayman
 show_downloads: false
-url: https://chanzuckerberg.github.io/cellxgene/
 
 nav:
-  - title: Home
-    url: /
   - title: Data
     url: data.html
   - title: FAQ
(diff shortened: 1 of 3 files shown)


neuromusic:
wait, no, I think I've got the config variable wrong... it should be `{{ site.baseurl }}` in the template...

neuromusic:
ok, I think this will work now, but I'm not entirely sure how to test it without pushing it to master since jekyll configures `{{ site.baseurl }}` based on the host and I'm pretty sure Github sets it to `cellxgene/`

neuromusic:
OK, made some changes based on https://byparker.com/blog/2014/clearing-up-confusion-around-baseurl/ & confirmed that this works locally, so it *should* work once deployed. Worst case, the link is still broken and we're no worse of than now.

PR #568, opened 2019-01-30
Set API path based on access address

ivirshup:
Building on #520, this should make using cellxgene through a proxy easier. Instead of using `request.url_root` this will use the whole address, making it easier to host cellxgene instances at multiple routes, or via https. Here's a simple example:

```apacheconf
# /etc/apache2/sites-enabled/reverseprox.conf
<VirtualHost *:80>
    # Setup
    ServerName {server_url}
    LoadModule proxy_module modules/mod_proxy.so
    ProxyPreserveHost On
    ProxyRequests Off
    # Setting routes for cellxgene
    ProxyPass "/dataset1/" "http://0.0.0.0:5005/"
    ProxyPassReverse "/dataset1/" "http://0.0.0.0:5005/"
    ProxyPass "/dataset2/" "http://0.0.0.0:5006/"
    ProxyPassReverse "/dataset2/" "http://0.0.0.0:5006/"
</VirtualHost>
```

The only remaining issue are the `/static` resources, whose url's seem to be set during the build process. I'd appreciate any suggestions on this bit.

### DIFF
--- a/client/configuration/webpack/webpack.config.prod.js
+++ b/client/configuration/webpack/webpack.config.prod.js
@@ -10,7 +10,7 @@ const nodeModules = path.resolve("node_modules");
 
 const babelOptions = require("../babel/babel.prod");
 
-const publicPath = "/";
+const publicPath = "";
 
 module.exports = {
   mode: "production",
--- a/client/index_template.html
+++ b/client/index_template.html
@@ -24,7 +24,7 @@
 <script type="text/javascript">
 window.CELLXGENE = {};
 window.CELLXGENE.API = {
-  prefix: "{{ prefix | safe }}",
+  prefix: window.location.href + "api/",
   version: "v0.2/"
 };
 </script>
--- a/server/app/web/templates/swagger.html
+++ b/server/app/web/templates/swagger.html
@@ -73,7 +73,7 @@
 <script>
 	window.onload = function () {
 		const ui = SwaggerUIBundle({
-			url: window.location.origin + "/api/swagger.json",
+			url: window.location.href.replace(/\/swagger$/, "") + "/api/swagger.json",
 			dom_id: '#swagger-ui',
 			deepLinking: true,
 			presets: [
--- a/server/app/web/webapp.py
+++ b/server/app/web/webapp.py
@@ -7,9 +7,8 @@ bp = Blueprint("webapp", __name__, template_folder="templates")
 
 @bp.route("/")
 def index():
-    url_base = request.url_root + "api/"
     dataset_title = current_app.config["DATASET_TITLE"]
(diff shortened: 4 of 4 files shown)


ivirshup:
No problem, fixed!

csweaver:
@ivirshup Thanks for your patience. It took me a little while to set up a local dev environment for testing this. I wound up using a docker container [rgoyard/apache-proxy ](https://hub.docker.com/r/rgoyard/apache-proxy) to set up a proxy server on my laptop and ran cellxgene on another container without exposing the port but connected to the same docker network as the proxy. Since the static files are all served either from external sources or with relative paths, there were no issues loading them in.  

At this point, I'm happy to merge. I just wanted to check with you about the static resources. Did your specific setup experience issues loading any of the files?

ivirshup:
I did, but I've worked around it. I'm pretty inexperienced at web stuff, so it's possible I've done something wrong, but here's what I ran into.:

I have cellxgene apps at `https://{server}/datasets/{datasetname}/`. This is done with a similar apache configuration to what I provided above. With just this setup the icons (like the color droplet, pan, and lasso) didn't load correctly. This was due to the page looking for them at: `https://{server}/static/media/` when they would be available at `https://{server}/datasets/{datasetname}/static/media/`. I've just made the static resources available on the expected path as a workaround.

csweaver:
Ahh, I just tried proxying with "/" not with a full prefix. I'll take a stab at that tomorrow. I'm going to wait on merging until then.

csweaver:
Hey @colinmegill! I need your help on this one. Webpack looks like it is adding a prefixing "/" to all the static assets when it builds the template. This breaks asset delivery when run behind a reverse proxy. 

We have two options: 
1. Add a <base> tag to the page that is dynamically generated and gets resolved before the assets load.
2. Updating the webpack config to remove the prefixing "/" from all static assets (is this possible?)

Thoughts? Opinions?

csweaver:
@ivirshup I am able to replicate your problem! And I'm able to fix the assets if I remove the first "/" so that they become relative paths instead of absolute paths. @colinmegill is in charge of the front end, so I want to get his weigh-in before figuring out how to fix this.

csweaver:
@ivirshup Solved! Colin suggested a fix and I've confirmed that it works. 

On line 13 of client/configuration/webpack/webpack.config.prod.js set public path to empty string. 
`const publicPath = ""; `

Once you push that commit, I'll approve and merge.

ivirshup:
Thanks!

csweaver:
No, thank you! We haven't been developing for deploying cellxgene on a server, so we would have never known about this bug let alone have fixed it. 

I'd love to know how you are using cellxgene for your projects. We have a slack channel you could join, or you can just drop me an e-mail.

PR #576, opened 2019-02-01
Dataframe

bkmartinjr:
Introducing a "dataframe" to cellxgene: monomorphic, columnar store for storing 2D matrix data with row/col indexing (a la Pandas dataframe).   This should provide both performance and abstraction benefits.   Examples:
* flatbuffers received from the server-side no longer have to be converted to row-major internal format;
* extraction of data for histogram rendering is "free" (ie, the underlying array of data is used directly)
* typed array (native types) continuous data no longer need to be transformed for use in crossfilters

The major change in app-internal abstraction is that mapping to/from the server "obsIndex" and "varIndex" row/col keys is done via the dataframe row/col indices, removing the need to manually maintain forward and back maps for world/universe sub-setting.

The dataframe module is used to manage all obs and var annotations, including obs X/Y layout.  It is not currently used to store matrix (expression) data, although that will likely land in a future set of changes.

Other changes buried in this PR:
* placeholder text for bulk add genes now displays gene names from the actual dataset (instead of fake, static-coded names) - components/geneExpression/index.js
* caching of "universe" crossfilter, dramatically improving performance of the Reset UI - reducers/controls.js
* significant rendering speed improvements in graph & scatterplot
* variety of other small performance improvements, primarily to crossfilter


client/src/util/stateManager/world.js
```diff
@@ -143,35 +125,11 @@ export function createWorldFromCurrentSelection(universe, world, crossfilter) {
...
-
-  /* now slice */
-  newWorld.obsAnnotations = sliceByIndex(universe.obsAnnotations, obsIndex);
-  newWorld.obsLayout = {
-    X: sliceByIndex(universe.obsLayout.X, obsIndex),
-    Y: sliceByIndex(universe.obsLayout.Y, obsIndex)
-  };
+  /* now subset/cut obs */
+  const mask = crossfilter.allFilteredMask();
+  newWorld.obsAnnotations = world.obsAnnotations.icutByMask(mask, null);
```
csweaver: Could use default parameters to make these function calls cleaner. 
```
# in dataframe.js
function icutByMask(rowMask, colMask=null)
# here
icutByMask(rowMask) 
```
I don't know if this breaks your optimizations, though.

client/src/middleware/updateCellColors.js
```diff
@@ -41,13 +41,13 @@ const updateCellColorsMiddleware = store => next => action => {
     action.type === "color by continuous metadata" ||
     action.type === "color by categorical metadata";
 
-  if (!filterJustChanged || !s.controls.world.obsAnnotations) {
+  const obsAnnotations = _.get(s.controls, "world.obsAnnotations", null);
```
csweaver: Is this for if the world hasn't been initialized yet?

client/src/actions/index.js
```diff
@@ -242,12 +242,14 @@ const requestDifferentialExpression = (set1, set2, num_genes = 10) => async (
...
-    const set2ByIndex = rangeEncodeIndices(
-      _.map(set2, s => universe.obsNameToIndexMap[s])
-    );
+
+    // These lines ensure that we convert any TypedArray to an Array.
+    // This is necessary because JSON.stringify() does some very strange
+    // things with TypedArrays (they are marshalled to JSON objects, rather
+    // than being marshalled as a JSON array).
+    const aset1 = Array.isArray(set1) ? set1 : Array.from(set1);
+    const aset2 = Array.isArray(set2) ? set2 : Array.from(set2);
```
csweaver: This was a general fix unrelated to dataframes, right?

client/src/reducers/controls.js
```diff
@@ -100,6 +100,28 @@ function selectedValuesForCategory(categorySelectionState) {
...
+        crossfilter,
+        gene
+      );
+      return acc;
+    }, {});
+  }
+
+  return {
+    ..._createGenesDimMap(userDefinedGenes, userDefinedDimensionName),
+    ..._createGenesDimMap(diffexpGenes, diffexpDimensionName)
```
csweaver: I'm don't understand what the spread operator is doing in the case.

client/src/util/stateManager/universe.js
```diff
@@ -73,61 +70,33 @@ function finalize(universe) {
...
-      acc[value.name] = idx;
-    },
-    {}
-  );
   universe.finalized = true;
   return universe;
 }
 
-function RESTv02AnotationsFBSResponseToInternal(arrayBuffer) {
+function RESTv02AnotationsFBSToDataframe(arrayBuffer) {
```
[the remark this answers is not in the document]
bkmartinjr: oops - will fix the misspelling.

Personally, I don't think we need the RESTv02 anymore.  It helped when the code was multi-protocol, but those days are gone.

client/src/util/dataframe/dataframe.js
```diff
@@ -0,0 +1,444 @@
...
+  }
+
+  iat(r, c) {
+    /*
+    Access a single value, for a row/col offset (integer) position
+    */
+    return this.__columns[c][r];
+  }
+
+  has(r, c) {
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/dataframe/dataframe.js
```diff
@@ -0,0 +1,444 @@
...
+    const [nRows, nCols] = this.dims;
+    const coff = this.colIndex.getOffset(c);
+    const roff = this.rowIndex.getOffset(r);
+    return coff >= 0 && coff < nCols && roff >= 0 && roff < nRows;
+  }
+
+  ihas(r, c) {
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/stateManager/world.js
```diff
@@ -16,37 +17,32 @@ of Universe.
...
+  A dataframe containing the X/Y layout for all obs.  Columns are named
+  'X' and 'Y', and rows are indexed in the same way as obsAnnotation.
 
-* crossfilter - a crossfilter object across world.obsAnnotations
+* summary: summary of each obsAnnotation column (eg, numeric extent for
+  continuous data, category counts for categorical metadata)
 
-* dimensionMap - an object mapping annotation names to dimensions on
-  the crossfilter
+* varDataCache: expression columns, in a kvCache.   XXX: maybe move to a
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/middleware/updateCellColors.js
bkmartinjr: yes

client/src/actions/index.js
bkmartinjr: it is related to the improved "mask" interface to crossfilter.    It now returns a TypedArray rather than an Array,   This in turn was motivated by Dataframe (connecting crossfilter to the Dataframe.icut function)

client/src/reducers/controls.js
bkmartinjr: this is classic reducer - we need to return a new object, but we want it populated with the contents of these other objects.  So the spread is an object literal spread: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals

csweaver: I phrased this poorly. I know what the spread operator does, but I'm used to seeing it in reducers applied to state. I think what I actually don't understand is how are dimension maps being used in the controls reducer. Once I understand that I can better review the dimmap changes in the reducer.

client/src/components/continuous/continuous.js
```diff
@@ -10,7 +10,6 @@ import HistogramBrush from "../brushableHistogram";
 
 @connect(state => ({
   ranges: _.get(state.controls.world, "summary.obs", null),
-  metadata: _.get(state.controls.world, "obsAnnotations", null),
```
csweaver: was this just unused?

client/__tests__/util/dataframe/dataframe.test.js
```diff
@@ -0,0 +1,296 @@
...
+
+    expect(df.at(0, "A")).toEqual(2);
+    expect(df.at(2, "B")).toEqual(3);
+    expect(df.iat(0, 0)).toEqual(0);
+    expect(df.iat(2, 1)).toEqual(5);
+  });
+});
+
+describe("dataframe subsetting", () => {
+  test("cutByList", () => {
```
csweaver: If you expect to be modifying the dataframe more, this would be a good location for nested describe blocks, so that you can label each of these cases (all rows one col, all rows two col, etc) for jest and it will run all of them instead of stopping at the first failure.

client/__tests__/util/dataframe/dataframe.test.js
```diff
@@ -0,0 +1,296 @@
+import * as Dataframe from "../../../src/util/dataframe";
+
+/*
+Known gaps in these tests:
+- row indexing, especially a la 'world', are not sufficiently tested
```
csweaver: I would like the `at/iat` methods to be at least cursorily tested too.

client/__tests__/util/stateManager/summarizeAnnotations.test.js
```diff
@@ -69,18 +75,27 @@ describe("summarizeAnnotations", () => {
...
-        name: "n1",
-        nameString: "hi",
-        nameBoolean: true,
-        nameFloat32: 39.3,
-        nameInt32: 99,
-        nameCategorical: 1
-      }
-    ];
-    const varAnnotations = [];
+    const obsAnnotations = new Dataframe.Dataframe(
```
csweaver: I recently learned about the faker.js package for creating data for tests, might be useful for cases like this. (Not recommending it for this particular PR, though)

client/__tests__/util/dataframe/dataframe.test.js
```diff
@@ -0,0 +1,296 @@
+import * as Dataframe from "../../../src/util/dataframe";
```
csweaver: Thanks for adding good tests!!

client/src/actions/index.js
csweaver: got it

client/src/util/stateManager/universe.js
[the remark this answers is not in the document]
csweaver: Not necessary in this PR, but I saw a few more cases of it. Maybe make a cleanup ticket for it?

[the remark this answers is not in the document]
bkmartinjr: I'll just fix them all.

client/src/util/stateManager/world.js
bkmartinjr: Doesn't break optimizations.  I'm a bit concerned about clarity though....I'll make the change, and we can see if it is helpful or not.

client/src/reducers/controls.js
bkmartinjr: In this case, it is setting state (or part of the state).   The reason I used a reducer here is for clarity - without it, I would have to construct the function return value (an object) by using Object.assign().  Semantically equivalent, but our coding style prefers the clarity of spread operators.

client/src/components/graph/graph.js
```diff
@@ -83,12 +84,13 @@ class Graph extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
+    const { renderCache } = this;
```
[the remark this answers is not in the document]
bkmartinjr: no pro or con, other than style.  Our established front-end style (via prettier / eslint) prefers use of deconstruction.   This is consistent with modern JS style.

client/src/components/geneExpression/index.js
```diff
@@ -39,39 +38,34 @@ const renderGene = (fuzzySortResult, { handleClick, modifiers, query }) => {
...
-};
 
 @connect(state => {
-  const metadata = _.get(state.controls.world, "obsAnnotations", null);
   const ranges = _.get(state.controls.world, "summary.obs", null);
   const initializeRanges = _.get(state.controls.world, "summary.obs");
 
   return {
     ranges,
-    metadata,
```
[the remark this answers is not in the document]
bkmartinjr: correct

client/src/components/geneExpression/cellSetButtons.js
```diff
@@ -14,7 +14,7 @@ class CellSetButton extends React.Component {
       eitherCellSetOneOrTwo
     } = this.props;
 
-    const set = _.map(crossfilter.allFiltered(), "name");
+    const set = World.getSelectedByIndex(crossfilter);
```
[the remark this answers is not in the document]
bkmartinjr: and also, _far_ more performant

client/src/components/brushableHistogram/index.js
```diff
@@ -35,9 +35,11 @@ class HistogramBrush extends React.Component {
       .scaleLinear()
       .range([this.height - this.marginBottom, 0]);
```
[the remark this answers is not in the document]
bkmartinjr: several of our performance sensitive components (scatterplot, graph) cache values that are derived from the component state, but very expensive to compute.  In this case, it is a bunch of info about the data range the histogram displays.  When the data range updates, we compute range extents and other values, and then cache them in the component.

client/src/components/continuous/continuous.js
bkmartinjr: correct

client/__tests__/util/dataframe/dataframe.test.js
bkmartinjr: fixed.

client/__tests__/util/dataframe/dataframe.test.js
bkmartinjr: will add

bkmartinjr:
@csweaver - I believe I have addressed all of your review comments.  TY!

client/src/util/stateManager/world.js
csweaver: coming from python I am all about default params, but I can see the argument

client/src/reducers/controls.js
csweaver: setting "part of the state" is what I was missing. thanks

client/src/components/brushableHistogram/index.js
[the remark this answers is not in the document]
csweaver: Got it, that helps me understand the rest of the changes in this file (which all look good btw)

client/src/reducers/controls.js
```diff
@@ -100,6 +100,28 @@ function selectedValuesForCategory(categorySelectionState) {
   return selectedValues;
 }
 
+/*
+build a crossfilter dimension map for all gene expression related dimensions.
+*/
+function createGenesDimMap(userDefinedGenes, diffexpGenes, world, crossfilter) {
+  function _createGenesDimMap(genes, nameF) {
```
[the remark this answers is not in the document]
bkmartinjr: how does `nameCreator` sound?   (this is what you called the file containing these functions: `client/src/util/nameCreators.js`)

client/src/components/geneExpression/index.js
```diff
@@ -84,6 +78,37 @@ class GeneExpression extends React.Component {
     };
   }
 
+  placeholderGeneNames() {
```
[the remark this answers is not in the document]
bkmartinjr: yeah, I got a bit carried away with this PR.   I need to improve my PR factoring.  Apologies.

PR #577, opened 2019-02-02
Build improvements

csweaver:
Improved the build process:
1. Moved from bash scripts to makefile
2. Instead of copying files directly to source tree, create new build directory
3. Added targets in the make file to help smooth the build/install release process.

Setuptools for python has a lot of quirks and failings as well as bad documentation for several features. Because of that, the build improvements were more complex than originally anticipated. So I tried to document what each make target does and what problem it solves when the problem isn't obvious. I also added 3 release commands and updated the release documentation so that it's easier to do the right thing.


makefile
```diff
@@ -0,0 +1,122 @@
...
+PART ?= patch
+
+# BUILDING PACKAGE
+
+build : clean build-server
+	@echo "done"
+
+build-server : build-client
+	mkdir -p $(SERVERBUILD)
+	cp -r server/* $(SERVERBUILD)
```
ttung: It's plausible that you want to pull the list of files from `git ls-files` if you want a clean set of files to package up.

it's also possible that setup.py already cleans the set of files too.

.gitignore
```diff
@@ -9,6 +9,7 @@ coverage
 
 # production
 build/
+build-dev/
```
ttung: where is this created?

PR #604, opened 2019-02-21
smoke tests

csweaver:
I've set up a demo of how to write and integrate smoke tests using the puppeteer library.

1. Add data attributes to UI I/O elements "data-testid" for unique, "data-testclass" for general. 
- TODO
    - add to all elements we may want to capture.
2. Write jest test file using puppeteer to manage browser interactions. 
- TODO
    - set up env depending on local or ci (dev - use port 3000, launch browser, slow interactions, ci- use port 5005, headless mode)
    - write more puppeteer helper functions to facilitate our use case, off the top of my head - type into text box (including special characters), dev snapshot utility, better click/box model handling.
3. Integrate with travis ci - I'm using start-server-and-test package for this as it automates launching, listening for response on the servers port, and then running the tests. I also had to exclude these tests from the general jest test since we need to start the server for these
- TODO
    - conditionally run smoke tests (either nightly or only on master, etc) 

The end goal is to use the UI as a canary to flag us if some code change broke the app from previous behavior (ex. the layout got randomized). The test I have now are just examples that show how puppeter handles cellxgene's more customized interactions  -- any of the test libraries can click a button, but I needed to ensure we can select cells using lasso, brush histograms, and search for genes (cypress cannot handle some of these 😭).

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -7,11 +7,14 @@
   "scripts": {
     "backend-dev": "python3.6 -m venv cellxgene && source cellxgene/bin/activate && yes | pip uninstall cellxgene || true && pip install -e .. && cellxgene launch ",
     "build": "npm run clean && webpack --config configuration/webpack/webpack.config.prod.js",
+    "e2e": "jest e2e",
     "dev": "npm run clean && webpack --config configuration/webpack/webpack.config.dev.js",
     "clean": "rimraf build",
     "start": "node server/development.js",
+    "start-server-for-test": "cellxgene launch -p 5000 ../example-dataset/pbmc3k.h5ad",
+    "ci": "start-server-and-test start-server-for-test :5000 e2e",
     "lint": "eslint src",
-    "test": "jest"
+    "test": "jest --testPathIgnorePatterns e2e"
   },
   "engineStrict": true,
   "engines": {
@@ -100,6 +103,7 @@
     "puppeteer": "^1.12.1",
     "rimraf": "^2.6.3",
     "serve-favicon": "^2.3.0",
+    "start-server-and-test": "^1.7.11",
     "style-loader": "^0.23.1",
     "sw-precache-webpack-plugin": "^0.11.5",
     "url-loader": "^1.1.0",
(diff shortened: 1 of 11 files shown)


client/package.json
```diff
@@ -7,11 +7,14 @@
...
     "build": "npm run clean && webpack --config configuration/webpack/webpack.config.prod.js",
+    "e2e": "jest e2e",
     "dev": "npm run clean && webpack --config configuration/webpack/webpack.config.dev.js",
     "clean": "rimraf build",
     "start": "node server/development.js",
+    "start-server-for-test": "cellxgene launch -p 5000 ../example-dataset/pbmc3k.h5ad",
+    "ci": "start-server-and-test start-server-for-test :5000 e2e",
     "lint": "eslint src",
-    "test": "jest"
+    "test": "jest --testPathIgnorePatterns e2e"
```
bkmartinjr: maybe rename this 'unittest' or something like that?

client/src/components/geneExpression/cellSetButtons.js
```diff
@@ -38,6 +38,7 @@ class CellSetButton extends React.Component {
           type="button"
           disabled={differential.diffExp}
           onClick={this.set.bind(this)}
+          data-testid={`cellset_button_${eitherCellSetOneOrTwo}`}
```
bkmartinjr: is there a style guide for ids?   I see use of both dash and underscore (eg, foo-bar, foo_bar).

client/package.json
csweaver: good suggestion

client/src/components/geneExpression/cellSetButtons.js
csweaver: You caught me. I had meant to clean that up and forgot. Do you have a preference?

bkmartinjr: I don't.   I would prefer whatever is typical style.  Taking a quick look at eslint rules for react, I don't see anything jumping out.    So I'd probably default to dash, eg, foo-bar, unless it violates a typical styling standard.

client/package.json
csweaver: updated

client/src/components/geneExpression/cellSetButtons.js
csweaver: fixed, I went with "foo-bar" and I'll rely on @colinmegill for additional tweaks

PR #608, opened 2019-02-22
Dataframe, part deux - add varData and summarize()

### LINKED ISSUE #233: Performance validation (1M+ cells) on "typical" laptop
Performance validation (1M+ cells) on "typical" laptop.  

Items to check/investigate:
- [x] CLI load is slow for large data.  Appears to be loading twice due to flask app.run debug flag.  Reported #351
- [x] select/deselect categorical metadata on immune_bone_marrow data set is laggy. In particular, check/uncheck louvain category values is O(500ms) due to code in Category::componentDidUpdate() and controls reducer.  PR #356
- [x] CLI fetch of expression data via /data/obs is slow.  On immune_bone_marrow data, it takes 90 seconds to load one gene (var) of data the first time.  It is memoized, so the second load is 2 seconds, but that doesn't help as we only load it once (it gets cached on the front-end).   Reported  as issue #355
- [x] For sparse data, diffexp is super slow.  Consider --dense #394
- [x] Validate correct scanpy read() options (cache, backed, etc) with benchmarking
- [x] GPU infinite loop for scatter plot, filed as #388
- [x] is CLI memoization buying much if a) anndata is in memory, and b) front-end saves universe in mem?
- [x] consider flask threading
- [x] histogram builds are very slow for large data sets (symptom: load 1.3M 10X data, hit reset button).   Partially fixed, but much more coming with Dataframe.   Fix included in PR #576
- [x] scanpy diffexp calc looks faster - benchmark and consider switching.  See #425
- [x] for expression loads (eg, add a gene, diff exp histograms), data transmission is the bottleneck, and that is primarily a matter of using JSON to transmit large numeric arrays.   Need to look at binary encoding alternatives which are both denser and cheaper to encode/decode.  #271
- [x] color-by with large number of cells is slow  #534
- [x] replace min/max in updateCellColors with the new finiteExtent() function
- [x] client-side columnar format for world/universe data would be faster than current row-major format (in  progress via Dataframe)  Fix included in PR #576
- [x] the crossfilter for universe is recreated each time world is reset to universe.   This has substantial overhead (primarily in sorting dimension indices).   The base (universe) crossfilter could be cached in universe, and just reset each time world is set to universe.  This would dramatically speed up the Reset UI function.  Fix included in PR #576
- [x] a "spatial" dimension for crossfilter could be substantially optimized if X/Y lasso selection becomes a bottleneck.   Currently it is just doing a brute force polygon containment test with a couple of small optimizations.

### LINKED ISSUE #590: Front-end dataframe
Dataframe refactor:
- [x] immutable dataframe
- [x] obs and var annotations
- [x] layout
- [x] integrate with crossfilter
- [x] Use a dataframe for expression/var data
- [x] replace the annotation summary with something like [pandas.Dataframe.describe()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.describe.html#pandas-dataframe-describe).  Will make it more generally useful (eg, on non-annotation data), and allow for smarter caching when mutable dataframes are introduced.

bkmartinjr:
This is the second Dataframe-related refactoring PR.   Changes in this PR:
1. Dataframe is now used to manage all expression data (`varData`) for the universe and world.   This includes:
    * Dataframe API extended to include:
        * `withCol()` and `dropCol()`, which are used to create a new dataframe containing/not-containing a column
      * the ability to specify a row index during a cut operation, via the `withRowIndex` parameter
    * expression data stored in a Dataframe (`universe.varData`, `world.varData`), with all of the access to this data now using the Dataframe abstraction
    * subsetting of expression data done with `Dataframe.subset()` and related functions - now matches the subsetting for annotations and layout
    * caching semantics are dramatically simplified.  The keyvaluecache is removed, and the controls reducer simply manages the varData cache based upon its own state.  Specifically, it removes (prunes) columns from the varData Dataframe periodically, when the expression data is no longer required.  It keeps a configurable number of most-recently-used columns of data in case they are needed again.   This is entirely implemented inside the controls reducer (in `controls.js::pruneVarDataCache()`).
2. Summarization of annotation/expression data is now provided by `Dataframe.col().summarize()`.  This function is memoized for performance, and replaces the "summary" POJO previously stored inside universe and world.  Changes:
    * Objects `universe.summary` and `world.summary` removed
    * `Dataframe.col("x").summarize()` is used instead, and provides the same information by column
    * Note: while the content of the summary is same as the prior POJO, the layout in the summary object is slightly different (the `range` sub-object is flattened).
3. small amount of dead code removed from Universe & World, including fields `api` and `finalized` and function `subsetVarData`
4. Unit tests are updated for all of the above

Fixes #590 
Fixes #233


client/src/util/stateManager/world.js
```diff
@@ -38,80 +37,55 @@ Notable keys in the world object:
...
-  /* build the varDataCache */
-  world.varDataCache = kvCache.map(
-    universe.varDataCache,
-    val => subsetVarData(world, universe, val),
-    { lowWatermark: VarDataCacheLowWatermark, minTTL: VarDataCacheTTLMs }
-  );
+  /*
+  Var data columns - subset of all
+  */
+  world.varData = universe.varData.clone();
```
colinmegill: Subset === clone?

client/src/reducers/controls.js
```diff
@@ -122,6 +121,46 @@ function createGenesDimMap(userDefinedGenes, diffexpGenes, world, crossfilter) {
   };
 }
 
+function pruneVarDataCache(varData, needed) {
```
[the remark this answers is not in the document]
bkmartinjr: nice idea.  Seems like a separate PR though.  Is that OK?    I filed issue #611 to capture this.

client/src/util/stateManager/world.js
bkmartinjr: nope.

clone === shallow copy of entire dataframe.
cutByList, icutByList, icutByMask === subset/slice

bkmartinjr: @colinmegill - per offline conversation, the "cut" family of API has been renamed to "subset".  Please give it a look.

client/src/util/stateManager/universe.js
```diff
@@ -118,7 +81,7 @@ function reconcileSchemaCategoriesWithSummary(universe) {
     ) {
       const categories = _.union(
         _.get(s, "categories", []),
-        _.get(universe.summary.obs[s.name], "categories", [])
+        _.get(universe.obsAnnotations.col(s.name).summarize(), "categories", [])
```
csweaver: I was a little thrown that summarize was added in the last PR...

client/src/actions/index.js
```diff
@@ -243,12 +242,16 @@ const requestDifferentialExpression = (set1, set2, num_genes = 10) => async (
     const state = getState();
     const { universe } = state.controls;
 
+    // Legal values are null, Array or TypedArray.  Null is initial state.
+    if (!set1) set1 = [];
```
csweaver: did this change have to do with the dataframes or just general improvements?

client/src/reducers/controls.js
```diff
@@ -295,28 +334,60 @@ const Controls = (
...
-            World.subsetVarData(world, universe, val)
+            worldValSlice,
+            world.obsAnnotations.rowIndex
           );
         }
       });
 
+      // Prune size of varData "cache" if getting out of hand....
+      const { userDefinedGenes, diffexpGenes } = state;
+      const allTheGenesWeNeed = _.uniq(
```
csweaver: javascript needs better set functions

client/src/util/stateManager/world.js
```diff
@@ -242,14 +203,6 @@ export function worldEqUniverse(world, universe) {
   return world.obsAnnotations === universe.obsAnnotations;
 }
 
-export function subsetVarData(world, universe, varData) {
```
csweaver: this was unused?

client/__tests__/util/dataframe/dataframe.test.js
```diff
@@ -293,6 +316,222 @@ describe("dataframe factories", () => {
...
+    test("KeyIndex", () => {
+      const df = new Dataframe.Dataframe(
+        [2, 2],
+        [["red", "blue"], [true, false]],
+        null,
+        new Dataframe.KeyIndex(["colors", "bools"])
+      );
+      const dfA = df.withCol("numbers", [1, 0]);
+
+      expect(dfA).toBeDefined();
```
csweaver: I'll be on the lookout for a better way to run tests when many of the lines within the test case are almost equivalent. These are getting hard to read.

client/src/util/dataframe/dataframe.js
```diff
@@ -218,8 +233,54 @@ class Dataframe {
...
+    /*
+    Create a new DF, which is `this` plus the new column. Example:
+    const newDf = df.withCol("foo", [1,2,3]);
+
+    Dimensionality of new column must match existing dataframe.
+
+    Special case: empty dataframe will accept any size column.  Example:
+    const newDf = Dataframe.empty().withCol("foo", [1,2,3]);
+
+    If `withRowIndex` specified, it will be used for the new dataframe.
```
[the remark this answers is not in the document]
csweaver: After going through more of the files, this looks to be used with regard to creating the var annotations, but I could still use slightly more explanation.

client/src/util/dataframe/dataframe.js
```diff
@@ -94,17 +97,21 @@ class Dataframe {
...
+
+    /* check for expected types */
     if (!Array.isArray(columnarData)) {
       throw new TypeError("Dataframe constructor requires array of columns");
     }
     if (!columnarData.every(c => isArrayOrTypedArray(c))) {
       throw new TypeError("Dataframe columns must all be Array or TypedArray");
     }
+    if (!isLabelIndex(rowIndex) || !isLabelIndex(colIndex)) {
+      throw new TypeError("Dataframe index just be of label index type");
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/dataframe/dataframe.js
```diff
@@ -113,6 +120,11 @@ class Dataframe {
         "Dataframe dimension does not match column data shape"
```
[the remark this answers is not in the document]
bkmartinjr: that would require a lot more code given how I used the _.every() function.  I don't think it is worth it, given that this is sanity checking for internal users (who can read the code).    Frankly, we don't normally even do this level of error checking (I was just trying to be helpful).  Thoughts?

client/src/util/dataframe/dataframe.js
```diff
@@ -113,6 +120,11 @@ class Dataframe {
...
       );
     }
+    if (nRows !== rowIndex.size() || nCols !== colIndex.size()) {
+      throw new RangeError(
+        "Dataframe indices must have same size as underlying data"
```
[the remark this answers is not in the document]
bkmartinjr: I'm also not sure this is worth the extra code bloat, but I added it as it was a cut & paste.

client/src/util/dataframe/dataframe.js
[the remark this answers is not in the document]
bkmartinjr: Explanation: it provides the index to be used in the newly created dataframe.

Basically, there is nothing magic here -- it is just a factory method that creates a new dataframe, largely by modifying an existing one.   Sometimes the user wants to also provide a new index at the same time (common use case - reset the index to have different labels).

I'll augment the comments.

client/src/util/dataframe/dataframe.js
```diff
@@ -233,7 +294,7 @@ class Dataframe {
     return new Dataframe(dims, columnarData, null, null);
   }
 
-  __cut(rowOffsets, colOffsets) {
+  __cut(rowOffsets, colOffsets, withRowIndex) {
```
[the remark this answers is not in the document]
bkmartinjr: good catch - not sure how I missed that!

client/src/util/dataframe/dataframe.js
```diff
@@ -296,7 +358,15 @@ class Dataframe {
     return new Dataframe(dims, columns, rowIndex, colIndex);
   }
 
-  cutByList(rowLabels, colLabels = null) {
+  subset(rowLabels, colLabels = null, withRowIndex = null) {
+    /*
+    Cut on row/col labels.
```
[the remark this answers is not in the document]
bkmartinjr: good catch - will fix.

client/src/util/dataframe/labelIndex.js
```diff
@@ -74,6 +74,26 @@ class IdentityInt32Index {
     }
     return new DenseInt32Index(labelArray, [minLabel, maxLabel]);
   }
+
+  cut(labelArray) {
+    return this.__promote(labelArray);
+  }
+
+  with(label) {
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/dataframe/labelIndex.js
```diff
@@ -175,14 +212,33 @@ class KeyIndex {
...
-  getMaxOffset() {
+  size() {
     return this.rindex.length;
   }
 
   cut(labelArray) {
     return new KeyIndex(labelArray);
   }
+
+  with(label) {
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/dataframe/labelIndex.js
```diff
@@ -140,6 +160,20 @@ class DenseInt32Index {
     }
     return new DenseInt32Index(labelArray, [minLabel, maxLabel]);
   }
+
+  cut(labelArray) {
+    return this.__promote(labelArray);
+  }
+
+  with(label) {
```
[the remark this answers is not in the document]
bkmartinjr: will fix

client/src/util/stateManager/universe.js
```diff
@@ -51,29 +37,6 @@ These functions are used exclusively by the actions and reducers to
 build an internal POJO for use by the rendering components.
 */
 
-/*
-generate any client-side transformations or summarization that
-is independent of REST API response formats.
-*/
-function finalize(universe) {
```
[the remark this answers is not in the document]
bkmartinjr: Originally, there was a bunch of derived state that could only be computed once the universe was initialized.  This was a post-initialization step.  Later, all of that derived state was either unnecessary, or was memoized/computed on demand.

client/src/util/stateManager/universe.js
bkmartinjr: yes, I need to be better about my PR factoring....

client/src/actions/index.js
bkmartinjr: latent bug found by visual inspection.   Given that the code in this general area is changing, it made sense to fix it.

client/src/util/stateManager/world.js
bkmartinjr: it is obsolete - replaced by Dataframe.subset()

client/__tests__/util/dataframe/dataframe.test.js
bkmartinjr: I am open to suggestions....

bkmartinjr:
@csweaver - I have made most of the suggested changes/fixes, with a couple of small items that I either disagreed with or assumed was an editorial comment :-).    Please feel free to comment further post-merge and I will happily keep refining this.

client/src/util/dataframe/dataframe.js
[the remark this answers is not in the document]
csweaver: Feel free to ignore these; after seeing how your changes are used, improving the error messages are not worth the trouble. (the grammar one I still recommend fixing)

client/src/util/dataframe/dataframe.js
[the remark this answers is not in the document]
csweaver: not worth it

client/src/util/dataframe/dataframe.js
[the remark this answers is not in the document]
csweaver: Could you explain the use case when adding vars? I was having a little trouble getting the full picture.

client/src/util/stateManager/universe.js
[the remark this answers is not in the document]
csweaver: 👍

client/src/util/stateManager/world.js
csweaver: got it

client/__tests__/util/dataframe/dataframe.test.js
csweaver: I'll keep my eyes open. I don't have any ideas currently.

bkmartinjr: I try to use before/after when it makes sense, but in this case it just doesn't...

csweaver: It's the number of similar expect statements across tests that is overwhelming to my brain. It is fine technically, I just wish there was a better way.

client/src/util/dataframe/dataframe.js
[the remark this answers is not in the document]
bkmartinjr: resolved offline

client/src/reducers/controls.js
```diff
@@ -122,6 +121,46 @@ function createGenesDimMap(userDefinedGenes, diffexpGenes, world, crossfilter) {
...
+  VarDataCacheLowWatermark - this cofig value sets the minimum cache size,
+  in columns, below which we don't throw away data.
+
+  The value should be high enough so we are caching the maximum which will
+  "typically" be used in the UI (currently: 10 for diffexp, and N for user-
+  specified genes), and low enough to account for memory use (any single
+  column size is 4 bytes * numObs, so a column can be multi-megabyte in common
+  use cases).
+  */
+  const VarDataCacheLowWatermark = 32;
```
csweaver: What do you think about putting this value in globals (or maybe a constants file?) for ease of discovery and tweaking.

bkmartinjr: When a constant/config is used only in one place, I prefer them to be with the code they parameterize.    For me, moving them outside of the context in which they are used makes it harder to understand.    The counter example is a widely used constant/config - those should be centralized.

I realized this is an opinion.  If we establish a standard, I'm happy to follow it.

PR #615, opened 2019-02-26
create helper file for controls reducer

### LINKED ISSUE #611: controls reducer refactor helpers
the controls.js reducer file is getting a lot of in-file helper functions.  It is time to factor these out into a utils/stateManager/ helper file for clarity.

FYI: @colinmegill (this is in response to your PR #608 comments)

bkmartinjr:
Refactoring - clean up controls reducer by moving a lot of helper code out into a separate file.

Fixes #611


bkmartinjr:
@colinmegill - agree that we need a reducer refactor/cleanup. I'll drop you a note with some ideas and how I propose we do it.

client/src/reducers/controls.js
```diff
@@ -2,7 +2,7 @@
 
 import _ from "lodash";
 
-import { World, WorldUtil } from "../util/stateManager";
+import { World, WorldUtil, ControlsHelper } from "../util/stateManager";
```
csweaver: IMO the code would be easier to read if you imported the helper functions directly from the ControlsHelper files:

`import {createCategoricalSelectionState} from "../util/stateManager/ControlsHelper";`

```
# old 
ControlsHelper.createCategoricalSelectionState()
# new
createCategoricalSelectionState()
```

bkmartinjr: This is another case I find the opposite to be true, as I lose the connection to where the code is defined.    But I'm super happy to follow a standard if one is established.

Do you want to talk to Colin about codifying this stuff into some standard?   I generally think consistency trumps any given position :-).     I'll happily fix my code.

PR #618, opened 2019-02-27
[WIP DO NOT MERGE] suppress display of continous annotation without a finite extent

### LINKED ISSUE #617: For select continuous params, 'use as color scale' button not working
I have a dataset with a (relatively) large number of continuous parameters (not sure if this is related).  But several of the params, when I click on 'use as color scale', nothing happens.  The button is not toggled, and cells are not re-colored.  

Example here:
http://m9seq.mouse.gofflab.org/
- Clicking on 'cluster' param appropriately colors cells with the continuous cluster assignment
- Clicking on ['Pattern_1','Pattern_14','Pattern_17','Pattern_21'] as e.g. does not recolor.

Is this related to the distribution of the values for the param?  Looks like it might be specific to params more zeros or generally lower values?

bkmartinjr:
Occasionally continuous metadata (annotation) will only contain NaNs, and therefore not have a continuous range of [min, max].  It is not possible to build a histogram from the range [NaN, NaN].

This PR suppresses the display of the continuous annotation lacking either a finite min and max.   Or alternatively stated, it only displays a histogram where both min and max are finite values.

Fixes #617


client/src/components/continuous/continuous.js
```diff
@@ -68,8 +68,15 @@ class Continuous extends React.Component {
               const summary = obsAnnotations.col(key).summarize();
               const isColorField =
                 key.includes("color") || key.includes("Color");
+              const nonFiniteExtent =
+                summary.min === undefined || summary.max === undefined;
```
sidneymbell: I believe this line actually suppresses display of any continuous annotation with a nonfinite `min` _OR_ a nonfinite `max`?

Semantic difference from the PR description (suppress if min _and_ max are nonfinite -- i.e., _all_ values are `NaN`), but how does JS handle these values? 
If it mirrors python behavior, then this may throw an error if there are _any_ nonfinite values?

bkmartinjr: typo in the PR description; code is correct.  Great catch - I will fix the PR!

Essential idea is that you can't build a histogram unless you know the extent (ie, both min and max)

sidneymbell:
What about the scenario where we have a single NaN value for a gene?

bkmartinjr:
@sidneymbell - numeric annotations and expression data (ie, adata.obs or adata.X) are treated the same.  If there is any finite value in the column, we display.     For columns with mixed (some finite, some non-finite), the effect is varied:
- histograms display finite values in the bins, and do not display NaN/Inf (there has been discussion about summarizing these somewhere on the histogram, but probably not in a bin)
- embedding graph will "color by" NaNs in a light grey, which is visually distinct from the normal finite color range
- Scatterplot suppresses (does not plot) a cell if the X or Y value is non-finite

PR #625, opened 2019-03-02
Fix bug where prod would automatically run after release stage 2

csweaver:


bkmartinjr:
@csweaver - just after I approved, I noticed the same bug in the `release-burned` target.  Suggest fixing it everywhere...

csweaver:
good catch, fixed.

PR #626, opened 2019-03-02
Add forced directed graph drawing to allowed layout options

falexwolf:
Hi, thank you for the awesome package! :smile:

Can we add two canonical layouts, which are widely used in the community, to the allowed options? Force-directed graph drawing has many flavors, but the ForceAtlas2 layout and the Fruchtermann Reingold are among the most popular and most established.

The SPRING viewer from the Klein lab, uses FA2 by default. Within Scanpy, people compute these layouts using the `tl.draw_graph` tool.

Let me attach a comparison for the typical features of tSNE, UMAP, FA2 (and an FA2 layout that has been initialized using PAGA, but that's not the point here). This is Supplemental Figure 10 from the PAGA paper (https://doi.org/10.1101/208819, in press at Genome Biology).
image

You see that also UMAP has a tendency to produce disconnected manifolds when they are actually connected in high dimensions.

In Figure 1 of the PAGA paper, the first 3 layouts are force-directed layouts. They are faithful to topology and known biology (lineages in hematopoiesis):
image

Thanks for considering!
  Alex

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -14,7 +14,12 @@ from server.app.util.utils import custom_format_warning
 @click.command()
 @click.argument("data", metavar="<data file>", type=click.Path(exists=True, file_okay=True, dir_okay=False))
 @click.option(
-    "--layout", "-l", type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr"]), default="umap", show_default=True, help="Method for layout."
+    "--layout",
+    "-l",
+    type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr"]),
+    default="umap",
+    show_default=True,
+    help="Method for layout."
 )
 @click.option(
     "--diffexp",
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -14,7 +14,7 @@ from server.app.util.utils import custom_format_warning
 @click.command()
 @click.argument("data", metavar="<data file>", type=click.Path(exists=True, file_okay=True, dir_okay=False))
 @click.option(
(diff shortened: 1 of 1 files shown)


csweaver:
Hey @falexwolf! Wow! Thanks for adding this. Do you have a dataset I could test this with (or could you send me the scanpy command to generate these layouts for `example-dataset/pbmc3k.h5ad`)?

Would you like to add those options to the prepare command too?

I can handle updating the documentation.

server/cli/launch.py
```diff
@@ -14,7 +14,7 @@
 @click.command()
 @click.argument("data", metavar="<data file>", type=click.Path(exists=True, file_okay=True, dir_okay=False))
 @click.option(
-    "--layout", "-l", type=click.Choice(["umap", "tsne"]), default="umap", show_default=True, help="Method for layout."
+    "--layout", "-l", type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr"]), default="umap", show_default=True, help="Method for layout."
```
csweaver: Add a line break, please. Max line length is 120

falexwolf:
Yes, you can easily generate the layouts by running `sc.tl.draw_graph` on the pbmc3k dataset. In almost all settings, UMAP will do great. But in some settings, graph drawing might provide better embeddings.  And, as mentioned, some people simply prefer for taste. The whole of SPRING is limited to FA2.

By default `sc.tl.draw_graph` will run an FR algorithm; if `fa2` is installed, it will run a ForceAtlas2 algorithm; the logging output informs you.

I don't think you have to add it to the prepare commands. But people who use Scanpy and want to visualize or share the results post analysis using cellxgene, will often times have these layouts in their `.h5ad` files. And then it's nice to have the option of calling it, especially since it's just a tiny change to the allowed fields.

falexwolf:
Thank you! :)

PR #636, opened 2019-03-09
do not reset color-by when subsetting world

### LINKED ISSUE #621: "subset to current selection" clears differential expression results
what I was doing...

1. made two selections and ran differential expression
2. colored by a gene
3. selected a new subset
4. clicked "subset to current selection"

what happened...

- the plot was no longer colored by the gene
- the differential expression results were gone

what I expected...

- plot would maintain my color selection on subset
- differential expression results would still be available for selecting additional genes to color by

bkmartinjr:
Fixes #621.   When subsetting cell "world", do not perform a reset on "color by" mode.

Primary changes in this PR:
* color-by mode is preserved during subset operation.
* updateCellColors middleware retired, and replaced with a standard set of reducer helper functions in  `util/stateManager/colorHelpers.js`.
* small lint fixes in modified files.


client/src/reducers/controls.js
```diff
@@ -83,10 +89,9 @@ const Controls = (
       /* first light - create world & other data-driven defaults */
       const { universe } = action;
       const world = World.createWorldFromEntireUniverse(universe);
-      const colorRGB = new Array(universe.nObs).fill(
-        parseRGB(globals.defaultCellColor)
-      );
-      const categoricalSelectionState = ControlsHelper.createCategoricalSelectionState(
+      const colorMode = null;
+      const colors = createColors(world, colorMode);
```
colinmegill: Does this replace `newArray(foo).fill(parseRGB(bar))`

colinmegill: After following the chain into state manager it seems like the answer is yes... this is a new abstraction

colinmegill: Generally like it better than having to manually mutate here.

Issue #629, opened 2019-03-04
State & reducer refactoring


bkmartinjr:
Notes on the proposed redux state & reducer rework project.

*Goals:*
1. Add redo/undo support to app:

    * includes underlying capability to undo/redo *all* user actions (we may not
      expose all of these, but the underlying capability should support these):
        * data subset/un-subset (reset) actions (eg, world as a subset of universe)
        * selection actions, likely a subset (eg, record histogram selection, but not
          each mouse drag during selection)
        * user-defined gene add/remove, diffexp add/remove
        * plus future user features, eg, add Category/Label annotation editing
    * Ideally, implement redo/unto using a common Redux module, such as
      https://github.com/omnidan/redux-undo, described here:
      https://redux.js.org/recipes/implementing-undo-history     Using this will give
      us flexibility in the future, good documentation, and a well-tested module.
    * Connect redo/undo plumbing to an initial UI (Colin has initial UI mocks:
      https://www.figma.com/file/Y7T9LqGByddn1cKAVH16EUkU/Undo%2Fredo?node-id=0%3A1)

2. make the reducers more modular - in particular, the "controls" reducer is getting
   quite large/complex, and it would be nice if we could break it into independent
   reducers that are more in keeping with the Redux mindset.

   It is also likely that in the process of sorting out what actions are recorded
   in redo/undo history, we will need to refactor a bit.

3. Proof-of-principle action helpers suitable to enable two future use cases
   (not directly part of this project, but a good test for (1) and (2) above):
    * Remote notebook viewer: drive app remotely, eg, from a notebook plug-in, via
      RPC over a websocket (in short, RPCs sent over a websocket are transformed
      into actions, which are dispatched into the app).  Specific RPCs are TBD,
      but expected to include the equivalent of load data, reload data, and get
      current selection.
    * Headless use for taking snapshots via API.  Requirements TBD.  See issue #603

*Challenges to overcome, in order of difficulty:*

* [x] crossfilter currently is incompatible with most Redux-oriented approaches, as it is a _mutable_ data structure.   PR #647
* [x] decomposition of Controls reducer will require finding some modularity in the World/Universe data structure, but that has become much simpler with Dataframe.   PR #659
* [x] choice of redux helper (eg Redux Undo)    PR #659
* [ ] Dataframe subsets (ie, having multiple worlds on the redo stack) are not a view (subsets are a copy).   This uses a moderately large amount of memory, which means that each "world subset" in past or present history will use memory.

PR #647, opened 2019-03-14
refactoring - immutable crossfilter

bkmartinjr:
Refactor the crossfilter (selection state management) to be Redux-friendly.   

Changes:
* immutable API - any function which modifies the crossfilter returns a new object, and does not mutate the old.  Specifically, the following return a new crossfilter: constructor, `addDimension`, `delDimension` and `select`.
* crossfilter now encapsulates all dimension and index management, making it simpler to integrate into reducer state (in particular, you only need to maintain a reference to the top-level crossfilter, and no longer need to separately manage all dimension objects).   See `delDimension()` and `addDimension()` crossfilter methods
* some API renaming for clarity (largely s/filter/select/)
* there is a single "select" (aka filter) function, which now takes a selection specifier, eg, `crossfilter.select(dimName, {mode:"all"}`
* app and tests are all updated to match.

This is related to issue #629, and is prep for adding redo/undo.


client/src/reducers/controls.js
```diff
@@ -476,13 +482,15 @@ const Controls = (
 
       // update the filter to match all selected options
       const cat = newCategoricalSelectionState[action.metadataField];
-      state.dimensionMap[obsAnnoDimensionName(action.metadataField)].filterEnum(
-        ControlsHelpers.selectedValuesForCategory(cat)
-      );
-
+      const dName = obsAnnoDimensionName(action.metadataField);
```
csweaver: Since category metadata selected and deselected actions are almost identical, we could use fallthrough in the switch statement. Only one line looks different: 
select - `newCategorySelected[action.categoryIndex] = true;`
deselect - `newCategorySelected[action.categoryIndex] = false;`

client/src/util/typedCrossfilter/index.js
```diff
@@ -11,781 +9,26 @@ time/space optimizations predicated upon the following assumptions:
...
-    Typically, data is one of:
-      - Array of objects/records
-      - Dataframe (util/dataframe)
-    */
-    this.data = data;
-
-    // filters: array of { id, dimension }
-    this.filters = [];
-    this.selection = new BitArray(data.length);
-    this.updateTime = 0;
```
csweaver: removing this seems like a real win

client/__tests__/util/typedCrossfilter/crossfilter.test.js
```diff
@@ -0,0 +1,330 @@
...
+
+    expect(payments).not.toBe(p2);
+    const p3 = p2.select("quantity", { mode: "all" });
+    expect(p3).not.toBe(p2);
+
+    const p4 = p3.delDimension("quantity");
+    expect(p4).not.toBe(p3);
+  });
+
+  test("select all and none", () => {
```
csweaver: This test seems a little long, Could you break it up into 1) creation, 2) select all, 3) select none?

client/src/util/typedCrossfilter/bitArray.js
```diff
@@ -211,6 +211,19 @@ class BitArray {
     }
   }
 
+  // select range of indices on a dimension
```
csweaver: My bit manipulation isn't strong enough to properly review this code

client/__tests__/util/typedCrossfilter/crossfilter.test.js
```diff
@@ -0,0 +1,330 @@
...
+      new Uint8Array(someData.length).fill(1)
+    );
+    expect(p.fillByIsSelected(new Uint8Array(someData.length), 99, 0)).toEqual(
+      new Uint8Array(someData.length).fill(99)
+    );
+    for (let i = 0; i < someData.length; i += 1) {
+      expect(p.isElementSelected(i)).toBeTruthy();
+    }
+
+    /* expect a selectAll on one dimension to chnage nothing */
```
csweaver:
```suggestion
    /* expect a selectAll on one dimension to change nothing */
```

client/__tests__/util/typedCrossfilter/crossfilter.test.js
```diff
@@ -0,0 +1,330 @@
...
+
+  test("select all and none", () => {
+    let p = payments
+      .addDimension("quantity", "scalar", (i, d) => d[i].quantity, Int32Array)
+      .addDimension("tip", "scalar", (i, d) => d[i].tip, Float32Array)
+      .addDimension("total", "scalar", (i, d) => d[i].total, Float32Array)
+      .addDimension("type", "enum", (i, d) => d[i].type);
+    expect(p).toBeDefined();
+
+    /* expect all records to be selecte - default init state */
```
csweaver:
```suggestion
    /* expect all records to be selected - default init state */
```

client/src/util/typedCrossfilter/crossfilter.js
```diff
@@ -0,0 +1,570 @@
...
+  lowerBound,
+  lowerBoundIndirect,
+  upperBoundIndirect
+} from "./util";
+
+class NotImplementedError extends Error {
+  constructor(...params) {
+    super(...params);
+
+    // Maintains proper stack trace for where our error was thrown (only available on V8)
```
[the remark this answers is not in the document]
bkmartinjr: sorry, don't understand the question...can you elaborate?

If you are asking about the `captureStackTrace`, see https://v8.dev/docs/stack-trace-api#stack-trace-collection-for-custom-exceptions

client/src/util/typedCrossfilter/crossfilter.js
```diff
@@ -0,0 +1,570 @@
...
+  }
+
+  /* eslint-disable class-methods-use-this */
+  selectRange() {
+    throw new Error("range selection unsupported on Enumerated dimension");
+  }
+  /* eslint-enable class-methods-use-this */
+}
+
+class ImmutableSpatialDimension extends _ImmutableBaseDimension {
```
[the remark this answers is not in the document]
bkmartinjr: If they have X/Y coordinates, it will work fine.  I'm not aware of other embeddings we might be considering, but we could add additional dimension types as needed.   Keep in mind that other modules in the app (eg, the graph component) have similar assumptions and would also need to change.

client/src/reducers/controls.js
bkmartinjr: Fall throughs are almost always (IMHO) a bad idea, as they save a bit of code but tend to surprise the reader unless they are paying close attention.

client/__tests__/util/typedCrossfilter/crossfilter.test.js
bkmartinjr: creation is already separate.

splitting all/none would require two tests that are almost as long as the current one, as the test requires multiple steps (ie, compound state).   In other words, they are not independent tests.

client/src/util/typedCrossfilter/bitArray.js
bkmartinjr: there are no significant changes to this code in quite some times.   I added a new interface in this PR, but the logic is identical.

client/__tests__/util/typedCrossfilter/crossfilter.test.js
csweaver: gotcha

client/src/util/typedCrossfilter/bitArray.js
csweaver: 👍

client/src/reducers/controls.js
```diff
@@ -95,61 +94,57 @@ const Controls = (
...
         world
       );
-
-      /* free dimensions not in cache (otherwise they leak) */
-      _.forEach(state.dimensionMap, (dim, dimName) => {
-        if (!fullUniverseCache.dimensionMap[dimName]) {
-          dim.dispose();
-        }
-      });
-      const dimensionMap = {
```
colinmegill: Confirming: dimensionMap as a concept is no longer around?

bkmartinjr: correct.  The equivalent is now part of the crossfilter object itself, and you reference dimensions by name.

Eg,  `crossfilter.addDimension("foo", ...)`

PR #650, opened 2019-03-15
Procedurally resize brush

### LINKED ISSUE #587: clearing histogram selection often results in tiny selection
I have commonly found that when I click in a histogram, to clear the brush selection, I end up creating a super tiny, hard to see, selection.    It is likely that this is caused by a small drag distance between the mousedown & mouseup events.

Suggestion:  we need a minimum brush distance, used to differentiate a click-to-clear and a click-to-select.

This may also be an issue for the new poly lasso.

colinmegill:
Fixes #587 

resize

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -128,34 +128,31 @@ class HistogramBrush extends React.Component {
     return () => {
       const { dispatch, field, isObs, isUserDefined, isDiffExp } = this.props;
       const { brushX } = this.state;
+      const minAllowedBrushSize = 50;
 
       if (d3.event.selection) {
         let _range;
 
-        if (d3.event.selection[1] - d3.event.selection[0] < 100) {
+        if (
+          d3.event.selection[1] - d3.event.selection[0] >
+          minAllowedBrushSize
+        ) {
           _range = [x(d3.event.selection[0]), x(d3.event.selection[1])];
         } else {
           /* the user selected range is too small and will be hidden #587, so take control of it procedurally */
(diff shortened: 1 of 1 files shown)


client/src/components/brushableHistogram/index.js
```diff
@@ -124,6 +124,63 @@ class HistogramBrush extends React.Component {
...
+
+          _range = [x(d3.event.selection[0]), x(procedurallyResizedBrushWidth)];
+
+          d3.event.target.move(brushX, [
+            d3.event.selection[0],
+            procedurallyResizedBrushWidth
+          ]);
+        }
+
+        dispatch({
```
bkmartinjr: style comment - the two dispatch() calls are largely redundant and could be combined.  Just pop _range up a scope, and initialize it to null.

client/src/components/brushableHistogram/index.js
```diff
@@ -124,6 +124,63 @@ class HistogramBrush extends React.Component {
...
+      const { dispatch, field, isObs, isUserDefined, isDiffExp } = this.props;
+      const { brushX } = this.state;
+      const minAllowedBrushSize = 10;
+      const smallAmountToAvoidInfiniteLoop = 0.1;
+
+      if (d3.event.selection) {
+        let _range;
+
+        if (
+          d3.event.selection[1] - d3.event.selection[0] >
```
bkmartinjr: question: do you need to abs() this before comparison (ie, can selection[1] be smaller than selection[0])?

colinmegill: I believe d3 handles this internally, because swiping both left and right works.

Issue #627, opened 2019-03-04
Reset does not clear histogram selection brush


bkmartinjr:
If you do a Reset (via Reset button), selection brushes on histograms are not cleared.

To reproduce:
1. load a dataset, eg, pbmc3k
2. brush select a section of any continuous metadata (eg, 'n_genes' in pbmc3k)
3. click Reset button

the entire UI, except for the histogram brush, is reset.   The brush should be cleared.

Update: this same underlying bug affects Undo/Redo.

PR #659, opened 2019-03-20
Undo/redo

bkmartinjr:
Implement undo/redo user functions.   Changes:
* Refactor reuducers, using a new "cascading" meta reducer.  
    * The `controls` reducer is now split into: universe, world, categoricalSelectionState, crossfilter, colors and controls (misc UI state)
    * net effect is that the factored reducers (eg, universe) are now in `state.universe` rather than `state.controls.universe`
* Add undo/redo support, which includes:
    * the new undoable meta reducer which manages the state history
    * configuration of which actions side-effect the history
    * Rework a few actions and action generators to improve history UI - ie, allow history to change appropriately based upon user action
    * add Redo/Undo buttons to the graph component

Initial cut at UI:
* Reset will clear the history
* All user actions should be undoable/redoable

There remain a number of edge cases & known issues:
* brushable histogram does not correctly clear its brush state when reset (via either Reset or Undo).   #627 
* lasso selector does not correctly update its paint state in Undo or Redo (the selection is correctly handled -- this is only an issue with the lasso layer painting).  #663


client/src/components/brushableHistogram/index.js
```diff
@@ -15,13 +15,13 @@ import actions from "../../actions";
...
+  world: state.world,
   scatterplotXXaccessor: state.controls.scatterplotXXaccessor,
   scatterplotYYaccessor: state.controls.scatterplotYYaccessor,
-  crossfilter: state.controls.crossfilter,
+  crossfilter: state.crossfilter,
   differential: state.differential,
-  colorAccessor: state.controls.colorAccessor,
-  obsAnnotations: _.get(state.controls.world, "obsAnnotations", null)
+  colorAccessor: state.colors.colorAccessor,
+  obsAnnotations: _.get(state.world, "obsAnnotations", null)
```
colinmegill: `state.world` is really nice.

client/src/components/brushableHistogram/index.js
```diff
@@ -289,7 +290,8 @@ class HistogramBrush extends React.Component {
       .call(
         d3
           .brushX()
-          .on("brush", this.onBrush(field, x.invert).bind(this))
+          .on("start", this.onBrush(field, x.invert, "start").bind(this))
+          .on("brush", this.onBrush(field, x.invert, "brush").bind(this))
```
colinmegill: Requesting a comment here re: why all 3 are necessary

client/src/components/categorical/categorical.js
```diff
@@ -6,7 +6,7 @@ import * as globals from "../../globals";
 import Category from "./category";
 
 @connect(state => ({
-  categoricalSelectionState: state.controls.categoricalSelectionState
+  categoricalSelectionState: state.categoricalSelectionState
```
colinmegill: A point for later, but `categoricalSelection` is probably the right name here, state is redundant.

client/src/reducers/categoricalSelectionState.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    case "categorical metadata filter all of these": {
+      const newCategoricalSelectionState = {
+        ...state,
+        [action.metadataField]: {
+          ...state[action.metadataField],
+          categorySelected: Array.from(
+            state[action.metadataField].categorySelected
+          ).fill(true)
```
colinmegill: Suggest a comment on this line for posterity

client/src/reducers/crossfilter.js
```diff
@@ -0,0 +1,189 @@
...
+            "scalar",
+            world.varData.col(gene).asArray(),
+            Float32Array
+          ),
+        state
+      );
+      return crossfilter;
+    }
+
+    case "clear differential expression": {
```
colinmegill: Do we think this belongs in the crossfilter reducer?

client/src/components/brushableHistogram/index.js
bkmartinjr: will add.    TL;DR - needed for History.

The essence is that we want to know when the drag starts so we can push the Redux state onto the history stack.   We don't want to push new history on each brush/drag, but only on the first one.

client/src/components/categorical/categorical.js
bkmartinjr: I'll fix, but for the record, I used the name that was previously in the code (ie, I didn't change it).

colinmegill: Yup, this is my name

client/src/reducers/categoricalSelectionState.js
bkmartinjr: I'm not sure what you are proposing - can you elaborate?

client/src/reducers/crossfilter.js
bkmartinjr: yes, it does.    All code that changes the crossfilter state is in this reducer.

Note that we also handle this action in other reducers, to change other state.

client/src/reducers/categoricalSelectionState.js
colinmegill: This is one of the 'chewier' lines, those reading from a non JS background might appreciate a plaintext reading of this operation. If you disagree feel free to skip.

client/src/components/categorical/categorical.js
bkmartinjr: fixed.

client/src/reducers/categoricalSelectionState.js
bkmartinjr: lol, you wrote this code :-)

I'll add comments.

bkmartinjr: OK, added a comment about the intent (to all four places we do something similar).  LMK if that resolves it, or propose a comment if you want something different (I wasn't sure if "chewie" referred to the intent or the code).

client/src/reducers/crossfilter.js
colinmegill: Right, yes, that's great.

client/src/reducers/categoricalSelectionState.js
colinmegill: Those are perfect!

PR #662, opened 2019-03-21
Support `diffmap` and `phate` layouts. Explicitly handle embeddings with >2 components.

### LINKED ISSUE #660: Support diffmap and phate layouts (i.e., basic trajectory visualization support)
We currently support most of the embeddings computed by scanpy, with the exception of `diffmap` and `phate`. 

The most basic version of support for trajectory visualization laid out in the planning docs consists of:
* `dpt_groups`: trajectory branch assignment, treated as categorical metadata
* `dpt_pseudotime`: pseudotime assignment, treated as continuous metadata  
* embedding in `phate` or `diffmap` coordinates  

`dpt_groups` and `dpt_pseudotime` are both treated as metadata and handled/visualized appropriately in the app as-is (if these values have been precomputed in scanpy and attached to the h5ad).

Thus, by supporting `X_diffmap` and `X_phate` layouts, we can easily provide a basic level of support for visualizing trajectories.

### LINKED ISSUE #661: Layout launch fails when run with `n_comps > 2`
All of the embedding methods in scanpy have the option to return > 2 components. The current `launch` process just takes the entire array (so far it doesn't break because most scanpy methods default to `n_comps = 2`). But, if `n_comps > 2`, this can cause a layout launch error. We just need to explicitly pull the first two components of `adata.obsm['X_layout']` on launch.

sidneymbell:
This PR fixes a potential bug in how we load layouts (Fixes #661). It also provides basic support for trajectory visualization (Fixes #660).  

The app already supports visualization of cells' positions in pseudotime (via continuous metadata field `dpt_pseudotime`) and trajectory branch assignments (via categorical metadata field `dpt_group`).  

By supporting `diffmap` and `phate` as supported layouts, we can enable users to explore the output of most of the trajectory inference methods implemented in the current version of scanpy:

image 

image

In the process of adding support for these, I also discovered that we don't explicitly pull the first two components of a given embedding (this is relevant for the currently-supported layouts, as well, as umap and tsne can also be exported with >2 components). Easy fix to head off any future confusion.

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -16,7 +16,7 @@ from server.app.util.utils import custom_format_warning
 @click.option(
     "--layout",
     "-l",
-    type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr"]),
+    type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr", "diffmap", "phate"]),
     default="umap",
     show_default=True,
     help="Method for layout."
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ You should see your web browser open with the following
 
 There are several options available, such as:
 
-- `--layout` to specify the layout as `tsne`, `umap`, `draw_graph_fa`, or `draw_graph_fr` 
+- `--layout` to specify the layout as `tsne`, `umap`, `diffmap`, `phate`, `draw_graph_fa`, or `draw_graph_fr`
 - `--title` to show a title on the explorer
 - `--open` to automatically open the web browser after launching (OS X only)
 
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -325,14 +325,11 @@ class ScanpyEngine(CXGDriver):
         * only returns Matrix in columnar layout
         """
         try:
-            import click
-            click.echo('click echo worked')
-            print('print worked')
-            raise KeyError('error raising works')
-            df_layout = self.data.obsm[f"X_{self.layout_method}"]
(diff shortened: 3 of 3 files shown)


csweaver:
One more thing. I think we should warn users if we truncate part of their layout, could we add a check on load?

server/cli/launch.py
```diff
@@ -16,7 +16,7 @@
 @click.option(
     "--layout",
     "-l",
-    type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr"]),
+    type=click.Choice(["umap", "tsne", "draw_graph_fa", "draw_graph_fr", "diffmap", "phate"]),
```
bkmartinjr: Why do we restrict the user to this set of layouts?   I would advocate for letting the user create and view their own custom layout, if they are so inclined.  Definitely an advanced user feature, but super easy to add and very useful.

sidneymbell:
Will do! Thanks Charlotte. OK to merge once those are implemented?

server/cli/launch.py
csweaver: Hmm... I've thought of the same thing before. I like how having options leads to discoverability and protects against non-viewable layouts. But I also agree that advanced users shouldn't be artificially limited.

sidneymbell:
@csweaver - Because I failed to notice that the linter did silly things to the whitespace. Fixed!

server/cli/launch.py
sidneymbell: All good points. As of now, I believe we support / enable all embeddings implemented in scanpy. Any additional / "custom" layouts would have to be computed externally and then attached manually to the anndata structure. Could happen, but I suspect even for advanced users it's not super common? Definitely something we should think about if / when we're ready to support `launch` input from other ecosystems, though.

PR #669, opened 2019-03-21
Better smoke tests

csweaver:
Completed smoke tests using puppeteer. Testing only things that can be accessed via the UI - I'm not tapping into the store. 

Tests - Testing main user workflow and ensuring values are correct and no javascript errors appear on the page. I'm not checking the svgs or canvases. Tests take approx 20s to run after server starts with pmbc3k dataset. 
Test Config - I'm configuring from env variables so that you can watch the tests run when developing locally. Tests are run with dataset specific values to ease adding additional datasets to the smoke tests. 
Puppeteer - callback hell has a successor - async/await hell. Most bugs that happen when writing test are 1) forgot to put an async or an await in the right place or 2) using an iterator that doesn't support async iterator (I'm looking at you forEach). I've written some utils to do some common actions. `puppeteerUtils` for any testing and `cellxgeneActions` for cellxgene specific stuff. This significantly cleans up the readability of the tests.
Test ids/classes - I've added/changed some of the test-ids and test-classes on the client. Changes were made to ensure that the name is more robust to UI changes (removing svg/canvas/button) or to match a use I didn't know I needed. 
Categorical Selection - I've made a UX change to the select all box in categorical selection. When in an indeterminate state, the select all box used to go to either select none or select all depending on if you had used it to select all or select none most recently. I've made it always select all if starting from an indeterminate state. This was necessary in order to be able to use puppeteer to reset the category without knowing how it got to the current state.  

Future PRs are coming for:
- starting the server using the `jest-puppeteer` package to enable some cool features it has
- adding additional datasets to be tested
- [possibly] slack notifications, snapshots of canvas elements to use for visual comparison, tests that read into the redux store.


csweaver:
tagging @smadness since he's doing end to end tests for meta

client/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "npm run clean && webpack --config configuration/webpack/webpack.config.prod.js",
     "clean": "rimraf build",
     "dev": "npm run clean && webpack --config configuration/webpack/webpack.config.dev.js",
-    "e2e": "jest e2e",
+    "e2e": "jest --verbose false --config __tests__/e2e/e2e-jest-config.json e2e",
```
bkmartinjr: Nit:  the change to the command line `unit-test` target (which happened on a prior PR) broke the ability to run single tests from the command line.   We used to be able to ay `npm test client/__tests__/foo.test.js` and it would run just that single test.   It looks like the way `jest` interprets the --testPathIgnorePatterns is incompatible with that.   If you see an easy fix, it would make it very convenient for unit test authors (the alternative is to use `node_modules/.bin/jest testname`.

client/__tests__/e2e/puppeteer-utils.js
```diff
@@ -0,0 +1,161 @@
...
+    for (const val of values) {
+      await puppeteerUtils(puppeteerPage).clickOn(
+        `categorical-value-select-${category}-${val}`
+      );
+    }
+  },
+
+  async reset() {
+    const reset_button = await puppeteerUtils(puppeteerPage).clickOn("reset");
+    // loading state never actually happens, reset is too fast
```
bkmartinjr: my favorite comment of the day :-)

client/__tests__/e2e/e2e-jest-config.json
```diff
@@ -0,0 +1,10 @@
+{
```
bkmartinjr: we are configuring the unit test jest inside package.json, but doing it as stand-alone files for puppeteer.     Should they both use the same method?   (I don't know the answer - just asking why we would do it differently)

client/__tests__/e2e/e2e.test.js
```diff
@@ -31,161 +39,244 @@ afterAll(() => {
...
+        cellset["coordinates-as-percent"]
+      );
+      await cxgActions.drag(hist_id, coords.start, coords.end);
+      const cell_count = await cxgActions.cellSet(1);
+      expect(cell_count).toBe(cellset.count);
+    }
+  });
+});
+
+describe("gene entry", async () => {
```
bkmartinjr: how much do you want to test error handling (at least in the common cases)?   Seems like it might be worthwhile adding a test for a) a known bad gene name, b) a known badly formatted bulk gene list (eg,   "foo;bar;baz") and ensure a toast/error appears.

client/src/components/brushableHistogram/index.js
```diff
@@ -385,7 +385,13 @@ class HistogramBrush extends React.Component {
       <div
         id={`histogram_${field}`}
```
bkmartinjr: I am not sure `field` is guaranteed to be anything other than a string, which means it could contain characters that are not legal in an element ID.    Might consider escaping these for the (rare) case we get a dataset with fields such as "foo bar"

bkmartinjr: actually, now that I look more, I think this is definitely needed - at a minimum space needs to be handled, as it is not legal in an ID, and seems to be legal (and not rare) as a column name.

client/__tests__/e2e/e2e-jest-config.json
csweaver: The jest-puppeteer package has configuration that break the presets for the unit tests. My understanding is that most future tests will be unit tests and so I feel like that should be the default configuration. I wish there was an easy way to set up alternate configurations for tests in different folders but my googling didn't bring up anything. Totally open to suggestions

client/__tests__/e2e/e2e.test.js
csweaver: @colinmegill didn't like checking on the toasts themselves. I also thought we are likely moving to error collapsing soon so I wasn't sure that this behavior was stable enough to test. What do you think?

client/src/components/brushableHistogram/index.js
csweaver: Will fix (but for the record, I didn't add this :)

client/package.json
csweaver: I bet I can fix this. Let me look into it.

client/__tests__/e2e/puppeteer-utils.js
csweaver: 😁

csweaver:
I actually don't think we should have to escape special characters for the `data-testid` These are [data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) and thus don't have the same restrictions on their values. Using id may be misleading, I went with `testid` for values that should be unique and `testclass` for values that may apply to many elements. 

I do agree on escaping special characters in the actual `ids`, though.

client/__tests__/e2e/e2e-jest-config.json
bkmartinjr: I wasn't proposing an identical config.  Rather that we have the structure of the config be the same (ie, all in package.json, or all in stand-alone files).   

That said, this is a super tiny issue, so if not easy, skip it.

client/__tests__/e2e/e2e.test.js
bkmartinjr: fair point about the toasts.

But knowing that bad data at least doesn't generate an error on the console is worth something....    There are not many places we accept user data entry - might as well have some "bad data" tests for them - ie, just confirm that we seem to survive the bad data.

Short term, this may require dismissing toasts, so I'm OK if you merge this without adding these tests.  We can augment the tests as the toasts stabilize.

client/src/components/brushableHistogram/index.js
bkmartinjr: lol, sorry!   And I'm only really concerned about places that are really going to cause a problem with legit user data.  So I think only `id` requires the space-escape.

bkmartinjr:
I'm good with your proposed (escape only ID) solution!

client/__tests__/e2e/e2e-jest-config.json
csweaver: I was thinking of it as the unit test config should be the default (thus in package.json) and the smoke test should be the exception (thus in a separate file). IMO future exceptions should have separate files as well.

client/package.json
csweaver: I added a script named "test" that just runs `jest` (this is the way it used to be) to fix your cli workflow.  We just won't use this script in the CI.

bkmartinjr: sounds reasonable.  if I understand, this will run all tests if executed as `npm run test`, but with a specific test argument, will work as it used to.   Thanks!

client/__tests__/e2e/e2e-jest-config.json
bkmartinjr: SGTM

csweaver:
@smadness thanks again for taking a look! I've just made a separate PR testing some error state, it was a little more complicated to attach ids to the error toasts so I didn't want to lump that in with this pr. I'll probably add more tests as we exercise these and get some feedback about how they break.

PR #670, opened 2019-03-22
brushable histogram brush selection responds to state updates

### LINKED ISSUE #627: Reset does not clear histogram selection brush

bkmartinjr:
Fixes #627 

The brushable histogram selection "brush" was not integrated with redux state, and would therefore not correctly clear on an undo or reset.

This change includes:
* add a `continousSelection` reducer and state to hold the brushable histogram state
* HistogramBrush component now responds to state changes by removing or moving its brush

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -106,7 +106,11 @@ class HistogramBrush extends React.Component {
         const selection = d3.brushSelection(brushXselection.node());
         if (!range && selection) {
           brushXselection.call(brushX.move, null);
-        } else if (range) {
+        } else if (range && !selection) {
+          const x0 = x(range[0]);
+          const x1 = x(range[1]);
+          brushXselection.call(brushX.move, [x0, x1]);
+        } else if (range && selection) {
           const moveDeltaThreshold = 1;
           const x0 = x(range[0]);
           const x1 = x(range[1]);
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -94,6 +94,10 @@ class HistogramBrush extends React.Component {
       this.renderAxesBrushBins(x, y, bins, numValues, svgRef, field);
     }
 
+    /*
+    if the selection has changed, ensure that the brush correctly reflects
+    the underlying selection.
+    */
     if (continuousSelection !== prevProps.continuousSelection) {
       const { isObs, isUserDefined, isDiffExp } = this.props;
       const myName = makeContinuousDimensionName(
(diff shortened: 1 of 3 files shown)


client/src/reducers/continuousSelection.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+    case "continuous metadata histogram start":
+    case "continuous metadata histogram brush":
+    case "continuous metadata histogram end": {
+      const name = makeContinuousDimensionName(
+        action.continuousNamespace,
+        action.selection
+      );
+      const continuousSelection = { ...state };
+      continuousSelection[name] = action.range;
```
csweaver: You could use computed property names `computed key` to keep the state return statement in the usual fashion:
```
const continuousSelection = { ...state, [name]: action.range };
```

client/src/reducers/index.js
```diff
@@ -63,6 +64,7 @@ const Reducer = undoable(
     ["universe", universe],
     ["world", world],
     ["categoricalSelection", categoricalSelection],
+    ["continuousSelection", continuousSelection],
```
csweaver: This data structure seems like it could be simplified since the strings and variables are always the same.

client/src/components/brushableHistogram/index.js
```diff
@@ -86,19 +87,51 @@ class HistogramBrush extends React.Component {
...
+      const { isObs, isUserDefined, isDiffExp } = this.props;
+      const myName = makeContinuousDimensionName(
+        { isObs, isUserDefined, isDiffExp },
+        field
+      );
+      const range = continuousSelection[myName];
+      const { brushXselection, brushX } = this.state;
+      if (brushXselection) {
+        const selection = d3.brushSelection(brushXselection.node());
+        if (!range && selection) {
```
csweaver: Suggest commenting the conditions. What are the scenarios that each get called? I'm having trouble figuring that out.

client/src/reducers/continuousSelection.js
bkmartinjr: will fix

client/src/reducers/index.js
bkmartinjr: could you suggest something concrete?   I don't think we can assume that the functions and fields have the same name, but they often will, and I'm open to improving this.  I just couldn't come up with a concrete idea!

client/src/components/brushableHistogram/index.js
bkmartinjr: comments added - please review.

csweaver: much better 👍

client/src/reducers/index.js
csweaver: I'm thinking on it. This should block the merge.

PR #680, opened 2019-03-29
add scripts from cli

### LINKED ISSUE #678: Add option to add script files in template
CLI option to include external script files in index_template.html. Possible use case is a user adding google analytics, etc. Needs to work with docker build.

csweaver:
Fixes #678 

Able to add scripts to the html from launch CLI with --scripts flag. This means it will be accessible from the CLI and from docker since docker runs the `cellxgene` CLI.

I reformatted the html file in the process. I could make that a separate PR, but I think it looks nicer now 😁

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -123,6 +123,8 @@ def launch(
 The --scripts flag is intended for developers to include google analytics etc. You could be opening yourself to a
 security risk by including the --scripts flag. Make sure you trust the scripts that you are including.
         """)
+        scripts_pretty = ", ".join(scripts)
+        click.confirm(f"Are you sure you want to inject these scripts: {scripts_pretty}?", abort=True)
 
     if not verbose:
         sys.tracebacklimit = 0
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -73,20 +73,20 @@ from server.app.util.utils import custom_format_warning
     show_default=True,
 )
 def launch(
-    data,
-    layout,
-    diffexp,
-    title,
-    verbose,
-    debug,
-    obs_names,
-    var_names,
-    open_browser,
-    port,
-    host,
-    max_category_items,
-    diffexp_lfc_cutoff,
-    scripts,
+        data,
+        layout,
+        diffexp,
+        title,
+        verbose,
+        debug,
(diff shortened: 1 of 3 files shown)


csweaver:
Added security warning message

server/cli/launch.py
```diff
@@ -113,6 +113,17 @@ def launch(
...
+    if scripts:
+        click.echo("""
+/ / /\ \ \__ _ _ __ _ __ (_)_ __   __ _
+\ \/  \/ / _` | '__| '_ \| | '_ \ / _` |
+ \  /\  / (_| | |  | | | | | | | | (_| |
+  \/  \/ \__,_|_|  |_| |_|_|_| |_|\__, |
+                                  |___/
+The --scripts flag is intended for developers to include google analytics etc. You could be opening yourself to a
+security risk by including the --scripts flag. Make sure you trust the scripts that you are including.
+        """)
```
ttung: I would further make people type in a confirmation that they understand this message before spawning the browser.

bkmartinjr: We have users who want to deploy in a docker-deployed environment, ie, scripted startup where they have control over the deployment script.   I like Tony's suggestion, but we need to ensure that the confirmation can be scripting (eg, pipe `yes` to it, or whatever).    Alternative to "|yes": provide a `--I-attest-that-really-mean-to-inject-scripts` CLI param that still prints the warning, but dispenses with the interactive prompt.    @ttung - thoughts?

csweaver: I should be able to add this such that `| yes` works.

ttung: Either way is fine.  The browser being fired up is sufficient, I believe, for bad stuff to run, so I do think the confirmation is important, at least for the first time.

csweaver: Added confirm.

`yes | cellxgene launch --scripts foo --scripts bar example-dataset/pbmc3k.h5ad` bypasses the confirm as expected.

csweaver:
From Sara on the security team:

> I think this works. It doesn’t open up any major security concerns. I think if this becomes an ongoing ask then we should look at other options but as a one off the warning works 

Looks ok, as long as we keep an eye on it.

Issue #542, opened 2019-01-07
select from all genes in dataset


Marius1311:
Hi all!

I have been using cellxgene browsers for a while now and I really like them. One feature I am missing is access to the .raw attribute of AnnData objects. When I use an AnnData object for a cellxgene browser, I only have access to genes which are contained in the .X layer of the object. However, we usually filter our AnnData objects to only contain highly variable genes before computing embeddings - meaning that I won't be able to color in the majority of genes via the cellxgene broweser because they are not in the .X layer, but also in the .raw layer which I initialise before filtering. This is a really important feature when working with biological cooperation partners who would like to color in genes they know. Currently, my cooperation partners cannot find these genes because they have been filtered out and there is no .raw access. 

All the best,
Marius

csweaver:
Hi @Marius1311! I'm happy to hear that you've been using cellxgene! Thank you for the feedback, we will be sure to discuss this at our team meeting this week. 

A couple of questions: 1. When you are accessing a gene from `.raw` do you want to see the original observations or would you like the observations normalized in some way? 2. If you want to see the original observations, would you ever want to be able to access these for the genes that appear in .X as well?

Marius1311:
Hi @csweaver, and thanks for your quick reply!
At which step in the process of normalisation and filtering one would like to set the .raw attribute is still up for debate, but I think most scanpy users first do some very basic filtering (remove a couple of cells which are very likely to be dead), normalise per cell and log transform. When the data is in this state, we save to adata.raw, which is what we use later on to visualise individual genes and also for statistical testing like differential gene expression because we would usually not want to use the batch corrected data for that. Please see this scanpy tutorial [Clustering 3K PBMCs](https://scanpy-tutorials.readthedocs.io/en/latest/pbmc3k.html) for an illustration of what I just said. Saving to .raw happens in line 15.

Regarding your second point: When I color in genes, I would always like to see their expression colored in according to their values in adata.raw, regardless of whether they appear in adata.X as well or have been filtered out.

Marius1311:
Hi @GenevieveHaliburton ! Thanks for your message. There aren't really any other features from .raw I would like to vizualise, it is manly really the filtered out genes. Filtered cells would be difficult I guess, because where would they appear in the embedding? In a PCA, one could just project them into the embedding but that's not possible in a tsne and at least not trivial in a UMAP. 

When I visualise gene expression, I would usually use normalised, but not batch corrected data - that's what I usually store in .raw

[...]

PR #688, opened 2019-04-02
Update scanpy version

sidneymbell:
Update scanpy version to 1.3.7.

Latest is 1.4, but all the features we want to integrate were in by 1.3.7 (most specifically the `import scanpy` instead of `import scanpy.api`).

### DIFF
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -2,7 +2,7 @@ import warnings
 
 import numpy as np
 from pandas.core.dtypes.dtypes import CategoricalDtype
-import scanpy.api as sc
+import scanpy as sc
 
 from server.app.driver.driver import CXGDriver
 from server.app.util.constants import Axis, DEFAULT_TOP_N
--- a/server/cli/prepare.py
+++ b/server/cli/prepare.py
@@ -63,7 +63,7 @@ def prepare(
     import matplotlib
 
     matplotlib.use("Agg")
-    import scanpy.api as sc
+    import scanpy as sc
 
     # scanpy settings
     sc.settings.verbosity = 0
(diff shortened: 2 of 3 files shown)


bkmartinjr:
@sidneymbell - has this also been tested against scanpy 1.4?

neuromusic:
does this fix #542?

csweaver:
@neuromusic short answer: No. Longer answer - this allows us to fix #542 but see #689 for some of the concerns we need to work out before we actually fix it.

sidneymbell:
@jk - No. Please see #689.

neuromusic:
I thought that the "latest scanpy" (is that 1.37? 1.4?) no longer drops genes from `.X` when performing louvain, etc, effectively solving the core issue in #542, since all genes are now present in `.X` and therefore available for viz.

sidneymbell:
@neuromusic - That's one source of genes being filtered out. But, they're also often filtered / removed during normalization. That's what makes it tricky: you may want to color by a gene, but you didn't normalize those expression values and thus can't visualize them on the same scale.

neuromusic:
I think I see where I was mixed up.

The new `highly_variable_genes` preprocessing function will leave them, but other preprocessing functions like `filter_genes` (which is used by prepare if there's no recipe) will still drop them.

csweaver:
Yeah, we never use the `highly_variable_genes` method.

jk:
> @jk - No. Please see #689.

I'm most probably not the one you want to mention.

jk:
> @jk - No. Please see #689.

@csweaver I'm most probably not the one you want to mention.

bkmartinjr:
@neuromusic - see two accidental mentions of "at" jk  above.

bkmartinjr:
@csweaver verbally confirmed that our CI pipeline tests this against the latest version of ScanPy.

PR #690, opened 2019-04-02
Add `calculate_qc_metrics` to `prepare`

### LINKED ISSUE #100: Describe the shape of the data overall
Some datasets come in with pre-calculated values for things like the number of genes each cell is expressing, the proportion of cells expressing each gene, read counts per cell, etc. It would be _super helpful_ to have these auto-computed and displayed for every dataset.

sidneymbell:
Adds [`calculate_qc_metrics`](https://scanpy.readthedocs.io/en/stable/api/scanpy.pp.calculate_qc_metrics.html#scanpy.pp.calculate_qc_metrics), implemented in scanpy 1.3.4, to `prepare`. Fixes #100.

I added this directly after data loading as these should be calculated upstream of any normalization. But, open to suggestion if you think they make more sense to add elsewhere. 

These show up as continuous metadata. 
image


server/cli/prepare.py
```diff
@@ -116,6 +120,9 @@ def load_data(data):
         if not adata._var.index.is_unique:
             click.echo("Warning: var index is not unique")
 
+        if calculate_qc_metrics:
```
csweaver: If this takes a non-trivial amount of time (test on the 1 mil dataset) move this to a function and add to the show_step dict. Otherwise, it's fine as is.

server/cli/prepare.py
```diff
@@ -185,7 +192,7 @@ def show_step(item):
     steps = [make_sparse, run_recipe, run_pca, run_neighbors, run_louvain, run_layout]
 
     click.echo(f"[cellxgene] Loading data from {data}, please wait...")
-    adata = load_data(data)
+    adata = load_data(data, calculate_qc_metrics)
```
csweaver: see above

PR #697, opened 2019-04-08
Add `calculate_qc_metrics` to `prepare

### LINKED ISSUE #100: Describe the shape of the data overall

sidneymbell:
Adds calculate_qc_metrics, implemented in scanpy 1.3.4, to prepare. Fixes #100.

These show up as continuous metadata.

Duplicate of [PR 690](https://github.com/chanzuckerberg/cellxgene/pull/690). [TIL] Apparently github closes PRs and doesn't allow editing of the base branch designation if the original base branch is merged and deleted -_-


server/cli/prepare.py
```diff
@@ -172,6 +182,7 @@ def run_layout(adata):
 
     def show_step(item):
         names = {
+            "calculate_qc_metrics": "Calculating QC metrics" if run_qc else "Skipping QC",
```
csweaver: I'm not a fan of inline if else's in python code. The exceptions being in list comprehension and lambda functions. Could you move this to an if statement above names?

server/requirements.txt
```diff
@@ -9,7 +8,7 @@ flatbuffers>=1.10.0
 matplotlib>=2.2
 numpy>=1.15.2
 pandas>=0.23.1
-scanpy>=1.3.2
+scanpy>=1.3.7
```
csweaver: Why did this change? Shouldn't this be in master?

csweaver: I think you may need to rebase this branch onto master and then force push. I can help with that if you want.

sidneymbell: Cool. We can do that, but force push terrifies me. Can I get your help on this so I don't accidentally break the whole world? ^_^

PR #711, opened 2019-04-12
Core library

csweaver:
This PR adds some features to the core library to aid in development of the notebook and native app development.

You can:
1. Import from app.py without launching the flask app
2. Initialize engine without loading data
3. Reload data and update config (layout method, etc)

To implement 2&3 I've moved the engine arguments from individual properties to a single config prop,  added an update method which will update the data/config and if there is data recompute the validation and initialization. I'm handling the case when methods in the engine require data with a decorator that raises an error if the data isn't set, but i'm not sure if that was the best solution - open to suggestions.

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -25,7 +25,6 @@ def create_app():
     SECRET_KEY = os.environ.get("CXG_SECRET_KEY", default="SparkleAndShine")
     app.config.update(SECRET_KEY=SECRET_KEY)
 
-
     resources = get_api_resources()
     app.register_blueprint(webapp.bp)
     app.register_blueprint(resources.blueprint)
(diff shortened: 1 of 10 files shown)


server/app/app.py
```diff
@@ -5,27 +5,28 @@
...
+from server.app.web import webapp
 
-REACTIVE_LIMIT = 1_000_000
-
-app = Flask(__name__, static_folder="web/static")
-app.json_encoder = Float32JSONEncoder
-cache = Cache(app, config={"CACHE_TYPE": "simple", "CACHE_DEFAULT_TIMEOUT": 860_000})
-Compress(app)
-CORS(app)
+# Application Data
```
bkmartinjr: In some scenarios, I can see us wanting to load and serve multiple data sets in one process (eg, using the future API in a notebook).  Have you considered making the "app" a class, and store data and cache within the instance?

csweaver: 👍 Done

Issue #687, opened 2019-04-02
Add native user experience for install and launch


csweaver:
Goal: ease the barrier of use for cellxgene by adding a OS-native app as a new usage mode. Targeting support for Windows and MacOS. (Linux will be still be supported with the CLI mode)

Motivation: the CLI mode in cellxgene is a significant barrier of entry for non-computational users (ex. wet-lab experimentalists, medical researchers). 

User story example: Erin an experimental biologist is in collaboration with Cary a computational biologist. Cary has prepared the dataset into the format required by cellxgene. Cary sends Erin a link to the dataset as well as the installer for cellxgene. Erin is able to download cellxgene and install and launch cellxgene without having to have experience with python or the command line. Then Erin can use the native file browser to find the dataset to load and explore the data. 

- [x] Installer for cellxgene and all of its Python dependencies (ie, excluding Chrome)
- [ ] Native launch ui (applications folder, start menu entry, etc.) 
- [x] Native load and save (file dialogue box, drag and drop, etc.)

Current usage modes will still be supported (pip installable CLI, Docker container)

ttung:
Possibly detect new versions? :)

neuromusic:
@ttung yes! #683

csweaver:
- [ ] ensure a user can `import cellxgene` 

(Make sure the namespace is set to cellxgene)

csweaver:
- [ ] Better naming of "server" related issues in general

See Bruce's comment from #730 
> I find the naming confusing as there are several "servers" all doing different things (and the term is super generic). `Server` is really a `RESTDataServer` (or something similar), as it serves up the r/o data to the front-end. Maybe make the names more descriptive.

PR #730, opened 2019-04-26
GUI app (experimental)

csweaver:
Addresses some of #687.

This PR adds an **unsupported, prototype**, "proof of concept" GUI app to cellxgene. It's written using pyqt5 for the GUI and cefpython for the browser embedded in the app. Tested on OSX, but needs testing on windows. The file load and server are threaded using qt's qrunnable and threadpool. The load screen and browser are implemented as a stacked layout and the browser automatically shows when you load a file. To get back to the browser, you use the file menu in the native menu bar. 

To launch app from command line:
`python server/gui/main.py`

*Initial Load Screen*
Initial Load Screen

*Select File*
Select File

*GUI App*
App Loaded

I'd like to get this code in and reviewed while I work on a couple of the other things I think the MVP still needs. 

TODO (GUI App)
1. Move to python thread from qt threadpool - data load is blocking the app despite being in a thread, I'm going to use python's threads instead and hope that fixes it.
2. Drag and drop files
3. UI to cancel load screen and go back to browser (if loading data after initial load)


bkmartinjr:
I would also consider for the MVP to-do:  error handling (need to flesh out what this means).

server/requirements.txt
```diff
@@ -1,3 +1,4 @@
+cefpython3>=66
```
bkmartinjr: is there a way to make these requirements optional (eg, put in extras)?   Does not seem appropriate to make them hard requirements given the experimental/prototype nature of this PR.

server/gui/main.py
```diff
@@ -0,0 +1,212 @@
...
+from cefpython3 import cefpython as cef
+from PyQt5.QtCore import *
+from PyQt5.QtWidgets import *
+
+from server.gui.browser import CefWidget, CefApplication
+from server.gui.cxg_server import cellxgeneServer, DataLoadWorker, ServerRunWorker
+from server.gui.utils import WINDOWS, LINUX, MAC
+
+# Configuration
+# TODO remember this or calculate it?
```
bkmartinjr: Suggest: start an issue to capture these UI rqmts in an epic.

server/gui/main.py
```diff
@@ -0,0 +1,212 @@
...
+        self.server.setup_app()
+        self.runServer()
+        self.setWindowTitle("cellxgene")
+
+        # Strong focus - accepts focus by tab & click
+        self.setFocusPolicy(Qt.StrongFocus)
+        self.setupLayout()
+        self.setupMenu()
+
+    def setupLayout(self):
```
bkmartinjr: Python style question: I thought method names should follow function naming conventions, ie, snake case?   Reference [PEP 8 Method Names and Instance Variables](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables).   I'm fine if you have a reason for deviating - mostly trying to learn.

I see that PyQt uses this convention, so my guess is that the practice is to conform to that?

server/gui/main.py
```diff
@@ -0,0 +1,212 @@
...
+            # before CEF browser is embedded, otherwise window is
+            # not resized and application hangs during resize.
+            self.show()
+
+        # Browser can be embedded only after layout was set up
+        self.cef_widget.embedBrowser()
+
+        if LINUX:
+            # On Linux with PyQt5 the QX11EmbedContainer widget is
+            # no more available. An equivalent in Qt5 is to create
```
bkmartinjr: "no more available"?  maybe "no longer available" or "not available"?

server/gui/cxg_server.py
```diff
@@ -0,0 +1,66 @@
...
+
+class cellxgeneServer():
+    def __init__(self, parent, host="127.0.0.1", port=8000):
+        self.app = None
+        self.host = host
+        self.port = port
+
+    def setup_app(self):
+        from server.app.app import Server
+        server = Server()
```
bkmartinjr: I find the naming confusing as there are several "servers" all doing different things (and the term is super generic).    `Server` is really a `RESTDataServer` (or something similar), as it serves up the r/o data to the front-end.   Maybe make the names more descriptive.

Also, in this context, do we need a separate `cellxgeneServer` class, or can this functionality be incorporated into `server.app.app`?

server/gui/cxg_server.py
```diff
@@ -0,0 +1,66 @@
...
+class ServerRunWorker(QRunnable):
+    def __init__(self, app, host, port, *args, **kwargs):
+        super(ServerRunWorker, self).__init__()
+        self.app = app
+        self.host = host
+        self.port = port
+
+    @pyqtSlot()
+    def run(self):
+        self.app.run(host=self.host, debug=False, port=self.port, threaded=True)
```
bkmartinjr: if port and host are passed directly to the worker, why does the cellxgeneServer class have instance variables for them?

server/gui/main.py
```diff
@@ -0,0 +1,212 @@
...
+        worker = ServerRunWorker(self.server.app, host=self.server.host, port=self.server.port)
+        self.thread_pool.start(worker)
+
+    def clearBrowserReferences(self):
+        # Clear browser references that you keep anywhere in your
+        # code. All references must be cleared for CEF to shutdown cleanly.
+        self.cef_widget.browser = None
+
+
+# TODO make central location for methods?
```
bkmartinjr: thumbs up on the generalization - don't think these kinds of general config belong in the Qt app.  We will want them exposed to other wrappers as well (notebook, Qt, CLI, ...).

server/requirements.txt
csweaver: yes, that was an oversight on my part. Will absolutely fix.

server/gui/main.py
csweaver: 👍

server/gui/main.py
csweaver: LOL! This was directly taken from the cefpython demo, I guess I forgot to copy edit. I'll fix it.

server/gui/cxg_server.py
csweaver: I agree, I'll rename.

Hmm, I'll think on integrating the functionality into server.app.app. That's a good idea.

server/gui/cxg_server.py
csweaver: incomplete refactoring. will fix

server/gui/main.py
csweaver: should I do this at another point? or get it in this PR?

server/gui/main.py
csweaver: PyQt is copying C++ style I believe, and there are a few methods in the classes I'm building that require that. In the interest of consistency if a class had at least one method in camel case, then I made all methods have camel case. Otherwise I used snake case. (see "A Foolish Consistency is the Hobgoblin of Little Minds" section of the PEP 8 link) I'm open to alternatives, I'm not sure I made the best choice for when to apply the rules and when not to.

csweaver:
> I would also consider for the MVP to-do: error handling (need to flesh out what this means).

I'm actually doing a little error handling when loading a bad file, I'll think about how to handle other errors in the app.

server/gui/main.py
bkmartinjr: OK with me if you do it progressively (ie, not in this PR)

server/gui/cxg_server.py
csweaver: Actually it's not just a REST data server, it's also serving the main html page and the assets. Can we postpone this naming issue to add to the `import cellxgene` ticket?

server/gui/main.py
csweaver: just did it, it was easy

server/gui/cxg_server.py
csweaver: I did actually move the functionality to server.app.app (which rippled through to server.cli.launch.py) but I think it makes it better code in general.

bkmartinjr: yes, OK delaying on naming issue

csweaver: changed cxg_server.py -> workers.py to reflect the deletion and changed role of the file

server/gui/main.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class LoadWidget(QFrame):
+    def __init__(self, parent):
+        super(LoadWidget, self).__init__(parent=parent)
+
+        # Init layout
+        self.MAX_CONTENT_WIDTH = 500
+        load_ui_layout = QVBoxLayout()
+        h_margin = int((WIDTH - self.MAX_CONTENT_WIDTH) / 2)
```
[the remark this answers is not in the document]
csweaver: thanks, good catch!

server/gui/workers.py
```diff
@@ -0,0 +1,50 @@
...
+        self.layout = layout
+        self.signals = WorkerSignals()
+
+    @pyqtSlot()
+    def run(self):
+        if not self.data_file:
+            self.signals.finished.emit()
+            return
+
+        from server.app.scanpy_engine.scanpy_engine import ScanpyEngine
```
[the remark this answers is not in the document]
csweaver: Actually to speed load time. I'll clarify.

server/gui/utils.py
```diff
@@ -0,0 +1,24 @@
...
+LINUX = (platform.system() == "Linux")
+MAC = (platform.system() == "Darwin")
+
+
+class WorkerSignals(QObject):
+    '''
+    Defines the signals available from a running worker thread.
+    Supported signals are:
+    finished
+    No data
```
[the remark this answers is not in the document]
csweaver: will do

Issue #735, opened 2019-04-29
Use pyside2 instead of pyqt5


csweaver:
Due to license issues we should avoid pyqt5 if possible (GPL). PySide2 is a good alternative (LGPL), must be done before we create the installer.

PR #738, opened 2019-04-30
PyQt5 -> PySide2

csweaver:
Switched qt wrapper libraries. PySide2 uses LGPL as opposed to PyQt5 which uses GPL. LGPL is more compatible with our MIT license and the API's were similar enough that the change was minimal.


csweaver:
fixes #735

PR #740, opened 2019-05-01
avoid overflow on clip dialogue

colinmegill:


client/src/components/graph/graph.js
```diff
@@ -362,9 +362,9 @@ class Graph extends React.Component {
   };
 
   handleClipOnKeyPress = e => {
-    /*
```
csweaver: prettier?

colinmegill: Yes, got a hold of some trailing spaces

PR #747, opened 2019-05-05
lodash cleanup

bkmartinjr:
Progressive cleanup of lodash use where we no longer need it.  This includes three changes:
* removed all legacy use of _.get, in favor of optioning chaining and nullish coallescing
* removed a few cases of _.forEach in favor of native ES6 functions
* consolidated all of our range and range-fill implementations into a single implementation, wrote test cases, and removed use of _.range.
* take advantage of previously added schema indexing by name (`schema.obsByName`), allowing redundant filtering of schema values for categories to be removed.

### DIFF
--- a/client/src/components/categorical/occupancy.js
+++ b/client/src/components/categorical/occupancy.js
@@ -1,6 +1,5 @@
 // jshint esversion: 6
 import React from "react";
-import _ from "lodash";
 import { connect } from "react-redux";
 import * as d3 from "d3";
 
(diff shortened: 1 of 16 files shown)


client/src/components/categorical/occupancy.js
```diff
@@ -17,9 +16,7 @@ class Occupancy extends React.Component {
     const width = 100;
     const height = 11;
 
-    const categories = _.filter(schema.annotations.obs, {
-      name: colorAccessor
-    })[0].categories;
+    const categories = schema.annotations.obsByName[colorAccessor]?.categories;
```
colinmegill: _.filter doesn't seem covered in the description of the PR, nor does `obsByName` - could you clarify here or there?

bkmartinjr: done!

`obsByName` feature was added some time ago, so we didn't have to keep reprocessing the schema for cases like this.   So this was an example of unnecessary compute being done because I did not previously finish refactoring.

Issue #751, opened 2019-05-06
Add test to ensure that all packages contain an __init__.py file


csweaver:
An `__init__.py` file is required for python packages. When installing cellxgene for server development a missing `__init__.py` file does not cause errors. The error is evident when it is release to pypi. 

The `tox` package may help with this.

PR #760, opened 2019-05-09
install from dist/ instead of build/ on travis

csweaver:
On travis, build a distribution and install from there, instead of installing from `build/`. This will surface missing `__init__.py` files and lessen the burden of testing releases.


csweaver:
Fixes #751

PR #761, opened 2019-05-09
port retry

### LINKED ISSUE #698: better handling of port in use
If the 5005 port is in use, cellxgene throws an error which doesn't make it clear to users where the problem is...

```
(cellxgene) hal-9000:cellxgene-explore justinkiggins$ cellxgene launch -l umap osmFISH_SScortex_mouse_all_cells.h5ad
[cellxgene] Loading data from osmFISH_SScortex_mouse_all_cells.h5ad, this may take awhile...
[cellxgene] Warning: Annotation X will be converted to 32 bit float and may lose precision. 
[cellxgene] Warning: Annotation Y will be converted to 32 bit float and may lose precision. 
[cellxgene] Warning: Annotation _tSNE_1 will be converted to 32 bit float and may lose precision. 
[cellxgene] Warning: Annotation _tSNE_2 will be converted to 32 bit float and may lose precision. 
[cellxgene] Warning: Annotation size_pix will be converted to 32 bit float and may lose precision. 
[cellxgene] Warning: Annotation size_um2 will be converted to 32 bit float and may lose precision. 
[cellxgene] Launching! Please go to http://127.0.0.1:5005 in your browser.
[cellxgene] Type CTRL-C at any time to exit.
*OSError: [Errno 48] Address already in use*
```

Possible fixes:

### 1) improve error handling

handle error and tell the user to try another port

### 2) try other ports

automatically increment the port until an available port is found. jupyter implements this nicely, with a `--port-retries` argument to cap the number of attempts:

```
--port-retries=<Int> (NotebookApp.port_retries)
    Default: 50
    The number of additional ports to try if the specified port is not
    available.
```

csweaver:
Fixes #698 

If port is in use tries incrementing the specified port up to 50 times to find an available one. Adds `--fixed-port` option to enforce using only the specified port. Also gives a more informative error message if the port launched is in use. 

Reviewers:
@bkmartinjr  (required)
@ttung  (optional, but would appreciate your advice)
@neuromusic (optional) - you submitted the feature request, does this meet your use case?

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -1,3 +1,4 @@
+import errno
 import logging
 from os import devnull
 from os.path import splitext, basename
@@ -190,4 +191,6 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
     try:
         server.app.run(host=host, debug=debug, port=port, threaded=True)
     except OSError as e:
-        raise click.ClickException(f"{e}. Port is in use, please specify another port using the --port flag.")
+        if e.errno == errno.EADDRINUSE:
+            raise click.ClickException(f"{e} Port is in use, please specify an open port using the --port flag.")
+        raise click.ClickException(f"{e}")
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -51,12 +51,10 @@ from server.utils.utils import find_available_port
     show_default=True,
     help="Open the web browser after launch.",
 )
-@click.option("--port", "-p", help="Port to run server on.", metavar="", default=5005, show_default=True)
+@click.option("--port", "-p", help="Port to run server on.", metavar="", show_default=True)
 @click.option("--obs-names", default=None, metavar="", help="Name of annotation field to use for observations.")
 @click.option("--var-names", default=None, metavar="", help="Name of annotation to use for variables.")
 @click.option("--host", default="127.0.0.1", help="Host IP address")
-@click.option("--no-port-scan", is_flag=True, default=False, show_default=True,
-              help="Disable automatic scan for unused port.")
 @click.option(
     "--max-category-items",
     default=100,
@@ -89,7 +87,6 @@ def launch(
         open_browser,
         port,
         host,
-        no_port_scan,
         max_category_items,
         diffexp_lfc_cutoff,
         scripts,
@@ -139,16 +136,8 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
         file_parts = splitext(basename(data))
         title = file_parts[0]
 
-    print(no_port_scan)
-    if not no_port_scan:
-        new_port = find_available_port(host, port)
-        # warn if the user specified port is being overridden
-        if new_port != port and port != 5005:
-            click.echo(
-                f"[cellxgene] Warning: the port you specified was in use, using port {new_port} instead. "
-                f"If you want to require cellxgene to only use the port specified and exit if that port is not "
-                f"available run launch with the --fixed-port flag")
-        port = new_port
+    if not port:
+        port = find_available_port(host)
 
     # Setup app
     cellxgene_url = f"http://{host}:{port}"
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -139,6 +139,7 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
         file_parts = splitext(basename(data))
         title = file_parts[0]
 
+    print(no_port_scan)
     if not no_port_scan:
         new_port = find_available_port(host, port)
         # warn if the user specified port is being overridden
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -194,4 +194,4 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
     except OSError as e:
         if e.errno == errno.EADDRINUSE:
             raise click.ClickException("Port is in use, please specify an open port using the --port flag.") from e
-        raise e
+        raise
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -11,6 +11,7 @@ from server.app.app import Server
 from server.app.util.errors import ScanpyFileError
 from server.app.util.utils import custom_format_warning
 from server.utils.constants import MODES
+from server.utils.utils import find_available_port
 
 
 @click.command()
@@ -199,36 +200,3 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
         server.app.run(host=host, debug=debug, port=port, threaded=True)
     except OSError as e:
         raise click.ClickException(f"{e}. Port is in use, please specify another port using the --port flag.")
-
-
-def find_available_port(host, port):
-    """
-    Helper method to find open port on host. Tries 50 ports incremented from the specified port
-    """
-    import errno
-    import socket
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    new_port = port
-    found_port = False
-    num_ports_to_try = 50
-    for i in range(num_ports_to_try):
-        new_port = port + i
-        while True:
-            try:
-                s.bind((host, new_port))
-                found_port = True
-                break
-            except socket.error as e:
-                # Reraise if error is not "address in use"
-                if e.errno != errno.EADDRINUSE:
-                    raise e
-                break
-        if found_port:
-            break
-    if not found_port:
-        click.echo(f"[cellxgene] No port in range {new_port} - {new_port + num_ports_to_try - 1} was"
-                   f" available to run cellxgene server on.")
-        click.echo(f"[cellxgene] Exiting")
-        raise OSError("No available ports")
-    s.close()
-    return new_port
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -203,32 +203,32 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
 
 def find_available_port(host, port):
     """
-    Helper method to find open port on host. Tries 50 ports incremented from specified port
+    Helper method to find open port on host. Tries 50 ports incremented from the specified port
     """
     import errno
     import socket
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     new_port = port
-    for i in range(3):
+    found_port = False
+    num_ports_to_try = 50
+    for i in range(num_ports_to_try):
         new_port = port + i
-        print(new_port)
         while True:
             try:
                 s.bind((host, new_port))
-                print("worked")
+                found_port = True
                 break
             except socket.error as e:
-                if e.errno == errno.EADDRINUSE:
-                    continue
-                else:
-                    raise
-            finally:
-                s.close()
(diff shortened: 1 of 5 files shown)


server/utils/utils.py
```diff
@@ -0,0 +1,34 @@
+import click
+
+
+def find_available_port(host, port):
+    """
+    Helper method to find open port on host. Tries 50 ports incremented from the specified port
+    """
+    import errno
```
ttung: curious why you import here.

server/utils/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    import socket
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
ttung: this is the snippet of code I may have written for hca/data-store to find an unused tcp port:

```
    import contextlib
    import socket

    with contextlib.closing(socket.socket()) as sock:
        sock.bind(('127.0.0.1', 0))
        return sock.getsockname()[1]
```

it closes the port cleanly with minimal code.

server/utils/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    new_port = port
+    found_port = False
+    num_ports_to_try = 50
+    for i in range(num_ports_to_try):
```
ttung:
```suggestion
    for port_to_try in range(port, port + num_ports_to_try):
```

server/utils/utils.py
csweaver: good advice, will do

server/utils/utils.py
csweaver: The IDE's refactor artifact, will fix

server/cli/launch.py
```diff
@@ -135,6 +139,15 @@ def launch(
         file_parts = splitext(basename(data))
         title = file_parts[0]
 
+    if not fixed_port:
```
bkmartinjr: if I am reading the code right, this will generate a warning in the case where the user has use defaults, and the default port is unavailable.

In this particular case,  where the user has let the system default, I'm not sure the warning adds any value or is all that friendly.  Consider suppressing it in this case...

server/cli/launch.py
```diff
@@ -54,6 +55,8 @@
 @click.option("--obs-names", default=None, metavar="", help="Name of annotation field to use for observations.")
 @click.option("--var-names", default=None, metavar="", help="Name of annotation to use for variables.")
 @click.option("--host", default="127.0.0.1", help="Host IP address")
+@click.option("--fixed-port", is_flag=True, default=False, show_default=True,
+              help="Error if request port is not available instead of searching for another available port.")
```
bkmartinjr: I'm not sure the help message is clear.  Ideas:
* rename option to `-no-port-scan`
* help="Disable automatic scan for unused port"

neuromusic: +1

ttung: For what it's worth, I think it might be easier to just have `--port`, and if it's not specified, just go and find a port.

server/utils/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    new_port = port
+    found_port = False
+    num_ports_to_try = 50
+    with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        for port_to_try in range(port, port + num_ports_to_try):
+            while True:
+                try:
+                    s.bind((host, port_to_try))
+                    found_port = True
```
ttung: at this point, I believe you can just `return port_to_try`.

you can remove `found_port`, and the associated logic with that.

server/utils/utils.py
```diff
@@ -0,0 +1,32 @@
...
+import socket
+
+def find_available_port(host, port):
+    """
+    Helper method to find open port on host. Tries 50 ports incremented from the specified port
+    """
+    new_port = port
+    found_port = False
+    num_ports_to_try = 50
+    with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
```
ttung: I do not know if you can use the same socket and attempt to bind it multiple times.

server/cli/launch.py
csweaver: yes, much better. naming things is hard 😞

server/utils/utils.py
csweaver: yes, yes I can. 👍

server/cli/launch.py
csweaver: done

csweaver:
FYI I don't think the client/package.json security issue is real.

server/cli/launch.py
```diff
@@ -135,6 +139,17 @@ def launch(
         file_parts = splitext(basename(data))
         title = file_parts[0]
 
+    print(no_port_scan)
```
ttung: remove pls.

server/cli/launch.py
```diff
@@ -54,6 +55,8 @@
 @click.option("--obs-names", default=None, metavar="", help="Name of annotation field to use for observations.")
 @click.option("--var-names", default=None, metavar="", help="Name of annotation to use for variables.")
 @click.option("--host", default="127.0.0.1", help="Host IP address")
+@click.option("--no-port-scan", is_flag=True, default=False, show_default=True,
```
ttung: I personally find this more confusing than just making `--port` optional.  Also, port scans have a funny connotation. :)

server/cli/launch.py
```diff
@@ -135,6 +139,17 @@ def launch(
...
 
+    print(no_port_scan)
+    if not no_port_scan:
+        new_port = find_available_port(host, port)
+        # warn if the user specified port is being overridden
+        if new_port != port and port != 5005:
+            click.echo(
+                f"[cellxgene] Warning: the port you specified was in use, using port {new_port} instead. "
+                f"If you want to require cellxgene to only use the port specified and exit if that port is not "
+                f"available run launch with the --fixed-port flag")
```
ttung: this help text is no longer relevant.

server/cli/launch.py
```diff
@@ -183,4 +198,7 @@ def launch(
         f = open(devnull, "w")
         sys.stdout = f
 
-    server.app.run(host=host, debug=debug, port=port, threaded=True)
+    try:
+        server.app.run(host=host, debug=debug, port=port, threaded=True)
+    except OSError as e:
+        raise click.ClickException(f"{e}. Port is in use, please specify another port using the --port flag.")
```
ttung: this is not consistent with the actual code.

csweaver:
@ttung I implemented your port scan on no --port flag suggestion instead of what I had before. Great idea.

server/utils/utils.py
```diff
@@ -0,0 +1,18 @@
+import contextlib
+import socket
+
+
+def find_available_port(host, port=5005):
+    """
+    Helper method to find open port on host. Tries 50 ports incremented from the specified port
```
bkmartinjr: comment does not match code s/50/5000/

server/cli/launch.py
```diff
@@ -183,4 +187,7 @@ def launch(
         f = open(devnull, "w")
         sys.stdout = f
 
-    server.app.run(host=host, debug=debug, port=port, threaded=True)
+    try:
+        server.app.run(host=host, debug=debug, port=port, threaded=True)
+    except OSError as e:
```
bkmartinjr: there are other reasons you may get an OSError throw. This catch makes them all seem like port selection errors.

server/utils/utils.py
csweaver: fixed

server/cli/launch.py
csweaver: I added an error number check to fix this.

server/cli/launch.py
```diff
@@ -183,4 +188,9 @@ def launch(
...
+        if e.errno == errno.EADDRINUSE:
+            raise click.ClickException(f"{e} Port is in use, please specify an open port using the --port flag.")
```
ttung: Personally, I prefer chained exceptions.

```suggestion
            raise click.ClickException(f"Port is in use, please specify an open port using the --port flag.") from e
```

server/cli/launch.py
```diff
@@ -183,4 +188,9 @@ def launch(
...
+        raise click.ClickException(f"{e}")
```
ttung: I'm not sure why you are raising click.ClickException rather than just reraising the original exception (just `raise` will do that).

server/cli/launch.py
```diff
@@ -50,7 +52,7 @@
     show_default=True,
     help="Open the web browser after launch.",
 )
-@click.option("--port", "-p", help="Port to run server on.", metavar="", default=5005, show_default=True)
+@click.option("--port", "-p", help="Port to run server on.", metavar="", show_default=True)
```
ttung: Also, consider making the help text indicate that the app will attempt to find an open port if this is not specified.

server/cli/launch.py
csweaver: fixed

server/cli/launch.py
csweaver: good catch, this isn't an expected error so I don't want click to handle it. will fix

server/cli/launch.py
csweaver: will fix

server/cli/launch.py
```diff
@@ -183,4 +189,9 @@ def launch(
         f = open(devnull, "w")
         sys.stdout = f
 
-    server.app.run(host=host, debug=debug, port=port, threaded=True)
+    try:
+        server.app.run(host=host, debug=debug, port=port, threaded=True)
+    except OSError as e:
+        if e.errno == errno.EADDRINUSE:
+            raise click.ClickException("Port is in use, please specify an open port using the --port flag.") from e
+        raise e
```
ttung: I would just do `raise`

I suspect you want the original stacktrace, and I'm not certain this will produce the original stacktrace.

(i tried it out with a toy example, and it does modify the stacktrace)

server/utils/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    # Takes approx 2 seconds to do a scan of 5000 ports on my laptop
+    num_ports_to_try = 5000
+    for port_to_try in range(port, port + num_ports_to_try):
+        with contextlib.closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+            try:
+                s.bind((host, port_to_try))
+                return port_to_try
+            except socket.error:
+                pass
+    raise OSError(errno.EADDRINUSE, f"No port in range {port} - {port + num_ports_to_try - 1} available.")
```
ttung: I would just make this raise socket.error(..) since that's really what it is.  You'd have to modify server/cli/launch.py:194.

If you must have a more generic exception class, try `IOError`

server/cli/launch.py
csweaver: Changed. I didn't notice any issue with my example test, but I didn't check that the stack trace was identical.

server/utils/utils.py
csweaver: done

PR #763, opened 2019-05-10
large file size guardrails

### LINKED ISSUE #723: add guardrails to launch for non-CSC sparse datasets
for performance when querying genes, cellxgene needs column-optimized expression matrix. Practically, this means that `.X` needs to be either a dense format or CSC-structured sparse array.

If the user tries to launch with another format (for example, CSR, DIA, DOK), they will hit performance issues like in #718 and #719 

We should add some guardrails to prevent users from doing this.

On launch, check the format of `.X`. If it is neither dense or CSC, then we should prevent cellxgene from launching, tell the user why, and give them a way to launch nonetheless. I'm open to either `click.prompt`/`click.confirm` or a command line flag to "force" the file to load.

### LINKED ISSUE #739: add CLI guardrails for large files
Currently cellxgene CLI prints a warning when started, letting the user know that it may "take a while" to load the data.

Two issues with this:
* it is printed indiscriminately, even when the file is small
* it fails to detect cases where load may succeed, but the file is larger than main memory and therefore may perform badly post-load (eg, diffexp on a dataset that exceeds main memory is going to be very, very slow).

Two proposed changes for discussion:
- [ ] Print the loading "take a while" warning only when the file size exceeds some threshold (say 100MB).
- [ ] If the file size exceeds total host RAM, print a warning that interactive performance may be slow (if we can't calculate RAM, which may be tricky on some platforms, default to something reasonable, eg, 16GB).

bkmartinjr:
This PR makes several changes to the CLI messages for input h5ad files:
* If a data file is loaded, containing a sparse matrix in a format other than CSC, print a warning message informing the user that using CSC may improve performance
* The initial data loading message will only print "this may take a while" when the file exceeds 100MB
* If the input data file size exceeds main memory size, the user is informed that performance may suffer (caveat: this will have false negatives if the file is compressed)

In addition, because cellxgene requires an h5ad file, this PR switches from `ScanPy.read()` to `anndata.read_h5ad()`.  The latter module imports significantly faster, removing several seconds from startup time.

Fixes #723 
Fixes #739

### DIFF
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -305,7 +305,8 @@ class ScanpyEngine(CXGDriver):
         """
         if var_mask is None:    # noop
             return X
-        if sparse.issparse(X):  # use tuned getcol/hstack for performance
+        if sparse.isspmatrix_csc(X):
+            # use tuned getcol/hstack for performance
             indices = np.nonzero(var_mask)[0]
             cols = [X.getcol(i) for i in indices]
             return sparse.hstack(cols, format="csc")
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -305,8 +305,7 @@ class ScanpyEngine(CXGDriver):
         """
         if var_mask is None:    # noop
             return X
-        if sparse.isspmatrix_csc(X):
-            # use tuned getcol/hstack for performance
+        if sparse.issparse(X):  # use tuned getcol/hstack for performance
             indices = np.nonzero(var_mask)[0]
             cols = [X.getcol(i) for i in indices]
             return sparse.hstack(cols, format="csc")
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -1,5 +1,5 @@
 import warnings
-import os
+
 import numpy as np
 from pandas.core.dtypes.dtypes import CategoricalDtype
 import anndata
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -1,8 +1,8 @@
 import warnings
-
+import os
 import numpy as np
 from pandas.core.dtypes.dtypes import CategoricalDtype
-import scanpy as sc
+import anndata
 from scipy import sparse
 
 from server.app.driver.driver import CXGDriver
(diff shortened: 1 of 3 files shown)


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
 from pandas.core.dtypes.dtypes import CategoricalDtype
-import scanpy as sc
+import anndata
```
sidneymbell: Why the switch from scanpy to anndata? I think it's fine (it actually makes it more clear that we're just using the data structure), just wondering what the thought process was?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -174,6 +173,11 @@ def _validate_and_initialize(self):
 
     @requires_data
     def _validate_data_types(self):
+        if sparse.isspmatrix(self.data.X) and not sparse.isspmatrix_csc(self.data.X):
+            warnings.warn(
+                f"Scanpy data matrix is sparse, but not a CSC (columnar) matrix.  "
+                f"Performance may be improved by using CSC."
```
sidneymbell: Can we point them to a reference for how to convert between the two? I suspect some biologists (like me, up until recently :) may not be super familiar with the nuances.

sidneymbell:
Made a couple small comments, none are blocking merge. Other than the failed CI check, this LGTM!

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: it imports in 1/2 of the time - so we get faster startup.

bkmartinjr: plus, we aren't importing a bunch of stuff into memory that is unused.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: that would be something for docs/FAQ, right?   What would you propose beyond the current advice in our FAQ, which points people at using `--sparse`?

server/cli/launch.py
```diff
@@ -13,6 +14,10 @@
 from server.utils.constants import MODES
 
 
+# anything bigger than this will generate a special message
+BIG_FILE_SIZE_THRESHOLD = 100 * 2**20
```
csweaver: Just so I don't have to do any real thinking...

```suggestion
BIG_FILE_SIZE_THRESHOLD = 100 * 2**20 #100MB
```

server/cli/launch.py
```diff
@@ -148,7 +153,16 @@ def launch(
         log = logging.getLogger("werkzeug")
         log.setLevel(logging.ERROR)
 
-    click.echo(f"[cellxgene] Loading data from {basename(data)}, this may take awhile...")
+    file_size = getsize(data) if exists(data) else 0
```
csweaver: Do we need this exists check since click already checks that for us? (Line 19) I'm ok with leaving it in though.

server/cli/launch.py
```diff
@@ -1,9 +1,10 @@
 import logging
 from os import devnull
-from os.path import splitext, basename
+from os.path import splitext, basename, exists, getsize
 import sys
 import warnings
 import webbrowser
+import psutil
```
csweaver: This is a third party import, so we should move to below click

server/cli/launch.py
bkmartinjr: will fix

server/cli/launch.py
bkmartinjr: I suppose not.  I'm just paranoid.  Will remove.

server/cli/launch.py
bkmartinjr: will do

Issue #725, opened 2019-04-22
support loading arbitrary layouts


neuromusic:
currently, `cellxgene launch` allows the user to select from the standard layouts that scanpy uses, each of which is stored in `.obsm` and prefaced with `X_`.

If a user has another layout (e.g. spatial), they cannot load it into cellxgene.

suggested change: allow the user to use any layout stored in an `.obsm` key with the syntax `X_<layout>`

preferred UX:

1. let user pass layout to use into CLI
2. if requested layout does not exist or no layout is passed in, inspect file and give user prompt with available layouts

sidneymbell:
Awesome! We also need to validate that the new layout is valid (e.g., no `NaN`s) and slice it to 2 dimensions.

bkmartinjr:
Based upon further conversation with @neuromusic and @sidneymbell, the UX proposal is now:

1. CLI will accept an zero or more `--layout` arguments, e,g, `--layout X_tsne --layout X_umap`.   
2. Any layouts specified from the CLI will be presented to the user via drop-down menu (see #594)
3. If no `--layout` CLI argument is specified, any field in `.obsm` with the name `X_*` will be treated as a layout for presentation to the user.

All candidate layouts (`.obsm[foo]`) data must conform to the following or it will be ignored:
- data is an ndarray of shape (n_obs, >=2) and dtype float or int
- data contains only finite values (if float)

As with the current implementation, all layout data will be coerced to float32 for transmission to the front-end.

sidneymbell:
This sounds good to me! Two small additions:

* If we are loading all `.obsm[X_*]` layouts, perhaps set a cap for the maximum number we import (e.g., 30 or however many the front end UI accommodates). I don't expect us to run into this -- my guess is still that most users will have between 1 - 15 layouts. But, this is one way that the app could break.

* Also verify that the data does not have any missing values (i.e., no `NaN`s).

bkmartinjr:
Agree with "max layouts" limit.   Great addition.

On: 

> * Also verify that the data does not have any missing values (i.e., no `NaN`s).

How is this different than "only finite values"?     For dtype float, finite == (!NaN && !Inf)

sidneymbell:
Cool. I recognize “finite” validation technically captures “not `NaN`“, but wanted to call it out explicitly.

bkmartinjr:
It implies more than !NaN which is why I was trying to be more precise in our terminology.   We have to handle +/- Inf as well.

[...]

Issue #594, opened 2019-02-11
Switch layouts from client


colinmegill:
Switch back and forth between t-sne & umap, for instance

Thanks to @dpeerlab for the suggestion

sidneymbell:
I love this. I think it could buy us a lot of additional functionality, too (e.g., trajectories via diffmap layout, spatial via coordinate layout, etc.)

neuromusic:
In the scanpy-centric version of this, it would be "switch between any embedding in `obs.X_<embedding>` from client"

sidneymbell:
Many thanks to  Loyal Goff for his input!
"I go back and forth between these different views of the data, being able to see them side by side would be a game changer." 

E.g., showing umap in the main viz, with tsne / diffmap / etc. in tray(s).

His request was to have a dropdown that enables assignment of embeddings to the main visualization or to any of 1-3 trays at the bottom (in place of the scatterplot). Selecting/brushing in one shows the same cells in all others.

colinmegill:
![image](https://user-images.githubusercontent.com/1770265/56929733-44e16300-6a8f-11e9-9b29-6db4778c5a36.png)

neuromusic:
omg @colinmegill that icon is amazing

Loyale:
I’m so jazzed for this.

colinmegill:
See also #593

sidneymbell:
@colinmegill - Love it! At one point, we'd also talked about potentially putting this in the 'visualization settings' menu? 

![image](https://user-images.githubusercontent.com/12618847/56931147-e10d6900-6a93-11e9-82e6-5bc020eda383.png)

colinmegill:
Yes, but I think in a submenu. Ie.: 

Visualization settings
     histograms
          clipping
     graph
          switch embeddings

colinmegill:
I think this is a larger issue, and earlier opened #736

colinmegill:
For now, we'll implement as its own button, as a radio, to emphasize discoverability

neuromusic:
I think the radio button makes sense. Much of the desire for this feature comes from wanting to toggle between layouts during the exploration process, whereas the visualization menu strikes me as more appropriate for settings that we expect the user to set once and forget.

bkmartinjr:
[...]

PR #766, opened 2019-05-15
add multi-layout support to back-end

bkmartinjr:
Partial implementation of multi-layout support (issues #594 and #725), focused on the changes required for back-end and CLI support.

This PR does not expose multiple layouts on the front-end (it just picks one to display,  preferentially choosing umap, then tsne, then pca, and then an arbitrary one).   Until front-end support is completed, you can force display of a named layout by specifying only that embedding on the CLI options (eg, `cellxgene launch foo.h5ad --layout umap`).

The remainder of this work will come in a later PR, focused on the front-end (user layout choice).

### DIFF
--- a/client/src/util/stateManager/universe.js
+++ b/client/src/util/stateManager/universe.js
@@ -107,11 +107,10 @@ function LayoutFBSToDataframe(arrayBuffer) {
   /*
   TODO: XXX
 
-  Temporary code to support the progressive introduction of multi-layout
-  support.  For now, we search for one of the following in the layouts
-  and use it if we find it.  If not, we just use the first available.
-
-  umap, then tsne, then pca, then whatever is first
+  TEMPORARY CODE AND COMMENT to support the progressive implementation
+  of multi-layout support.  For now, we search for one of the following 
+  in the layouts and use it if we find it: umap, then tsne, then pca, 
+  then whatever is first in the list.
   */
   let layoutIndex = 0;
   ["umap", "tsne", "pca"].some(name => {
--- a/client/src/util/stateManager/universe.js
+++ b/client/src/util/stateManager/universe.js
@@ -108,12 +108,22 @@ function LayoutFBSToDataframe(arrayBuffer) {
   TODO: XXX
 
   Temporary code to support the progressive introduction of multi-layout
-  support.  For now, just accept the first layout in the FBS, and
-  assume that is the desired default.
+  support.  For now, we search for one of the following in the layouts
+  and use it if we find it.  If not, we just use the first available.
+
+  umap, then tsne, then pca, then whatever is first
   */
+  let layoutIndex = 0;
+  ["umap", "tsne", "pca"].some(name => {
+    const idx = fbs.colIdx.indexOf(`${name}_0`);
+    if (idx !== -1) {
+      layoutIndex = idx;
+    }
+    return idx !== -1;
+  });
   const df = new Dataframe.Dataframe(
     [fbs.nRows, 2],
-    [fbs.columns[0], fbs.columns[1]],
+    [fbs.columns[layoutIndex], fbs.columns[layoutIndex + 1]],
     null,
     new Dataframe.KeyIndex(["X", "Y"])
   );
--- a/client/src/util/stateManager/universe.js
+++ b/client/src/util/stateManager/universe.js
@@ -78,6 +78,7 @@ function AnnotationsFBSToDataframe(arrayBuffer) {
   The application has strong assumptions that all scalar data will be
   stored as a float32 or float64 (regardless of underlying data types).
   For example, clipping of value ranges (eg, user-selected percentiles)
+  depends on the ability to use NaN in any numeric type.
 
   All float data from the server is left as is.  All non-float is promoted
   to an appropriate float.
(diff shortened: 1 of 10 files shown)


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -166,11 +167,58 @@ def _validate_and_initialize(self):
...
+            else:
+                valid_layouts.append(layout)
+
+        if len(valid_layouts) == 0:
+            raise PrepareError(f"No valid layout data.")
+
+        # cap layouts to MAX_LAYOUTS
+        self.config['layout'] = valid_layouts[0:MAX_LAYOUTS]
+
+    def _is_valid_layout(self, arr):
```
csweaver: Add `@requires_data` or `n_obs` as a parameter

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -166,11 +167,58 @@ def _validate_and_initialize(self):
...
+            c) cap total list of layouts at global const MAX_LAYOUTS
+        """
+        layouts = self.config['layout']
+        # handle default
+        if layouts is None or len(layouts) == 0:
+            # load default layouts from the data.
+            layouts = [key[2:] for key in self.data.obsm_keys() if type(key) == str and key.startswith("X_")]
+            if len(layouts) == 0:
+                raise PrepareError(
+                    f"Unable to find precomputed layouts within the dataset."
```
csweaver: Nit: 
```suggestion
                    f"Unable to find any precomputed layouts within the dataset."
```

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -166,11 +167,58 @@ def _validate_and_initialize(self):
...
+                    f"You can run `cellxgene prepare --layout {self.config['layout']} <datafile>` "
```
csweaver: I don't want to push `prepare` too hard, could we add in the text about calculating a layout and saving it back to the h5ad too?

server/cli/launch.py
```diff
@@ -25,8 +24,8 @@
 @click.option(
     "--layout",
     "-l",
-    type=click.Choice(MODES),
-    default="umap",
+    default=[],
+    multiple=True,
     show_default=True,
     help="Method for layout."
```
csweaver: Can we add examples from MODES to the help text?

client/src/util/stateManager/universe.js
```diff
@@ -98,13 +99,31 @@ function AnnotationsFBSToDataframe(arrayBuffer) {
...
     throw new Error("Unexpected layout data type returned from server");
   }
+
+  /*
+  TODO: XXX
+
+  Temporary code to support the progressive introduction of multi-layout
+  support.  For now, we search for one of the following in the layouts
+  and use it if we find it.  If not, we just use the first available.
+
```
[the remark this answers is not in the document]
bkmartinjr: sorry, I am unclear which is extra - can you clarify?

(also keep in mind that this comment will be deleted once I finish the PR - this is only in the code because we are trying to keep the PRs small)

client/src/util/stateManager/universe.js
```diff
@@ -98,13 +99,31 @@ function AnnotationsFBSToDataframe(arrayBuffer) {
...
+  umap, then tsne, then pca, then whatever is first
+  */
+  let layoutIndex = 0;
+  ["umap", "tsne", "pca"].some(name => {
```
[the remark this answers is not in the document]
bkmartinjr: it is temporary code, as the comment indicates.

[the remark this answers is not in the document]
bkmartinjr: also, FYI - this is one of the recommended patterns suggested at the MDN web site...   I'm also not in love with it, but others (eg, reduce, or a for loop) are also ugly/verbose in their own way.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: fixed

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: fixed

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: This text is copied from a similar warning/error in the layout route :-)

I think that adequate explanation may require more prose than comfortably fits in a warning.   I think we should add this to a data prep document in end user docs (a long outstanding need).  

I will simply remove the text.

server/cli/launch.py
bkmartinjr: This points out a confusing issue.   The argument takes 'layout_name', e,g 'umap', but the data must contain 'X_layout_name', eg, 'X_umap'

i wonder if the CLI shouldn't just require the literal variable name, eg, 'X_umap'

It would be less confusing IMHO.

client/src/util/stateManager/universe.js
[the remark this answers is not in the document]
csweaver: line 113 is extra whitespace it seems. not a major issue.

client/src/util/stateManager/universe.js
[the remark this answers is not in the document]
csweaver: Yep, I have no problem with the actual code. I just would want to abstract it behind a function if it makes it into permanent production. No change necessary.

server/cli/launch.py
csweaver: Message looks good. I see the discussion happening on slack as far as including the X_. I'll trust @sidneymbell  on this.

client/src/util/stateManager/universe.js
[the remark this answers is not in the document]
bkmartinjr: removed.

server/cli/launch.py
bkmartinjr: @sidneymbell did not like this approach, so I have reverted and changed the help text to `"Layout name, eg, 'umap'."`

csweaver: That works, too. I don't have a good instinct as to which is more informative and less confusing.

bkmartinjr: ditto.   so I'm going simple, and we can refine from here....

PR #768, opened 2019-05-17
coordinate system fixes for embedded graph

### LINKED ISSUE #655: Many cells out of frame on load
It would be great if all cells were visible (at least on load) regardless of aspect ratio.
@colinmegill is showing me that this squishes when you decrease height but clips when you decrease x.

I have the dataset if you want it. 

Screen Shot 2019-03-20 at 1 43 17 PM
Screen Shot 2019-03-20 at 1 43 04 PM

### LINKED ISSUE #722: main embedding pan does not have 1:1 drag with cursor
in pan/zoom mode, if you drag the main embedding, it will not move 1:1 with the cursor.  Underlying view move is scaled differently (slightly less on my machine).

found by @colinmegill

bkmartinjr:
There were a variety of issues with the embedded graph view:
* it did not correctly transform to/from the webgl coordinate system, and would occasionally generate coordinates outside the expected [-1,1] range (thereby getting clipped)
* when the screen aspect ratio was below 1, it would start clipping on the short axis rather than scaling to fit the entire embedding
* as a result of the previous issue, initial layout would often have part of the graph off-canvas (not visible)

Code changes:
* regl point rendering code will add a scale factor to the default projection when the aspect ratio is less than one.
* old attempt to center the graph by mass removed (this was causing the invalid coordinates, and is obsolete now that the entire embedding is visible upon load).
* camera default pan speed is 1.0

UX changes from the user perspective:
* at startup, the embedded graph will now fill the entire viewport (ie, is completely visible).  It will be centered between the min/max values on each axis.
* mouse pan (when in pan/zoom mode) speed is 1:1 (1 px move, 1 px translate)
* when the window resizes, on either axis, the graph will scale with the canvas

Fixes #655 
Fixes #722


bkmartinjr:
confirmed fixed using dataset from @batson - thanks Josh!

client/src/components/graph/drawPointsRegl.js
```diff
@@ -1,5 +1,6 @@
 // jshint esversion: 6
 const mat4 = require("gl-mat4");
+const mat3 = require("gl-vec3");
```
freeman-lab: shouldn't it be `const vec3 = require("gl-vec3");`?

client/src/components/graph/drawPointsRegl.js
```diff
@@ -38,7 +39,20 @@ export default function(regl) {
...
+        const aspectRatio = viewportWidth / viewportHeight;
+        let m = mat4.perspective(
+          [],
+          Math.PI / 2,
+          viewportWidth / viewportHeight,
+          0.01,
+          1000
+        );
+        if (aspectRatio < 1) {
+          m = mat4.scale(m, m, mat3.fromValues(1, 1, 1 / aspectRatio));
```
freeman-lab: similarly here should be `m = mat4.scale(m, m, vec3.fromValues(1, 1, 1 / aspectRatio));` if you make the change above

freeman-lab:
nice fixes! made one small comment on naming

client/src/components/graph/drawPointsRegl.js
bkmartinjr: wow, not sure what caused that brain fault.  Fixing.

client/src/components/graph/drawPointsRegl.js
bkmartinjr: will fix

bkmartinjr:
merge of this PR will await the e2e test update

Issue #643, opened 2019-03-12
Improve contributor docs


csweaver:
Topics to include
- style guide
- setting up development environment for client
- setting up development environment server
- how to run tests locally
- guide to cellxgene's CI integration
- how to develop in docker
- link to how to test networked apps
- link to PR guidelines
- link to roadmap
- creating issue guidelines

bkmartinjr:
@csweaver - there is a CONTRIBUTING.md in the tree.   Is this issue resolved?

sidneymbell:
I think it's partially resolved; could use some additional content.

sidneymbell:
Fixed via #1030

PR #769, opened 2019-05-17
Add developer guidelines

csweaver:
Added developer guidelines for getting dev-env setup, code linting, and running tests.

Fairly rough. I'd like to add more rationale eventually, but I think even this first pass will help new developers.

Addresses some of #643

### DIFF
--- a/dev_docs/developer_guidelines.md
+++ b/dev_docs/developer_guidelines.md
@@ -34,7 +34,7 @@ We use `flake8` to lint code. Travis CI runs `flake8 server`.
 ## Client dev
 ### Install
 1. Install prereqs for client: `npm install --prefix client/ client`
-2. Install cellxgene server: `pip install -e .` Caveat: this will not build the production client package - you must use the [server install](#install-1) instructions below to serve web assets.
+2. Install cellxgene server: `pip install -e .` Caveat: this will not build the production client package - you must use the [server install](#install) instructions below to serve web assets.
 
 ### Launch
 To launch with hot reloading you need to launch the server and the client separately. Node's hot reloading starts the client on its own node server and auto-refreshes when changes are made.
@@ -76,14 +76,14 @@ End to end tests use two env variables:
 On CI the end to end tests are run with `JEST_ENV` set to `prod` using the `smoke-test` npm script
 
 To run end to end tests as they will be run on CI
-1. cellxgene should be built and installed as [specified in server dev](#install-1)
+1. cellxgene should be built and installed as [specified in server dev](#install)
 2. `export JEST_ENV='prod'`
 3. `export JEST_CXG_PORT='5000'`
 4. Run `npm run --prefix client/ smoke-test`
 
 Run end to end tests interactively during development
-1. cellxgene should be installed as [specified in client dev](#install-2)
-2. Follow [launch](#launch-2) instructions for client dev
+1. cellxgene should be installed as [specified in client dev](#install-1)
+2. Follow [launch](#launch-1) instructions for client dev
 3. Run `npm run --prefix client/ e2e`
 4. To debug a failing test `export JEST_ENV='debug'` and re-run.
 
--- a/dev_docs/developer_guidelines.md
+++ b/dev_docs/developer_guidelines.md
@@ -7,7 +7,6 @@
 
 [See dev section of README](../README.md)
 
-
 ## Server Dev
 ### Install
 * Build the client and put static files in place: `make build-for-server-dev`
@@ -36,7 +35,7 @@ We use `flake8` to lint code. Travis CI runs `flake8 server`.
 2. Install cellxgene server: `pip install -e .` Caveat: this will install necessary files for launching the cellxgene REST API but, you will need to build the client if you want to be able to view cellxgene in the browser when launched from the CLI.
 
 ### Launch
-To launch with hot reloading you need to launch the server and the client separately. Node's hot reloading starts the client on it's own node server and auto-refreshes when changes are made.
+To launch with hot reloading you need to launch the server and the client separately. Node's hot reloading starts the client on its own node server and auto-refreshes when changes are made.
 1. Launch server (the client relies on the REST API being available): `cellxgene launch [options] <datafile>`
 2. Launch client: `npm run start`
 3. Client will be served on localhost:3000
@@ -54,6 +53,8 @@ Run `npm run unit-test`
 * You can also install/launch the server side code from npm scrips (requires python3.6 with virtualenv) `npm run backend-dev`
 
 ## Running tests
+Client and server tests run on Travis CI for every push, PR, and commit to master on github. Smoke tests run nightly on master only. 
+
 ### Server
 Install development requirements `pip install -r server/requirements-dev.txt`
 Run tests `pytest server/test`
@@ -72,7 +73,7 @@ Smoke tests use two env variables
 
 Run smoke tests locally
 1. cellxgene should be built and installed
-2. run `npm run --prefix client/ smoke-test` from cellxgene directory
+2. Run `npm run --prefix client/ smoke-test` from cellxgene directory
 
 Run tests interactively
 1. cellxgene should be installed
--- a/dev_docs/developer_guidelines.md
+++ b/dev_docs/developer_guidelines.md
@@ -55,7 +55,7 @@ In `client/` directory run `npm run unit-test`
 * You can also install/launch the server side code from npm scrips (requires python3.6 with virtualenv) in `client/` directory run `npm run backend-dev`
 
 ## Running tests
-Client and server tests run on Travis CI for every push, PR, and commit to master on github. Smoke tests run nightly on master only. 
+Client and server tests run on Travis CI for every push, PR, and commit to master on github. End to end tests run nightly on master only. 
 
 ### Server unit tests
 Install development requirements `pip install -r server/requirements-dev.txt`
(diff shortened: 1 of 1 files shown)


dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+2. Run tests `pytest server/test`
+
+### Tips
+* Install in a virtualenv
+* May need to rebuild/reinstall when you make client changes
+
+## Client Dev
+### Install
+1. Install prereqs for client: `npm install --prefix client/ client`
+2. Install cellxgene server: `pip install -e .` Caveat: this will install necessary files for launching the cellxgene REST API but, you will need to build the client if you want to be able to view cellxgene in the browser when launched from the CLI.
```
bkmartinjr: not sure I understand what the Caveat is saying....

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Launch
+To launch with hot reloading you need to launch the server and the client separately. Node's hot reloading starts the client on its own node server and auto-refreshes when changes are made.
+1. Launch server (the client relies on the REST API being available): `cellxgene launch [options] <datafile>`
+2. Launch client: `npm run start`
```
bkmartinjr: from within the client directory...

csweaver: good catch, will fix

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+3. Client will be served on localhost:3000
+
+### Build
+To build only the client: `make build-client`
```
bkmartinjr: from which directory?

bkmartinjr: maybe prefix all of these instructions with "run from top-level directory unless otherwise specified"?

dev_docs/developer_guidelines.md
csweaver: I wanted to say the REST API will work but not any HTML pages or other static files. Rephrasing help is welcome.

dev_docs/developer_guidelines.md
csweaver: will clarify

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+### Test
+Run `npm run unit-test`
+
+### Tips
+* You can also install/launch the server side code from npm scrips (requires python3.6 with virtualenv) `npm run backend-dev`
+
+## Running tests
+Client and server tests run on Travis CI for every push, PR, and commit to master on github. Smoke tests run nightly on master only. 
+
+### Server
```
bkmartinjr: these are really "Server Unit Tests"

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+Install development requirements `pip install -r server/requirements-dev.txt`
+Run tests `pytest server/test`
+
+### Client
```
bkmartinjr: These are really "Client Unit Tests"

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+Run `npm run unit-test`
+
+### Smoke tests
```
bkmartinjr: these are really "End-to-end Tests"

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+Smoke tests use two env variables
+* `JEST_ENV` - environment to run smoke tests. Default `dev`
+    * `prod` - run headless with no slowdown, chromium will not open.
+    * `dev` - opens chromimum, runs tests with minimal slowdown, close on exit.
+    * `debug` - opens chromium, runs tests with 100ms slowdown, dev tools open, chrome stays open on exit.
+* `JEST_CXG_PORT` - port that smoke tests are being run on. Default `3000` (client hosted port).
+
+Run smoke tests locally
+1. cellxgene should be built and installed
```
bkmartinjr: which build target?  I'm not sure what 'built and installed' means...   Installed from PyPi?  Where?  etc.

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+2. Run `npm run --prefix client/ smoke-test` from cellxgene directory
```
bkmartinjr: I think you mean 'top-level directory', as the user may not have named their clone dir 'cellxgene'

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+Run tests interactively
+1. cellxgene should be installed
```
bkmartinjr: see above comment - not sure what 'installed' means.  Is there a command or build target?

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+Install development requirements `pip install -r server/requirements-dev.txt`
+Run tests `pytest server/test`
+
+### Client
+Run `npm run unit-test`
+
+### Smoke tests
+
+Smoke tests use two env variables
+* `JEST_ENV` - environment to run smoke tests. Default `dev`
```
bkmartinjr: which target does our CI use?  (ie, which target should be used to test prior to pushing)?

dev_docs/developer_guidelines.md
csweaver: I think I need to define that for all tests.

dev_docs/developer_guidelines.md
csweaver: will fix

dev_docs/developer_guidelines.md
csweaver: I'm going to clarify that point of what built and installed means

dev_docs/developer_guidelines.md
bkmartinjr: given that these are instructions for client dev, I'm not sure the caveat is needed.

alternatively, you could say:  "Caveat: this will not build the production client package - you must use the Launch instructions below to serve web assets."

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Tips
+* You can also install/launch the server side code from npm scrips (requires python3.6 with virtualenv) `npm run backend-dev`
+
+## Running tests
+Client and server tests run on Travis CI for every push, PR, and commit to master on github. Smoke tests run nightly on master only. 
+
+### Server
+Install development requirements `pip install -r server/requirements-dev.txt`
+Run tests `pytest server/test`
```
ttung: btw, if you're not using the parallel version of pytest, you may want to consider it. :)

bkmartinjr: @csweaver - this seems like an excellent idea give how long it takes to run our tests....

ttung: pytest-xdist is your friend.

csweaver: So it's would be a little complicated considering some of our tests launch the server there might be port collisions. (Not extremely complicated, just not trivial).

As for the speed of the tests, they actually only take ~13 seconds for all of them. The reason the travis build takes so long is building the client. So for true speed improvements we should look to webpack 😁

dev_docs/developer_guidelines.md
csweaver: I clarified how travis runs the smoke tests

csweaver:
@bruce, Thanks! I think I addressed all issues, care to take another look?

dev_docs/developer_guidelines.md
bkmartinjr: file the webpack issue and assign to me - happy to work on it!

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,89 @@
...
+We use `prettier` to lint the code.
+
+### Test
+In `client/` directory run `npm run unit-test`
+
+### Tips
+* You can also install/launch the server side code from npm scrips (requires python3.6 with virtualenv) in `client/` directory run `npm run backend-dev`
+
+## Running tests
+Client and server tests run on Travis CI for every push, PR, and commit to master on github. Smoke tests run nightly on master only.
```
bkmartinjr: s/smoke test/end-to-end tests/g

csweaver: Aaaaahhh! I thought I caught all of those, but then I introduced a new one 🤦‍♀

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,89 @@
...
+2. Run tests `pytest server/test`
+
+### Tips
+* Install in a virtualenv
+* May need to rebuild/reinstall when you make client changes
+
+## Client dev
+### Install
+1. Install prereqs for client: `npm install --prefix client/ client`
+2. Install cellxgene server: `pip install -e .` Caveat: this will not build the production client package - you must use the [server install](#install-1) instructions below to serve web assets.
```
bkmartinjr: are all of these links correct?

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,89 @@
...
+End to end tests use two env variables:
+* `JEST_ENV` - environment to run end to end tests. Default `dev`
+    * `prod` - run headless with no slowdown, chromium will not open.
+    * `dev` - opens chromimum, runs tests with minimal slowdown, close on exit.
+    * `debug` - opens chromium, runs tests with 100ms slowdown, dev tools open, chrome stays open on exit.
+* `JEST_CXG_PORT` - port that end to end tests are being run on. Default `3000` (client hosted port).
+
+On CI the end to end tests are run with `JEST_ENV` set to `prod` using the `smoke-test` npm script
+
+To run end to end tests locally
```
bkmartinjr: I'm struggling to understand the difference between these two sets of instructions ("test locally" and "test interactively").    When would I use each?  What are their differences?

dev_docs/developer_guidelines.md
csweaver: fixed. `install` is first, then `install-1`

dev_docs/developer_guidelines.md
csweaver: reworded them. the first is to match travis. the second is interactive development.

dev_docs/developer_guidelines.md
```diff
@@ -0,0 +1,91 @@
...
+
+To run end to end tests as they will be run on CI
+1. cellxgene should be built and installed as [specified in server dev](#install)
+2. `export JEST_ENV='prod'`
+3. `export JEST_CXG_PORT='5000'`
+4. Run `npm run --prefix client/ smoke-test`
+
+Run end to end tests interactively during development
+1. cellxgene should be installed as [specified in client dev](#install-1)
+2. Follow [launch](#launch-1) instructions for client dev
```
bkmartinjr: I think this needs to be specific about how it is launched (eg, which dataset is launched)

PR #776, opened 2019-05-19
layout change UI

### LINKED ISSUE #594: Switch layouts from client

### LINKED ISSUE #725: support loading arbitrary layouts

bkmartinjr:
This PR completes the layout switch UI.  Changes:
* add tool bar menu to switch between available layouts
* add layout schema information to the REST API

Fixes #594 
Fixes #725


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -115,6 +115,7 @@ def _create_schema(self):
                 "type": str(self.data.X.dtype),
             },
             "annotations": {"obs": [], "var": []},
+            "layout": {"obs": [], "var": []}
```
csweaver: With no plans on the horizon to support var layout, I'd just as soon leave it out.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -140,6 +141,14 @@ def _create_schema(self):
                     )
                 self.schema["annotations"][ax].append(ann_schema)
 
+        for layout in self.config['layout']:
+            ly_schema = {
```
csweaver: nit: `ly` is not how I would abbreviate layout (`lo` could be defensible, but I'd prefer `layout`)

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: fixed

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: fixed

PR #780, opened 2019-05-21
gui multiprocess - experimental feature

csweaver:
Switch data load/server to multiprocessing instead of threads. This fixes the issue where extremely large files would block the UI and exit of the app (~750k cells) even in the multithreaded mode.

On file selection a new process will start to launch the server and load the data.  On exit or error this process will be killed. When the data has loaded, the browser navigates to the cellxgene url and is revealed in the gui window. (FYI we can't separate the server process and the data load process because the anndata data object cannot be pickled - what multiprocessing uses for interprocess serialization for communication).

I turned off reload since I wanted to put this out before I went on PTO for the next two days, but I know how to put it back in.

Also, I have a new thread that waits for the sever to start before navigating to it -- it's the same as how we do it in our unit tests.

### DIFF
--- a/server/gui/main.py
+++ b/server/gui/main.py
@@ -18,7 +18,7 @@ from server.utils.constants import MODES
 # TODO remember this or calculate it?
 WIDTH = 1024
 HEIGHT = 768
-GUI_PORT = 8004
+GUI_PORT = 8000
 
 class MainWindow(QMainWindow):
     def __init__(self):
@@ -36,7 +36,7 @@ class MainWindow(QMainWindow):
         # Strong focus - accepts focus by tab & click
         self.setFocusPolicy(Qt.StrongFocus)
         self.setupLayout()
-        self.setupMenu()
+        # self.setupMenu()
 
     def setupLayout(self):
         self.resize(WIDTH, HEIGHT)
@@ -73,6 +73,7 @@ class MainWindow(QMainWindow):
             self.stacked_layout.addWidget(self.container, 1, 0)
 
     def setupMenu(self):
+        # TODO add communication to subprocess on reload
         main_menu = self.menuBar()
         file_menu = main_menu.addMenu('File')
         load_action = QAction("Load file...", self)
@@ -171,7 +172,6 @@ class LoadWidget(QFrame):
         self.title = splitext(basename(file_name))[0]
         if file_name:
             self.signals.selectedFile.emit(file_name)
-
 
     def onDataReady(self):
         site_ready_worker = SiteReadyWorker(self.window().url)
--- a/server/gui/main.py
+++ b/server/gui/main.py
@@ -8,9 +8,8 @@ from cefpython3 import cefpython as cef
 from PySide2.QtCore import *
 from PySide2.QtWidgets import *
 
-from server.app.app import Server
 from server.gui.browser import CefWidget, CefApplication
-from server.gui.workers import DataLoadWorker, ServerRunWorker
+from server.gui.workers import Worker, SiteReadyWorker
 from server.gui.utils import WINDOWS, LINUX, MAC, FileLoadSignals, Emitter, WorkerSignals
 from server.utils.constants import MODES
 
@@ -19,6 +18,7 @@ from server.utils.constants import MODES
 # TODO remember this or calculate it?
 WIDTH = 1024
 HEIGHT = 768
+GUI_PORT = 8004
 
 class MainWindow(QMainWindow):
     def __init__(self):
@@ -29,9 +29,8 @@ class MainWindow(QMainWindow):
         self.load_emitter = Emitter(self.parent_conn, WorkerSignals)
         self.emitter_thread = threading.Thread(target=self.load_emitter.run, daemon=True)
         self.emitter_thread.start()
-        self.server = Server()
-        self.server.create_app()
-        self.runServer()
+        self.worker = None
+        self.url = f"http://localhost:{GUI_PORT}/"
         self.setWindowTitle("cellxgene")
 
         # Strong focus - accepts focus by tab & click
@@ -90,11 +89,6 @@ class MainWindow(QMainWindow):
         if self.cef_widget.browser:
             self.cef_widget.browser.CloseBrowser(True)
             self.clearBrowserReferences()
-
-    def runServer(self):
-        worker = ServerRunWorker(self.server.app, host="127.0.0.1", port=8000)
-        self.httpd = threading.Thread(target=worker.run, daemon=True)
-        self.httpd.start()
 
     def clearBrowserReferences(self):
         # Clear browser references that you keep anywhere in your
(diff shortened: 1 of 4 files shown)


csweaver:
One other interesting tidbit -- setting the flask server's debug param to true causes a separate new window containing the qt app to open. I didn't go too deep on the why, since it took long enough to find out what was actually causing this bug.

server/gui/utils.py
```diff
@@ -13,13 +13,49 @@ class WorkerSignals(QObject):
...
+            getattr(self.signals, signature).emit(args)
+        else:
+            getattr(self.signals, signature).emit()
+
+    def run(self):
+        while True:
+            try:
+                signature = self.transport.recv()
+            except Exception:
+                break
```
bkmartinjr: I assume that the silent break out of the loop is to handle pipe EOF (eg, if the child process exits, the pipe is closed and you expect an EOF).  However, this will not occur because both processes keep both ends of the pipe open.

General best practices for pipes:  after the fork, the parent should close the child end of the pipe, and the child should close the parent end.   If both processes have both ends open, you will not get an EOF and your emitter won't terminate.

server/gui/main.py
```diff
@@ -155,14 +154,15 @@ def __init__(self, parent):
...
-        self.embedding_selection = MODES[idx]
+        self.embedding_selection = [MODES[idx]]
 
     def createScanpyEngine(self, file_name):
-        worker = DataLoadWorker(file_name, self.embedding_selection)
-        worker.signals.result.connect(self.onDataSuccess)
-        worker.signals.error.connect(self.onDataError)
-        self.load_worker = threading.Thread(target=worker.run, daemon=True)
-        self.load_worker.start()
+        worker = Worker(self.window().child_conn, file_name, self.title, host="127.0.0.1", port=GUI_PORT,
```
bkmartinjr: See comment in Emitter about keeping both ends of the pipe open in both processes.

server/gui/main.py
```diff
@@ -155,14 +154,15 @@ def __init__(self, parent):
...
+                        layout=self.embedding_selection)
```
bkmartinjr: do you want to be restricting to a single layout?   The back-end is no longer responsible for selecting layouts, but rather simply for enumerating them and providing all to the front-end.  See PR #766.

server/gui/utils.py
bkmartinjr: Additional thought on the silent break -- unless you watch for the emitter thread exit (which I don't see in the code), this silent break could result in the HTTP server dying (app no longer functional), and the UI not noticing.    Suggest that the exception catch at least emit an error signal, so the GUI knows something went wrong and can inform the user (and then either bail/quit, or restart the Worker).

server/gui/utils.py
```diff
@@ -13,13 +13,49 @@ class WorkerSignals(QObject):
...
         error - `str` error message
         result - `object` data returned from processing, anything
     """
     finished = Signal()
     error = Signal(str)
     result = Signal(object)
+    ready = Signal()
+
+
+class SiteReadySignals(QObject):
```
bkmartinjr: Any reason this isn't a subclass of EmittingProcess (like Worker)?

server/gui/main.py
csweaver: For now, yes. I want a UI element besides file selection in the layout. With the intention to fill out the selection UI later (after installer is proved).

server/gui/utils.py
csweaver: Because the signals aren't tied to the worker being a subprocess or a thread (If I understand your question correctly).

server/gui/utils.py
csweaver: Good feedback. Will fix emitter.

csweaver:
Yeah, I though about a few options for checking if the server is up, but I still think this is the best option.

server/gui/main.py
bkmartinjr: I don't think we necessarily want to move these UI controls out of the front-end and into the wrapper frame.  This seems like an open design question....  (I'm actually not sure what I think is the right approach, but we need mocks & UX testing on the larger issue)

csweaver: Totally agree, I just want to delay that issue until after I've got installer working but before release.

server/gui/main.py
```diff
@@ -7,33 +8,60 @@
...
+        self.load_emitter = Emitter(self.parent_conn, WorkerSignals)
+        self.emitter_thread = threading.Thread(target=self.load_emitter.run, daemon=True)
+        self.emitter_thread.start()
+        # send to load with error message?
+
+    def terminateWorker(self):
+        if not self.worker:
+            return
+        try:
+            self.worker.close()
```
bkmartinjr: Left comment on this in the other PR.  TL;DR - not sure why you can't just call `terminate()` in all cases.

server/gui/main.py
```diff
@@ -155,14 +180,19 @@ def __init__(self, parent):
...
     def createScanpyEngine(self, file_name):
-        worker = DataLoadWorker(file_name, self.embedding_selection)
-        worker.signals.result.connect(self.onDataSuccess)
-        worker.signals.error.connect(self.onDataError)
-        self.load_worker = threading.Thread(target=worker.run, daemon=True)
-        self.load_worker.start()
+        worker = Worker(self.window().parent_conn, self.window().child_conn, file_name, self.title, host="127.0.0.1", port=GUI_PORT,
+                        layout=self.embedding_selection)
+        self.window().load_emitter.signals.ready.connect(self.onDataReady)
+        self.window().load_emitter.signals.engine_error.connect(self.onEngineError)
```
bkmartinjr: is there any value in differentiating between engine and server errors?  Ie, will the UI (down the road) need to treat them differently?   If not, might be cleaner to just combine them into one handler.   If they need differentiated UI/handling, fine to keep as is (I just can't think of why we would need that).

server/gui/workers.py
```diff
@@ -1,47 +1,94 @@
...
-        else:
-            self.signals.result.emit(data_results)
+            self.emit("server_error", str(e))
+            self.emit("finished")
+            return
+        # load data
+        try:
+            args = {
+                "layout": self.layout,
+                "diffexp": "ttest",
```
bkmartinjr: `diffexp` config is obsolete/unused (AFAIK)

server/gui/workers.py
```diff
@@ -1,47 +1,94 @@
...
-            "obs_names": None,
-            "var_names": None,
-        }
+        # create server
         try:
-            data_results = ScanpyEngine(self.data_file, args)
+            server = Server()
+            server.create_app()
         except Exception as e:
             traceback.print_exc()
```
bkmartinjr: debugging code?

server/gui/workers.py
```diff
@@ -1,47 +1,94 @@
...
-        self.signals = WorkerSignals()
+        self.title = title
+        self.host = host
+        self.port = port
 
     def run(self):
+        super(Worker, self).run()
         if not self.data_file:
-            self.signals.finished.emit()
+            self.emit("finished")
```
bkmartinjr: strictly a style issue:  I remember from a previous PR review that you wanted all strings like this to be in constant variables, eg, `SIG_FINISHED`.    Should these be variables, or inline strings (I personally *like* inline strings where the variable name adds no additional value).

server/gui/main.py
csweaver: yes, if the engine fails, it's likely a bad file. If the server fails, I'd eventually want to restart the server with the same file because the error could be just one specific call. And then keep a restart counter so we don't get stuck in an infinite loop of restarts.

server/gui/workers.py
csweaver: actually, I think i had captured the traceback in the past, but I can delete that now.

server/gui/main.py
csweaver: I checked, and you're right. Terminate is always going to be called.

server/gui/workers.py
csweaver: yeah, after trying things "Colin style" as I like to call it, I like using strings better.

server/gui/workers.py
csweaver: I delete, but I've got updating diffexp in the gui in a future PR.

PR #785, opened 2019-05-23
do not hard-wire column names in annotations

### LINKED ISSUE #779: Can't load h5ad with "name" obs
It appears that cellxgene chokes when trying to load a dataset containing an obs column called "name". It'd be great if it could at least provide a more meaningful error message and ideally provide compatibility with datasets containing such a column.

```python
import scanpy as sc

import numpy as np
import pandas as pd

# create an AnnData instance with random data
X = np.random.randint(0, 1000, size=(1000, 50))

adata = sc.AnnData(X)
adata.obs = pd.DataFrame({"x":np.random.random(size=(1000,))})
sc.tl.tsne(adata)

adata.write("adata.h5ad")

# now add a column "name"
adata.obs["name"] = ["a", "b"] * 500
adata.write("adata_with_name.h5ad")
```

Running on adata_with_name.h5ad causes the following error:

```
$ cellxgene launch -v --debug -l tsne  adata_with_name.h5ad
[cellxgene] Loading data from adata_with_name.h5ad.
  File "/mnt/home/nspies/.local/bin/cellxgene", line 11, in <module>
    load_entry_point('cellxgene', 'console_scripts', 'cellxgene')()
  File "/mnt/home/nspies/.local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/home/nspies/.local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/mnt/home/nspies/.local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/mnt/home/nspies/.local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mnt/home/nspies/.local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/mnt/home/nspies/Downloads/cellxgene/server/cli/launch.py", line 185, in launch
    server.attach_data(ScanpyEngine(data, args), title=title)
  File "/mnt/home/nspies/Downloads/cellxgene/server/app/scanpy_engine/scanpy_engine.py", line 35, in __init__
    self._validate_and_initialize()
  File "/mnt/home/nspies/Downloads/cellxgene/server/app/util/utils.py", line 42, in wrapped_function
    return func(self, *args, **kwargs)
  File "/mnt/home/nspies/Downloads/cellxgene/server/app/scanpy_engine/scanpy_engine.py", line 169, in _validate_and_initialize
    self._validate_data_types()
  File "/mnt/home/nspies/Downloads/cellxgene/server/app/util/utils.py", line 42, in wrapped_function
    return func(self, *args, **kwargs)
  File "/mnt/home/nspies/Downloads/cellxgene/server/app/scanpy_engine/scanpy_engine.py", line 234, in _validate_data_types
    datatype = curr_axis[ann].dtype
  File "/usr/local/lib64/python3.7/site-packages/pandas/core/generic.py", line 4376, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'dtype'
```
Running `cellxgene launch -l tsne adata.h5ad` works fine.

bkmartinjr:
Previous to this PR, the app reserved the obs["name"] and var["name"] to contain unique human-readable values.   This is now parameterized to select a name that will not stomp on existing data.  The previous support for user-specific columns, via the CLI options `--obs-names` and `--var-names` is unchanged.

In addition, the existing but unchecked requirement for uniqueness on var and obs column names is now checked (`adata.obs.columns.is_unique and adata.var.columns.is_unique`).

Changes:
* the back-end /schema now contains an `index` field for both var and obs annotations, indicating which annotation column has the unique, user-readable names.    Example:
```
  schema: {
    annotations: {
      obs: {
        index: "name",
        columns: [
          { name: "name", type: "string" },
          { name: "field1", type: "int32" },
        ]
      },
...
```

* the front-end uses this parameter to determine which dataframe column contains names (eg, for the gene search box)
* the CLI `--obs-names` and `--var-names` options directly set this value in the scanpy engine
* obs and var column names are tested for uniqueness, and an error issued if they are not
* the obs_name and var_name (aka front-end index) columns must also be unique. This test existed previously. 

Fixes #779

### DIFF
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -164,6 +164,10 @@ class ScanpyEngine(CXGDriver):
 
     @requires_data
     def _validate_and_initialize(self):
+        # var and obs column names must be unique
+        if not self.data.obs.is_unique or not self.data.var.is_unique:
+            raise KeyError(f"All annotation column names must be unique.")
+
         self._alias_annotation_names(Axis.OBS, self.config["obs_names"])
         self._alias_annotation_names(Axis.VAR, self.config["var_names"])
         self._validate_data_types()
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -51,14 +51,17 @@ class ScanpyEngine(CXGDriver):
         }
 
     @staticmethod
-    def _create_unique_column_name(df, prefix):
-        """ given a dataframe and a name prefix, return a column name which does not
-            exist in the dataframe.
+    def _create_unique_column_name(df, col_name_prefix):
+        """ given the columns of a dataframe, and a name prefix, return a column name which 
+            does not exist in the dataframe, AND which is prefixed by `prefix`
+
+            The approach is to append a numeric suffix, starting at zero and increasing by
+            one, until an unused name is found (eg, prefix_0, prefix_1, ...).
         """
         suffix = 0
-        while f"{prefix}{suffix}" in df.columns:
+        while f"{col_name_prefix}{suffix}" in df:
             suffix += 1
-        return f"{prefix}{suffix}"
+        return f"{col_name_prefix}{suffix}"
 
     def _alias_annotation_names(self):
         """
@@ -83,11 +86,11 @@ class ScanpyEngine(CXGDriver):
                         "Please prepare data to contain unique index values, or specify an "
                         "alternative with --{ax_name}-name."
                     )
-                name = self._create_unique_column_name(df_axis, "name_")
+                name = self._create_unique_column_name(df_axis.columns, "name_")
                 self.config[config_name] = name
                 # reset index to simple range; alias name to point at the
                 # previously specified index.
-                df_axis.rename_axis(name, axis=0, inplace=True)
+                df_axis.rename_axis(name, inplace=True)
                 df_axis.reset_index(inplace=True)
             elif name in df_axis.columns:
                 # User has specified alternative column for unique names, and it exists
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -52,7 +52,7 @@ class ScanpyEngine(CXGDriver):
 
     @staticmethod
     def _create_unique_column_name(df, col_name_prefix):
-        """ given the columns of a dataframe, and a name prefix, return a column name which 
+        """ given the columns of a dataframe, and a name prefix, return a column name which
             does not exist in the dataframe, AND which is prefixed by `prefix`
 
             The approach is to append a numeric suffix, starting at zero and increasing by
--- a/server/app/scanpy_engine/scanpy_engine.py
+++ b/server/app/scanpy_engine/scanpy_engine.py
@@ -50,41 +50,58 @@ class ScanpyEngine(CXGDriver):
             "diffexp_lfc_cutoff": 0.01,
         }
 
-    def _alias_annotation_names(self, axis, name):
+    @staticmethod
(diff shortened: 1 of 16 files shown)


server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -50,41 +50,58 @@ def _get_default_config():
...
+                    raise KeyError(
+                        f"Values in {ax_name}.index must be unique. "
+                        "Please prepare data to contain unique index values, or specify an "
+                        "alternative with --{ax_name}-name."
+                    )
+                name = self._create_unique_column_name(df_axis, "name_")
+                self.config[config_name] = name
+                # reset index to simple range; alias name to point at the
+                # previously specified index.
+                df_axis.rename_axis(name, axis=0, inplace=True)
```
csweaver: I don't think you need the axis=0, that's the default, isn't it?

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -50,41 +50,58 @@ def _get_default_config():
...
-
-        ax_name = str(axis)
-        df_axis = getattr(self.data, ax_name)
-        if name is None:
-            # reset index to simple range; alias "name" to point at the
-            # previously specified index.
-            df_axis.reset_index(inplace=True)
-            df_axis.rename(inplace=True, columns={"index": "name"})
-        elif name in df_axis.columns:
-            if name not in df_axis.columns:
```
csweaver: I'm a little worried about doing git blame on the above two lines 😳

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -50,41 +50,58 @@ def _get_default_config():
             "diffexp_lfc_cutoff": 0.01,
         }
 
-    def _alias_annotation_names(self, axis, name):
+    @staticmethod
+    def _create_unique_column_name(df, prefix):
```
csweaver: Can we change this method's signature: `(name, columns)` 
1. Since you are only using `df.columns` no need to pass the whole dataframe, doing that makes knowing what the function does with the dataframe harder to understand.
2. `prefix` was not what I expected it to be. I thought it was going to be what we prefix non-unique column names with.

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -50,41 +50,58 @@ def _get_default_config():
...
+        """ given a dataframe and a name prefix, return a column name which does not
+            exist in the dataframe.
```
csweaver: Can you add that it does this by appending a number?

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: true.  I added for clarity, but will remove.

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: you and me both :-).    I suspect I am likely guilty!!!

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: will make change (1), and will add better comments and variable names for prefix, in an attempt to add clarity....

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: done.

PR #793, opened 2019-05-29
fix incompatibility of flask reload and port searching

bkmartinjr:
in the 0.10 pre-release, if both --port and --debug were specified, the CLI will error out.  This is caused by flask reload causing the port code to run twice.   

Changes:
* `--debug` now only prints verbose warnings and disables browser launch. 
* a new flag, `--developer` enables hot reload and implies `--debug`
* because `--developer` and `--port` are incompatible, a new human-readable error is triggered by the combination.

So, these are legal:
* `cellxgene launch foo.h5ad --debug --port 9999` - verbose logging, serving on port 9999
* `cellxgene launch foo.h5ad --developer` - scan to find available port, verbose logging, hot reload

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -11,7 +11,7 @@ import click
 from server.app.app import Server
 from server.app.util.errors import ScanpyFileError
 from server.app.util.utils import custom_format_warning
-from server.utils.utils import find_available_port
+from server.utils.utils import find_available_port, is_port_available
 
 
 # anything bigger than this will generate a special message
@@ -143,6 +143,9 @@ security risk by including the --scripts flag. Make sure you trust the scripts t
 
     if not port:
         port = find_available_port(host)
+    else:
+        if not is_port_available(host, int(port)):
+            raise click.ClickException(f"The port selected {port} is in use, please specify an open port using the --port flag.")
 
     # Setup app
     cellxgene_url = f"http://{host}:{port}"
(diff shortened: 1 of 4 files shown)


server/cli/launch.py
```diff
@@ -199,7 +212,7 @@ def launch(
         sys.stdout = f
 
     try:
-        server.app.run(host=host, debug=debug, port=port, threaded=True)
+        server.app.run(host=host, debug=developer, port=port, threaded=True, use_reloader=developer, use_debugger=False)
```
csweaver:
```suggestion
        server.app.run(host=host, debug=debug, port=port, threaded=True, use_reloader=developer, use_debugger=False)
```
I think this was just a typo or oversight

server/cli/launch.py
```diff
@@ -143,6 +149,13 @@ def launch(
 
     if not port:
         port = find_available_port(host)
+    else:
```
csweaver: if you are feeling like it, you could flip the if else so that it's a positive `if` and not an `if not`

server/cli/launch.py
```diff
@@ -115,6 +118,9 @@ def launch(
     if extension != ".h5ad":
         raise click.FileError(basename(data), hint="file type must be .h5ad")
 
+    if developer:
```
csweaver: I think that we may not need all of debug, verbose, and developer, but it would take more time than we have to get the release out and is low priority.

PR #803, opened 2019-05-31
[Do not merge] Update pbmc3k example data

sidneymbell:
The current example dataset doesn't utilize cellxgene's full feature set. Unfortunately, there is little available metadata for this dataset. However, I was able to attach a few new kinds of data so that a fuller set of features is at least visible/usable by new users.

This PR adds:
* Categorical metadata: cluster labels for both of the currently-available clustering methods in scanpy (louvain and leiden)  
* Continuous metadata: lots of QC metrics generated by `sc.pp.calculate_qc_metrics()`  
* A few new embeddings: umap, tsne, and force directed graph  

This enables users to try out cluster occupancy and switching between embeddings.

To avoid confusion, I kept the normalization and preprocessing consistent with the [scanpy tutorial](https://scanpy-tutorials.readthedocs.io/en/latest/pbmc3k.html).

The (minimal) notebook used to process this dataset is available [here](https://github.com/chanzuckerberg/cellxgene-vignettes/blob/master/dataset-processing/pbmc3k-processing.ipynb).


bkmartinjr:
Agree with @csweaver.  I think it makes the most sense to split "demo" and "test" data:
* demo data on an S3 bucket or other easy to access location, for all users.  Maybe more than one data set.
* small test data, part of the repo, and only used for e2e testing.  Allows it to be revised/versioned with the tests & code.

Our "example" could be removed, and replaced with the above two.

sidneymbell:
I think this all makes a lot of sense! I'll stick this on an open AWS bucket or something once all the demo data is good to go. Thanks.

PR #804, opened 2019-06-01
menubar

### LINKED ISSUE #728: refactor graph.js widget
graph.js has become housing for both the graph and the toolbar. We should refactor at some point to add a layout react component for the right hand side.

colinmegill:
image

Fixes #728


client/src/reducers/controls.js
```diff
@@ -138,6 +138,11 @@ const Controls = (
     /*******************************
              User Events
      *******************************/
+    case "change graph interaction mode":
```
bkmartinjr: You need to configure undo/redo handling whenever you change actions (either add/del an action, or change its semantics). 

If you simply want this to push a history undo point, as I suspect is the case, add this action type to the `saveOnActions` Set in undoableConfig.js.   If you want something more complex, let me know.

client/src/components/graph/graph.js
```diff
@@ -3,43 +3,24 @@ import React from "react";
...
 import setupSVGandBrushElements from "./setupSVGandBrush";
-import actions from "../../actions";
 import _camera from "../../util/camera";
 import _drawPoints from "./drawPointsRegl";
 import scaleLinear from "../../util/scaleLinear";
-import { World } from "../../util/stateManager";
 
 /* https://bl.ocks.org/mbostock/9078690 - quadtree for onClick / hover selections */
 
 @connect(state => ({
```
bkmartinjr: there are unused properties, eg, `libraryVersions`.  Please remove so that they do not trigger unnecessary updates.

client/src/components/graph/graph.js
```diff
@@ -188,16 +143,30 @@ class Graph extends React.Component {
...
-      layoutChoice
+      layoutChoice,
+      graphInteractionMode
     } = this.props;
-    const { reglRender, mode, regl, svg } = this.state;
+    const { reglRender, regl, svg } = this.state;
     let stateChanges = {};
 
-    if (reglRender && this.reglRenderState === "rendering" && mode !== "zoom") {
+    if (
```
bkmartinjr: these two if statements (this one and the next one) are almost identical, except for the test of `graphInteractionMode`.   They are quite hard to parse as written.   Suggest you restructure as:
* outside if that tests for initialization state (if reglRender && this.reglRenderState === "rendering")
* inside if that branches on `graphInterationMode`
That makes the actual tests much clearer.

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+import {
+  Button,
+  AnchorButton,
+  Tooltip,
+  Popover,
+  Menu,
+  MenuItem,
+  Position,
+  NumericInput,
+  Divider,
```
bkmartinjr: not used - lint

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+import { World } from "../../util/stateManager";
+import * as globals from "../../globals";
+import actions from "../../actions";
+import CellSetButton from "./cellSetButtons";
+
+@connect(state => ({
+  differential: state.differential,
+  world: state.world,
+  loading: state.controls.loading,
+  datasetTitle: state.config?.displayNames?.dataset,
```
bkmartinjr: this will be null if not defined, which seems wrong given how it is later used.    Probably should be defaulted to an empty string:  `datasetTitle: state.config?.displayNames?.dataset ?? ""`

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+          differential.celllist1,
+          differential.celllist2
+        )
+      );
+    }
+  }
+
+  clearDifferentialExpression() {
+    const { dispatch, differential } = this.props;
+    dispatch({
```
bkmartinjr: FYI: if any of these multi-action sequences have changed, we will need to reconfigure undo/redo.  If they are all the same, then no changes are needed.

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+    return (
+      <div
+        style={{
+          paddingTop: 10,
+          paddingLeft: 10,
+          paddingRight: 10,
+          backgroundColor: "white",
+          display: "flex",
+          justifyContent: "space-between"
+          // boxShadow: "0px -3px 6px 2px rgba(153,153,153,0.4)",
```
bkmartinjr: lint

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+              position="bottom"
+            >
+              <AnchorButton
+                disabled={!haveBothCellSets}
+                intent="primary"
+                data-testid="diffexp-button"
+                loading={differential.loading}
+                fill
+                type="button"
+                onClick={this.computeDiffExp.bind(this)}
```
bkmartinjr: if you define these functions as a class variable, they will be already bound to `this`, and you can get rid of the explicit bind (and make the linter happy).

In other words, define them as:
```
computeDiffExp = () => { ... }
```

bkmartinjr: there is an example in `isClipDisabled()`

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+                    <MenuItem
+                      href="https://github.com/chanzuckerberg/cellxgene"
+                      target="_blank"
+                      icon="git-branch"
+                      text="Github"
+                    />
+                    <MenuItem
+                      target="_blank"
+                      text={`cellxgene v${
+                        libraryVersions && libraryVersions.cellxgene
```
bkmartinjr: when I test, the version number does not show up in the menu.    It displays as "vnull"

bkmartinjr: the property is not present in the state `@connect`

client/src/components/menubar/index.js
```diff
@@ -0,0 +1,693 @@
...
+      type: "clear differential expression",
+      diffExp: differential.diffExp
+    });
+    dispatch({
+      type: "clear scatterplot"
+    });
+  }
+
+  render() {
+    const {
```
bkmartinjr: a large number of these properties are missing from the state `@connect`, including (not limited to):  libraryVersions, undoDisabled, redoDisabled....

client/src/components/graph/graph.js
bkmartinjr: and please make sure you add them to the menu bar, which needs them :-)

client/src/components/graph/graph.js
colinmegill: No, that's not quite correct. The inner loop has two conditionals: `=== zoom` & `!== zoom`. I can break out reglRender, however!

colinmegill: "If it's rendering and it's not zoom, stopped rendering. If it isn't rendering and it is zoom, start rendering."

client/src/components/menubar/index.js
colinmegill: Indeed, when it was omitted we had vnull

client/__tests__/e2e/e2e.test.js
```diff
@@ -154,10 +154,18 @@ describe("gene entry", async () => {
...
+    // these load asynchronously, so we need to wait for each histogram individually
+    const elements = await Promise.all(
+      testGenes.map(g => page.waitForSelector(`[data-testid='histogram-${g}']`))
+    );
+    const idHandles = await Promise.all(
+      elements.map(el => el.getProperty("id"))
+    );
+    const ids = await Promise.all(idHandles.map(h => h.jsonValue()));
+    expect(ids.map(id => id.substring("histogram_".length))).toEqual(
+      expect.arrayContaining(testGenes)
```
csweaver: Please move this code to `getAllHistograms` in `e2e/puppeteerUtils.js` and have e2e.test.js continue to use the helper function.

PR #807, opened 2019-06-04
adds section on risks of hosted instances

### LINKED ISSUE #806: clearly communicate risks of hosting cellxgene
this is related to #791 

we need to clearly document the risks of hosting cellxgene

key points that need to be communicated:

- Flask dev server is not recommended for hosted deployment
- We have no testing or official support for multi-tenancy
- Likely to fall over if too many people hit the server, especially if using differential expression
- One instance per dataset

neuromusic:
fixes #806

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,16 +52,18 @@ There are 4 key features we plan to implement in the near term.
 
 For more detail on these features and where we are going, see [our roadmap](ROADMAP.md).
 
-## supported and experimental usage modes
+## risks of hosting cellxgene
 
-_cellxgene_ is built on standard web technologies, but is currently designed as a single-user desktop application. 
+_cellxgene_ is built on standard web technologies, but is currently designed as a single-user desktop application.
+We've done this so we can prioritize the items on [our roadmap](ROADMAP.md).
+
 Some of our users have experimented with hosting _cellxgene_ either for their lab or for public use, but please note that the _cellxgene_ team does not officially support, troubleshoot, or maintain any web deployments at this time.
 
-If you choose setup _cellxgene_ as a hosted service, you should be aware of the following risks:
+If do you choose setup _cellxgene_ as a hosted service, you should be aware of the following risks:
 
 - `$ cellxgene launch` uses Flask's development server, which is not recommended for hosted deployment (see the [Flask documentation](http://flask.pocoo.org/docs/1.0/tutorial/deploy/#run-with-a-production-server))
 - We have no testing or official support for deployments where multiple users are accessing the same _cellxgene_ instance.
-- Your _cellxgene_ instance is likely to fall over if too many people access it at the same time, especially if they using functions that call the Python backend (such as differential expression, updating the layout, or coloring by gene).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -52,6 +52,18 @@ There are 4 key features we plan to implement in the near term.
 
 For more detail on these features and where we are going, see [our roadmap](ROADMAP.md).
 
+## supported and experimental usage modes
```
csweaver: This title does not relate to the content. I think the title should directly refer to hosting cellxgene.

Something like "[Unsupported] Hosting cellxgene"

README.md
```diff
@@ -52,6 +52,18 @@ There are 4 key features we plan to implement in the near term.
...
+
+_cellxgene_ is built on standard web technologies, but is currently designed as a single-user desktop application. 
+Some of our users have experimented with hosting _cellxgene_ either for their lab or for public use, but please note that the _cellxgene_ team does not officially support, troubleshoot, or maintain any web deployments at this time.
```
csweaver: I think you should add a sentence on why we made that choice since it can seem inconceivable for most users that we are not pursuing that mode currently.

README.md
```diff
@@ -52,6 +52,18 @@ There are 4 key features we plan to implement in the near term.
...
+
+If you choose setup _cellxgene_ as a hosted service, you should be aware of the following risks:
+
+- `$ cellxgene launch` uses Flask's development server, which is not recommended for hosted deployment (see the [Flask documentation](http://flask.pocoo.org/docs/1.0/tutorial/deploy/#run-with-a-production-server))
```
csweaver: well said

README.md
```diff
@@ -52,6 +52,18 @@ There are 4 key features we plan to implement in the near term.
...
+- We have no testing or official support for deployments where multiple users are accessing the same _cellxgene_ instance.
+- Your _cellxgene_ instance is likely to fall over if too many people access it at the same time, especially if they using functions that call the Python backend (such as differential expression, updating the layout, or coloring by gene).
```
csweaver: Can we find a more technical term than "fall over"? I can't think of the right word at the moment, but maybe it will come to me tomorrow.

colinmegill: Is the issue that the server runs out of RAM because it's trying to repeatedly load data into memory to do the diffexp computation?

README.md
neuromusic: 😄

README.md
neuromusic: hang?

csweaver: hang sounds good. It's the cpu, not the memory in diffexp case, I believe.

README.md
```diff
@@ -52,6 +52,20 @@ There are 4 key features we plan to implement in the near term.
 
 For more detail on these features and where we are going, see [our roadmap](ROADMAP.md).
 
+## risks of hosting cellxgene
+
+_cellxgene_ is built on standard web technologies, but is currently designed as a single-user desktop application.
+We've done this so we can prioritize the items on [our roadmap](ROADMAP.md).
```
csweaver: items -> features
```suggestion
We've done this so we can prioritize the features on [our roadmap](ROADMAP.md).
```

bkmartinjr:
@neuromusic - it might be nice to indicate whether hosted support may be considered in the future, or is *never* going to be considered due to some underlying limitation/constraint.

PR #808, opened 2019-06-04
add reload back in gui

csweaver:
Make File->Load work again in the GUI app.

I removed the file->load menu when adding multiprocessing because you can't connect signals across processes. To add it back in, instead of passing the new data to the running server process, I shut the server process down and restart. I chose this method because passing the new file would require multithreading on the server process or some other architecture change and restarting the server takes only a second or two max (a fraction of the data load time). 

In the process I made errors more generic; all errors send you to the file load page; server errors shut down the server too. I also only create the emitting process and server worker when the user selects a file instead of at the start.

(The do not merge tag is because I'm basing this PR on a branch and need the reminder to switch the base to master if the branch gets merged)

### DIFF
--- a/server/gui/main.py
+++ b/server/gui/main.py
@@ -1,4 +1,5 @@
 # flake8: noqa F403, F405
+from functools import partialmethod
 from multiprocessing import Pipe, Process
 from os.path import splitext, basename
 import sys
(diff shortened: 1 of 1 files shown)


server/gui/main.py
```diff
@@ -26,31 +27,33 @@ def __init__(self):
...
         self.emitter_thread.start()
         # send to load with error message?
 
+    def shutdownServer(self):
+        if self.worker:
+            self.terminateWorker()
+        if self.parent_conn:
+            self.parent_conn.close()
+
     def terminateWorker(self):
```
bkmartinjr: Question: why call Process.close()?  My understanding from the 3.7 docs is that it will only succeed if the underlying process has already exited.    The `terminate()` function seems to also release resources, so couldn't you replace this entire function with a call to `self.worker.terminate()` and have the same effect?

csweaver: Fixed on the other. You're right. Close was never being called.

PR #809, opened 2019-06-08
draw labels marking the centroids of category value clusters

### LINKED ISSUE #595: Show categorical labels on centroid in graph
Likely implemented as a secondary button next to `color by variable` which adds the labels. 

image

A good implementation will be smart about the color of the label, or add a white background to the label at 50% opacity for visibility. The latter is much quicker to implement, and probably better anyway.

image

This feature is potentially useful for categorical metadata which correlates to the embedding - ie., `cluster id` will work well, but variables like `gender` might congregate around the center of the graph: 

image

Thanks to @dpeerlab for the suggestion

As an aside, we've also discussed (see issue #9) 'discovering' which cluster is which by allowing a hover on a cluster in the graph to highlight the list of metadata in the sidebar. This is more problematic - is it dependent on the current `color by`? Showing the labels is probably a nicer interaction than having to chase down something in the left panel, as there's no guarantee that it will be visible if the list is long.

seve:
_EDIT: Some additional features were removed from this PR to prevent feature creep. See #811_

Categorical data is currently hard to parse when there is a wide variety of labels.  By implementing centroid labels for each category label, clusters will be more easily distinguished. The label location is calculated using the median x and y values of all relevant cells. 

- A new reducer which holds:
  - The selected metadata and category values
  - The XY position of the centroid
- A util which calculates the centroid given a category value
- An additional SVG layer on top of the existing lasso layer and necessary refactors to distinguish one another
- A file for setting up and drawing the centroid label

Fixes #595

### DIFF
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -59,7 +59,8 @@ class CategoryValue extends React.Component {
       colorScale,
       i,
       schema,
-      world
+      world,
+      centroidLabel
     } = this.props;
 
     if (!categoricalSelection) return null;
@@ -95,7 +96,11 @@ class CategoryValue extends React.Component {
         style={{
           display: "flex",
           alignItems: "baseline",
-          justifyContent: "space-between"
+          justifyContent: "space-between",
+          borderRadius: "3px",
+          background: centroidLabel.categoryIndex === categoryIndex &&
+                           centroidLabel.metadataField === metadataField
+                           ? "rgba(115, 134, 148, 0.3)":"inherit"
         }}
         data-testclass="categorical-row"
         onMouseEnter={this.handleMouseEnter.bind(this)}
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -35,7 +35,7 @@ class CategoryValue extends React.Component {
   handleMouseEnter() {
     const { dispatch, metadataField, categoryIndex } = this.props;
     dispatch({
-      type: "mouse enter",
+      type: "category value mouse hover start",
       metadataField,
       categoryIndex
     })
@@ -44,7 +44,7 @@ class CategoryValue extends React.Component {
   handleMouseExit() {
     const { dispatch, metadataField, categoryIndex } = this.props;
         dispatch({
-      type: "mouse exit",
+      type: "category value mouse hover start",
       metadataField,
       categoryIndex
     })
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -44,7 +44,7 @@ class CategoryValue extends React.Component {
   handleMouseExit() {
     const { dispatch, metadataField, categoryIndex } = this.props;
     dispatch({
-      type: "category value mouse hover start",
+      type: "category value mouse hover end",
       metadataField,
       categoryIndex
     })
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -9,7 +9,7 @@ import * as globals from "../../globals";
   categoricalSelection: state.categoricalSelection,
   colorScale: state.colors.scale,
   colorAccessor: state.colors.colorAccessor,
-  schema: state.world?.schema,
+  schema: state.world ?.schema,
   world: state.world,
   centroidLabel: state.centroidLabel,
 }))
@@ -43,7 +43,7 @@ class CategoryValue extends React.Component {
 
   handleMouseExit() {
     const { dispatch, metadataField, categoryIndex } = this.props;
-        dispatch({
+    dispatch({
       type: "category value mouse hover start",
       metadataField,
       categoryIndex
@@ -79,7 +79,7 @@ class CategoryValue extends React.Component {
     let occupancy = null;
 
     if (isColorBy && schema) {
-      categories = schema.annotations.obsByName[colorAccessor]?.categories;
+      categories = schema.annotations.obsByName[colorAccessor] ?.categories;
     }
 
     if (colorAccessor && !isColorBy && categoricalSelection[colorAccessor]) {
@@ -99,8 +99,8 @@ class CategoryValue extends React.Component {
           justifyContent: "space-between",
           borderRadius: "3px",
           background: centroidLabel.categoryIndex === categoryIndex &&
-                           centroidLabel.metadataField === metadataField
-                           ? "rgba(115, 134, 148, 0.3)":"inherit"
+            centroidLabel.metadataField === metadataField
+            ? "rgba(115, 134, 148, 0.3)" : "inherit"
         }}
         data-testclass="categorical-row"
         onMouseEnter={this.handleMouseEnter.bind(this)}
@@ -136,15 +136,15 @@ class CategoryValue extends React.Component {
           </label>
           <span style={{ flexShrink: 0 }}>
             {colorAccessor &&
-            !isColorBy &&
-            categoricalSelection[colorAccessor] ? (
-              <Occupancy
-                occupancy={occupancy.get(
-                  category.categoryValues[categoryIndex]
-                )}
-                {...this.props}
-              />
-            ) : null}
+              !isColorBy &&
+              categoricalSelection[colorAccessor] ? (
+                <Occupancy
+                  occupancy={occupancy.get(
+                    category.categoryValues[categoryIndex]
+                  )}
+                  {...this.props}
+                />
+              ) : null}
           </span>
         </div>
         <span>
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -10,7 +10,8 @@ import * as globals from "../../globals";
   colorScale: state.colors.scale,
   colorAccessor: state.colors.colorAccessor,
   schema: state.world?.schema,
-  world: state.world
+  world: state.world,
+  centroidLabel: state.centroidLabel,
 }))
 class CategoryValue extends React.Component {
   toggleOff() {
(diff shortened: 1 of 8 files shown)


client/src/components/graph/setupCentroidSVG.js
```diff
@@ -0,0 +1,33 @@
...
+    .attr("transform", `translate(${xy[0]}, ${xy[1]})`);
+
+  label
+    .append("text")
+    .attr("text-anchor", "middle")
+    .text(text)
+    .style("font-size", "25px")
+    .style("font-family", "Roboto Condensed")
+    .style("fill", "white")
+    .style("stroke", "black");
```
colinmegill:
```
    .style("font-size", "16px")
    .style("font-weight", "700")
```

client/src/util/centroid.js
```diff
@@ -0,0 +1,55 @@
+import quantile from "./quantile";
+
+/*
+  Centroid coordinate calculation
+
```
colinmegill: Whitespace

client/src/reducers/undoableConfig.js
```diff
@@ -29,7 +29,10 @@ const skipOnActions = new Set([
   "clear all user defined genes",
 
   "get single gene expression for coloring started",
-  "get single gene expression for coloring error"
+  "get single gene expression for coloring error",
+
```
colinmegill: Whitespace

colinmegill:
+ Run prettier on all 8 files

client/src/components/categorical/value.js
```diff
@@ -79,6 +97,8 @@ class CategoryValue extends React.Component {
           justifyContent: "space-between"
         }}
         data-testclass="categorical-row"
+        onMouseEnter={this.handleMouseEnter.bind(this)}
```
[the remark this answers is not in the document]
seve: Should I fix the preexisting binds for toggleOn and toggleOff?

client/src/reducers/undoableConfig.js
seve: I might be looking at the wrong this, but I believe the whitespace here is purposeful, to separate the actions of each reducer.

colinmegill: Roger. Feel free to run prettier and let it decide.

client/src/components/categorical/value.js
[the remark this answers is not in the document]
colinmegill: Sure

client/src/components/graph/setupCentroidSVG.js
seve: image

16px is a bit hard to read but I'm going to move forward with it

client/src/reducers/index.js
```diff
@@ -16,37 +16,39 @@ import layoutChoice from "./layoutChoice";
...
-        "categoricalSelection",
-        "continuousSelection",
-        "graphSelection",
-        "crossfilter",
-        "colors",
-        "controls",
-        "differential",
-        "layoutChoice"
-    ],
-    undoableConfig
```
csweaver: What actually changed here? Why so many lines of change?

client/src/util/centroid.js
```diff
@@ -0,0 +1,49 @@
+import quantile from "./quantile";
+
+/*
+  Centroid coordinate calculation
+*/
+const calcMeanCentroid = (world, annoName, annoValue, layoutDimNames) => {
+  const centroid = [0, 0, 0];
```
csweaver: I'd prefer to use an object `{ centroidX: 0, centroidY: 0, count: 0 }` or at least separate [x, y] and count.

client/src/util/centroid.js
```diff
@@ -0,0 +1,49 @@
...
+
+  for (let i = 0, len = annoArray.length; i < len; i += 1) {
+    if (annoArray[i] === annoValue) {
+      centroid[0] += layoutXArray[i];
+      centroid[1] += layoutYArray[i];
+      centroid[2] += 1;
+    }
+  }
+
+  if (centroid[2] !== 0) {
```
csweaver: Are you handling if the count is 0? This shouldn't happen, but it looks like if it did, then it would be in the 0,0 xy position.

client/src/util/centroid.js
```diff
@@ -0,0 +1,49 @@
...
+    if (annoArray[i] === annoValue) {
+      centroidX.push(layoutXArray[i]);
+      centroidY.push(layoutYArray[i]);
+    }
+  }
+
+  const medianX = quantile([0.5], Float64Array.from(centroidX));
+  const medianY = quantile([0.5], Float64Array.from(centroidY));
+
+  return [medianX, medianY];
```
csweaver: You are returning different results from your mean and median calculations [x,y] vs [x,y,count]. This should be consistent.  

I'd actually recommend fusing the calcCentroid functions and add another parameter `calcType='median'`. Makes alternate calculation methods more discoverable.

client/src/reducers/centroidLabel.js
```diff
@@ -0,0 +1,44 @@
...
+  const categoryField =
+    categoricalSelection?.[metadataField]?.categoryValues[categoryIndex];
+  switch (action.type) {
+    case "category value mouse hover start":
+      return {
+        ...state,
+        metadataField,
+        categoryIndex,
+        categoryField,
+        centroidXY: calcCentroid(
```
csweaver: It's cool that actually calling the calculation function is just a tiny bit of code. Great work!

client/src/components/graph/setupSVGandBrush.js
```diff
@@ -21,10 +21,12 @@ export default (
   const svg = d3
     .select("#graphAttachPoint")
     .append("svg")
+    .attr("id", "tool")
```
csweaver: @colinmegill going to defer to you on this, but tool seems like a pretty generic id, maybe graph-tools?

client/src/components/categorical/value.js
```diff
@@ -13,23 +13,41 @@ import * as globals from "../../globals";
...
     dispatch({
       type: "categorical metadata filter select",
       metadataField,
       categoryIndex
     });
-  }
+  };
+
+  handleMouseEnter = () => {
+    const { dispatch, metadataField, categoryIndex } = this.props;
```
csweaver: Oh this is odd, where are `metadataField` and `categoryIndex` getting set? I was expecting them to come in through the event. (Not a requested change, just for my own education)

client/src/util/centroid.js
seve: I only handled to prevent division by zero.  I didn't make a conscious choice on how to draw that centroid. How this is handled is probably a question for @sidneymbell and @colinmegill

client/src/util/centroid.js
seve: calcMeanCentroid uses an array because the function to convert an xy point to screen space asks for an array for its parameter.  Also for the moment calcMeanCentroid is unused.

client/src/util/centroid.js
sidneymbell: I believe that we pull the categories + labels (i.e., `annoValue`s) directly from the data. This means that if a label exists, then there is at least one cell that has that label.

The one place I could see this potentially getting into trouble would be if all of those cells are somehow assigned a nonfinite value (i.e., `NaN`, `Inf` or `-Inf`) for their x and/or y coordinates.
None of the layouts from scanpy should return non-finite coordinates, but if someone's passed a custom embedding, it could happen. 

@colinmegill / @csweaver  - how do we currently handle cells with non-finite embedding coordinates? 

The tl;dr -- it would be an extreme edge case, but if all of the cells with a given `annoValue` have non-finite values for their layout coordinates, this will currently implode. I'd suggest adding a quick conditional to check and just not showing a label in that case.

client/src/util/centroid.js
csweaver: You could do the calculations using an object and then transform to an array on return. 

`return [centroid.X, centroid.Y]`

client/src/util/centroid.js
seve: For the sake of getting this PR merged, I'm just going to remove the size from the return of calcMean, but I have plans on fusing the two together.

csweaver: You could make a clean up issue with additional plans for after the PR merge.

client/src/reducers/index.js
seve: A beautiful whitespace diff is the reason for the excess of lines.  Real changes are on lines 36 and 37

csweaver: Is that from prettier? Odd that it hadn't been fixed before.

client/src/reducers/centroidLabel.js
seve: 🎉

client/src/components/categorical/value.js
seve: The magic of redux.  I'll have to give you a quick run down.

client/src/components/graph/setupCentroidSVG.js
```diff
@@ -0,0 +1,33 @@
...
+import styles from "./graph.css";
+
+export default (responsive, graphPaddingRight, xy, text) => {
+  const containerWidth = responsive.width - graphPaddingRight;
+
+  const svg = d3
+    .select("#graphAttachPoint")
+    .append("svg")
+    .attr("id", "centroid")
+    .attr("data-testid", "centroid-overlay")
```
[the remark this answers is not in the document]
seve: 👍

client/src/util/centroid.js
seve: Made it so no centroids are drawn if there are no finite values in the category

PR #813, opened 2019-06-12
[EASY] fix max-category-items

### LINKED ISSUE #812: max-category-items flag not working
The max-category-items flag is supposed to limit the number of categories displayed, however due to a underscore (_) vs dash (-) naming conflict, this value never gets applied on the client.

csweaver:
Fixes #812 

The `--max-category-items` flag has never actually worked. On the back-end it was called "max_category_items" and on the front end it was "max-category-items" and since it was set using the spread operator, the user specified value, never overrode the client default. I renamed it on the backend for minimum code changes.

### DIFF
--- a/server/test/test_scanpy_engine.py
+++ b/server/test/test_scanpy_engine.py
@@ -78,8 +78,6 @@ class EngineTest(unittest.TestCase):
     def test_schema(self):
         with open(path.join(path.dirname(__file__), "schema.json")) as fh:
             schema = json.load(fh)
-            print(schema)
-            print(self.data.schema)
             self.assertEqual(self.data.schema, schema)
 
     def test_schema_produces_error(self):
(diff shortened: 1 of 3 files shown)


server/test/test_scanpy_engine.py
```diff
@@ -78,8 +78,6 @@ def test_obs_and_var_names(self):
     def test_schema(self):
         with open(path.join(path.dirname(__file__), "schema.json")) as fh:
             schema = json.load(fh)
-            print(schema)
```
csweaver: I snuck this in, because it was bugging me and I didn't want to make a whole new PR for removing two print statements.

seve:
> On the back-end it was called "max-category-items" and on the front end it was "max-category-items"

Typo? I'm assuming it meant to say 

> On the back-end it was called "max_category_items"

PR #816, opened 2019-06-15
GUI UI Elements

csweaver:
Fleshing out the load window for the GUI UI.
Added
- drag and drop for files
- loading indicator
- advanced options section
- logo
- window name

Screen Shot 2019-06-19 at 11 09 00 PM

There still is a fair amount of styling (CSS work) to be done, but this pr adds all the ui elements and hooks up their utility.


seve:
Right now the progress bar that appears when uploading the file seems like its communicating progress when it's not.  Consider replacing it with an indeterminate progress bar.

server/gui/main.py
```diff
@@ -237,13 +286,208 @@ def onError(self, err, server_error=False):
...
+        self.header_layout.addWidget(self.form_visibility_toggle)
+        self.header_layout.setStretch(0, 4)
+        self.options_layout.addLayout(self.header_layout)
+
+    def setupForm(self):
+        # Option Widgets
+        self.title_widget = QLineEdit()
+        self.title_widget.setAccessibleName("title")
+        self.embedding_widget = QComboBox(self)
+        self.embedding_widget.addItems([""] + MODES)
```
bkmartinjr: why are the layout names hard-coded as a constant?  The scanpy engine derives them from the dataset (see `_default_and_validate_layouts`) as they will vary by dataset.    Suggest that the same happen here.

server/gui/main.py
```diff
@@ -237,13 +286,208 @@ def onError(self, err, server_error=False):
...
+        self.embedding_widget.currentIndexChanged.connect(self.updateOpt)
+        self.obs_widget.textChanged.connect(self.updateOpt)
+        self.var_widget.textChanged.connect(self.updateOpt)
+        self.max_category_items_widget.valueChanged.connect(self.updateOpt)
+        self.diff_exp_lfc_cutoff_widget.valueChanged.connect(self.updateOpt)
+
+        # Labels and tooltips
+        title_label = QLabel("title:")
+        title_label.setToolTip("Title to display (if omitted will use file name)")
+        embedding_label = QLabel("embedding:")
```
bkmartinjr: We should be consistent and use either embedding or layout, but not both.  Suggest using the same as the CLI option (currently `layout`)

server/gui/main.py
```diff
@@ -237,13 +286,208 @@ def onError(self, err, server_error=False):
...
+        self.resetProgress()
         self.window().stacked_layout.setCurrentIndex(LOAD_INDEX)
         self.error_label.setText(f"Error: {err}")
         self.error_label.resize(self.MAX_CONTENT_WIDTH, self.error_label.height())
+        # self.error_label.setFrameStyle(QFrame.Panel)
 
     onServerError = partialmethod(onError, server_error=True)
 
 
+class AdvancedOptionsArea(QFrame):
```
bkmartinjr: As an alternative to this relatively complex property sheet, have you considered a simple text box that just accepts the CLI text parameters?    Advanced options are "advanced", and so there is less of a usability concern.  And it has the advantage of making all CLI params, now and future, available for use without a Qt implementation.

csweaver: Love that idea! Will re-implement

server/gui/main.py
csweaver: will fix

server/gui/main.py
csweaver: We don't have access to the file at form fill out time. I could do some pre-loading of the file and read from it to get the layouts. I think the point is moot, though, if I go with your CLI options suggestion as planned.

bkmartinjr: in that case, I'd either use my CLI options suggestion, or just remove it entirely...   It is only helpful as a way to reduce the amount of data copied to the front-end....   Otherwise it has little effect....

csweaver: Agreed!

csweaver:
Converted advanced options form to a text box that accepts the CLI text parameters. This did require some refactoring of the CLI so that I could consistently pull in common options as the cli gets updated but ignore options that wouldn't belong in the gui version (ex --open or --port). Click was pretty resistant to parsing args outside of launching your app from the CLI so I copied some of the parsing function directly from their core code.

Screen Shot 2019-06-26 at 11 40 53 PM

server/cli/launch.py
```diff
@@ -13,22 +14,53 @@
 from server.app.util.utils import custom_format_warning
 from server.utils.utils import find_available_port, is_port_available
 
-
 # anything bigger than this will generate a special message
-BIG_FILE_SIZE_THRESHOLD = 100 * 2**20  # 100MB
+BIG_FILE_SIZE_THRESHOLD = 100 * 2 ** 20  # 100MB
+
+
+def common_args(func):
```
bkmartinjr: function name doesn't mean much to me -- "common" to what?   Maybe just add a comment?

server/gui/options_parser.py
```diff
@@ -0,0 +1,24 @@
...
+from server.cli.launch import common_args
+
+
+@click.command()
+@common_args
+def cli(**kwargs):
+    pass
+
+
+def parse_opt_string(opts):
```
bkmartinjr: how much of this function is using private (click-internal) interfaces, vs. public API?   Just thinking about maintaining this in the long term....     Short term, could protect by adding some unit tests.

server/cli/launch.py
csweaver: common to gui/cli. will add comment

server/gui/options_parser.py
csweaver: `add_to_parser` (line 18) is the only truly internal function. Technically OptionParser is an internal class, but it's included in their documentation. 

I'll add some unit tests. Long term, click may be the wrong option if we want to keep this pattern with the gui. Reading some of the issues on click's github implies they are deliberately against adding the flexibility I'd need to do this the right way.

PR #822, opened 2019-06-24
increase readability of centroid labels

seve:
Increases the readability of the labels by removing the text label of the graph and replacing it with a target icon.  Also signifies the displayed label by a hover effect on the left sidebar.

* Styling changes to the `value` component
* Removal of the text label in `setupCentroidSVG` in favor of an SVG icon

_Note: This addresses, but does not fully close. #811_


csweaver:
closed, we will take this in another direction.

PR #827, opened 2019-06-26
categorical vs continuous mini histograms

### LINKED ISSUE #616: Visualize how continuous & categorical metadata values relate
The current app does a great job of showing how categorical metadata values relate to one another via "cluster occupancy". For example, I can see how `mouse.id` varies by `louvain` cluster assignment.

However, I frequently want to visualize how _continuous_ metadata values relate to categorical metadata values. For example, how does expression of `myfavoritegene` vary by louvain cluster assignment?

### LINKED ISSUE #762: support for violin plots
Frequently, the distribution of the expression of a gene wants to be compared across samples, tissues, cell types etc. (any categorical data). I would like to know if support for such visualizations would be possible for cellxgene. Particularly, violin plots to show gene expression across  categorical values are common.

Using scanpy, such a plot is produced as follows:
```PYTHON
# load sample data
adata = sc.datasets.pbmc68k_reduced()
sc.pl.violin(adata, 'CST3', groupby='bulk_labels', rotation=90)
```
image

My suggestion would be that if the user selects a gene for coloring, then, next to each of the categorical metadata an option would appear to show a violin plot for that category and the actual gene being colored.

seve:
This PR adds the ability to view continuous metadata in relation to categorical. This is done via mini histograms found on the side panel:
image

On hover of one of these mini histograms, a popup will display details about the specific histogram:
image

**@bkmartinjr**
* A histogram dataframe util which allows for creation of:
    * Categorical histograms
    * Categorical by histograms, which is used in occupancy
    * Continuous histograms, which can be used in the already implemented histograms
    * Continuous by histograms, which are use in the mini histograms
* Memoization of the dataframe util functions
* Tests for dataframe util functions

**@seve**
* Introduction of the mini histograms to the `occupancy` component
* Refactoring of occupancy to use canvas instead of SVG's to increase preformance
* Use of blueprint.js `Popover` for on hover action in occupancy
* Adding `shouldComponentUpdate` in the `value` component for preformance
* Move occupancy logic from `value` to the `occpuancy` component

This pr closes #616 and resolves #762


client/__tests__/util/dataframe/histogram.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+		const h1 = df.col("value").histogram(3, [0, 2]);
+		expect(h1).toMatchObject([1, 1, 1]);
+		// memoized?
+		expect(df.col("value").histogram(3, [0, 2])).toMatchObject(h1);
+	});
+
+	// XXX categorical test
+
+	/*
```
bkmartinjr: these comments look obsolete to me - ie, there are tests covering these items.  Please remove the comment if you agree.

client/src/components/categorical/value.js
```diff
@@ -22,6 +21,32 @@ class CategoryValue extends React.Component {
     });
   };
 
+  shouldComponentUpdate = nextProps => {
+    const { props } = this;
+    const { metadataField, categoryIndex, categoricalSelection } = props;
+    const { categoricalSelection: newCategoricalSelection } = nextProps;
+
+    const diff = Object.keys(this.props).reduce((value, key) => {
```
bkmartinjr: Why compute the difference of the entire props object, only to later check only for `world` and `colorAccessor` difference?   

Unless I am misreading the code (and apologies if I am), you could remove the computation of `diff` entirely (lines 29-35), and change the two line below (lines 45-46) to just read  `props.world !== nextProps.world || props.colorAccessor !== nextProps.colorAccessor`.

Seems simpler, clearer and less compute?

seve: Totally right, I originally computed the diff to see _what_ was different. Now that I know what I'm looking for I can just specifically look for that 👍

client/src/components/categorical/occupancy.js
```diff
@@ -2,60 +2,172 @@
...
+    const occupancyMap = world.obsAnnotations
+      .col(colorAccessor)
+      .histogram(groupBy);
+
+    const occupancy = occupancyMap.get(category.categoryValues[categoryIndex]);
 
     const x = d3
       .scaleLinear()
       /* get all the keys d[1] as an array, then find the sum */
       .domain([0, d3.sum(Array.from(occupancy, d => d[1]))])
```
bkmartinjr: small nit - you should iterate occupancy.values(), rather than the object itself.

You should be able to just do:   `Array.from(occupancy.values())` and skip the map function.

client/src/components/categorical/occupancy.js
```diff
@@ -2,60 +2,172 @@
...
-  render() {
-    const { occupancy, colorScale, colorAccessor, schema, world } = this.props;
-    const width = 100;
-    const height = 11;
+  _WIDTH = 100;
 
-    const categories = schema.annotations.obsByName[colorAccessor]?.categories;
+  _HEIGHT = 11;
+
+  createHistogram = () => {
```
bkmartinjr: general comment - all changes lack comments.  Please comment sufficiently that someone else can read and understand the code quickly (never having seen it before).

client/src/util/dataframe/histogram.js
```diff
@@ -0,0 +1,137 @@
...
+    return `${column.__id}:${by.__id}`;
+  }
+  return `${column.__id}:`;
+}
+
+/*
+Bin counts for continuous/scalar values, with optional group-by category.
+Values outside domain are ignored.
+*/
+export function histogramContinuous(column, bins = 40, domain = [0, 1], by) {
```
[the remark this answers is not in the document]
seve: @bkmartinjr how should we handle this?

client/src/components/categorical/occupancy.js
```diff
@@ -2,60 +2,172 @@
...
+      world.obsAnnotations.col(colorAccessor) ||
+      world.varData.col(colorAccessor);
+
+    const range = col.summarize();
+
+    const histogramMap = col.histogram(
+      50,
+      [range.min, range.max],
+      groupBy
+    ); /* Because the signature changes we really need different names for histogram to differentiate signatures  */
```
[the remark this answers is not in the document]
seve: > client/src/util/dataframe/dataframe.js, line 256:

`      get.histogram = (bins, domain, by) => mFn(get, bins, domain, by);
`
and 
`
      get.histogram = by => mFn(get, by);
`

seve:
@bkmartinjr fixed your requested changes, how should I handle charlotte's requests dealing with the histogram util?

client/src/util/dataframe/histogram.js
```diff
@@ -0,0 +1,137 @@
...
+  return by
+    ? _histogramContinuousBy(column, bins, min, max, by)
+    : _histogramContinuous(column, bins, min, max);
+}
+
+/*
+Memoization hash for histogramContinuous
+*/
+export function hashContinuous(column, bins, domain, by) {
+  bins = bins || "";
```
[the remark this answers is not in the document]
seve: the return of hashContinous is a string hash, bins is defaulted to a string as it will be casted to a string anyways.

seve:
@sidneymbell plugging you in as a reviewer to check out the language in the popover:

>  The x axis scale for all mini histograms are the same as the active
            color scale: `colorAccessor`
            The y axis scale for each mini histogram is normalized to its
            largest value

colinmegill:
@sidneymbell specifically, do you have a better suggestion for `mini histogram` that is more specific?

client/src/util/dataframe/histogram.js
```diff
@@ -0,0 +1,137 @@
...
+  domain = domain || [0, 0];
```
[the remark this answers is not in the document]
bkmartinjr: agree - @seve can you make this change before you merge?

client/src/util/dataframe/histogram.js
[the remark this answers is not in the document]
bkmartinjr: I would leave as-is.

client/src/util/dataframe/histogram.js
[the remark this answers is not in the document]
bkmartinjr: @seve - I would also make the change to the domain argument.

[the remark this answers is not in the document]
csweaver: don't forget about domains too

client/src/components/categorical/occupancy.js
```diff
@@ -2,60 +2,182 @@
...
+          width={this._WIDTH}
+          height={this._HEIGHT}
+          ref={ref => {
+            this.canvas = ref;
+            if (colorByIsCatagoricalData) this.createOccupancyStack();
+            else this.createHistogram();
+          }}
+        />
+        <div key="text" style={{ fontFamily: "Roboto", fontSize: "14px" }}>
+          <p>
```
sidneymbell: Love the explanatory popovers! Axes = <3

Suggested wording:
"These histograms show the distribution of <strong>{colorAccessor}</strong> within each category. The x axis is the same for each histogram, while the y axis is scaled is scaled to the highest bin within each histogram."

I might also suggest a "Ok, I've got it" button after the text that disables subsequent popovers.

seve: image

sidneymbell: Nit: new line between first and second sentences?

liaprins-czi:
From Alok Saldanha:
"I need to follow up with our users, but they may want to toggle between raw and normalized values in the histograms. I think they are happy to have differential analysis, etc happen on the X array, so it’s probably not ideal to switch at launch time."

PR #833, opened 2019-07-01
disable centroids

seve:
Disable centroid for a future release.

Changes:

* Change redux action type to have: `---DISABLED---` implicitly not changing the redux store


bkmartinjr:
I agree with @csweaver - I would not fire actions that are not being used....

that said, if this is a temporary thing and won't be released, I'm not hugely concerned.

seve:
>  if this is a temporary thing and won't be released, I'm not hugely concerned.

This removal is in preparation for 0.11.0 so I'll remove the non-functional reactions

PR #839, opened 2019-07-08
remove options widget + launch on file selection

csweaver:
Remove the ability to set options from gui until we can remove click as a dependancy. By removing the options I'm able to allow launch to happen directly on file selection instead of the two steps of selecting a file and then clicking the launch button. 

I've made the choice to keep some of the dead code for options in since I expect to replace click someone soon.


csweaver:
Screen Shot 2019-07-08 at 4 51 03 PM

PR #840, opened 2019-07-09
Installer

csweaver:
Using pyinstaller to build the native app for windows and mac. Added make scripts, hooks, and spec files to support that. Had to make a few minor changes to ensure that it works. 
1) pkg_resources doesn't work in native mode, needed to add __version__ attribute to the cellxgene package. 
2) multiprocessing needed to add freeze_support so that it would work in a built package.

Currently creating a folder with an executable. Need to do a bit more investigation to figure out how to get cefpython to work with the single file version.


bkmartinjr:
@csweaver - this looks great as an interim merge.   However, I am wondering why it requires a version bump?

csweaver:
It's actually not doing a version bump, it's just a new location for the current version. The `pkg_resources` library doesn't play well with pyinstaller and so I needed to introduce a new way to get the current version for the `/config` endpoint.

bkmartinjr:
> It's actually not doing a version bump, it's just a new location for the current version. 

Sorry, that was my mistake/misreading of the code.  TY!

PR #841, opened 2019-07-09
add point dilatation on hover

### LINKED ISSUE #811: Further communicate the cell population a centroid represents
Currently centroids are only displaying the categorical value the centroid represents as a text label.  The app may further benefit from additional representations of the categorical value.

Final implementation:
ezgif com-video-to-gif (2)

Potential methods include:

- Popping the cells forward on the z-axis
- Draw an encompassing circle around the cells
   - A concave hull
   - Ellipse with width/height showing max/min x and y
   - Radius displaying the value's population
- Grow the represented cells
- Shrink the non-represented cells
- Indicate the displayed value on the sidebar via styling

Screenshots showing a circle label based on population with sidebar styling:
Screen Shot 2019-06-11 at 3 16 36 PM
Screen Shot 2019-06-11 at 3 16 27 PM

seve:
This PR slightly re-enables centroid labels, effectively undoing #833 for #809.  It also features graph point dilation on hover of category fields.  The state of centroid labels is that they are being computed but not shown.  This is because the centroid logic feeds into the point dilation.

ezgif com-video-to-gif (2)
----
- In the `value` component
  - Removed the comments added in #809 to enable #833
  -  Add styling to aid hover effect
  - Inverse hover actions on checkbox (hover on resets centroid state, hover off adds it)
- In `categorical.css`
  - Add hover psuedoelement styling
- In the `graph` component
  - Add functionality to `computePointSizes()` to size by hovered category field in addition to selection
 - Remove centroid labels
---
This PR will close #811

### DIFF
--- a/client/src/components/graph/graph.js
+++ b/client/src/components/graph/graph.js
@@ -298,15 +298,13 @@ class Graph extends React.Component {
       stateChanges = { ...stateChanges, centroidSVG: newCentroidSVG };
     };
 
-    // SVG creation is disabled for now
-
+    // Centroid SVG creation is disabled for now but should go into the first and third cases if enabled
     if (
       prevProps.responsive.height !== responsive.height ||
       prevProps.responsive.width !== responsive.width
     ) {
       // If the window size has changed we want to recreate all SVGs
       createToolSVG();
-      // createCentroidSVG();
     } else if (
       (responsive.height && responsive.width && !toolSVG) ||
       selectionTool !== prevProps.selectionTool ||
@@ -319,7 +317,6 @@ class Graph extends React.Component {
       (responsive.height && responsive.width && !centroidSVG)
     ) {
       // First time for centroid or label change
-      // createCentroidSVG();
     }
 
     /*
--- a/client/src/components/graph/graph.js
+++ b/client/src/components/graph/graph.js
@@ -52,30 +52,27 @@ class Graph extends React.Component {
     return colors;
   });
 
-  computePointSizes = memoize((len, crossfilter) => {
-    /*
+  computePointSizes = memoize(
+    (len, crossfilter, metadataField, categoryField) => {
+      /*
     compute webgl dot size for each point
     */
-    console.log("call");
 
-    const centroidLabel = null;
-    const sizes = new Float32Array(len);
-    crossfilter.fillByIsSelected(sizes, 4, 0.2);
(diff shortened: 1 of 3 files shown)


client/src/components/graph/graph.js
```diff
@@ -279,13 +298,15 @@ class Graph extends React.Component {
...
+    // SVG creation is disabled for now
+
     if (
       prevProps.responsive.height !== responsive.height ||
       prevProps.responsive.width !== responsive.width
     ) {
       // If the window size has changed we want to recreate all SVGs
       createToolSVG();
-      createCentroidSVG();
+      // createCentroidSVG();
```
bkmartinjr: please remove unused code

client/src/components/graph/graph.js
```diff
@@ -298,7 +319,7 @@ class Graph extends React.Component {
       (responsive.height && responsive.width && !centroidSVG)
     ) {
       // First time for centroid or label change
-      createCentroidSVG();
+      // createCentroidSVG();
```
bkmartinjr: please remove unused code

bkmartinjr:
@seve - can you confirm you tested this on 1M+ cell data sets (eg, the 10X neuron data) and it performs adequately?

seve:
@bkmartinjr 10x_mouse_13MM doesn't have any metadata attached other than louvain, but running with that, there's no noticeable performance difference.

client/src/components/graph/graph.js
seve: Was leaving this in to allow for a quick reinstantiation of the feature. Removed the comments and added a clarifying comment above

sidneymbell:
This is awesome progress!! 

I see the appeal of the text labels for categories. But, for more complex metadata, I'm still concerned that they can be misleading. 

In this example, while I think the point dilation totally helps the user figure out which cells are in the tissue category "Fat", the difference is still kind of subtle; paired with the label landing over an unassociated cluster, I'm worried it could still confused people.
image

What do you think about perhaps:
1) Simultaneously reducing the size of the other cells
2) Removing the text labels (until a toggle is added to manually turn them on)

The PR itself looks great. Thanks for your great work on this!

seve:
Hey @sidneymbell I think your branch might be outdated. I dropped the labels after Colin and yours feedback :)

sidneymbell:
@seve - This makes sense, thanks!

PR #843, opened 2019-07-11
reset colorAccessor and colorMode if colored diffexp gene is removed

seve:
this pr adds the functionality of clearing `colorAccessor` and `colorMode` within the sharedState `color`

Behavior before merge:
1) User creates a diffexp 
1) colors by a gene exposed by diffexp
1) mini histograms appear and graph points are colored
1) user clears diffexp or makes selection world
1)  __colorScale still exists and is not reset__

Behavior after merge:
           5.  __colorScale is reset__

* In the `colorHelpers` stateManger util:
  * Switched `createColors()` from default to named export
  * Created `resetColors` to reset state values to defaults
  * Created `checkIfColorByDiffexpAndResetColors` to check if colorAccessor is set to a diffexp gene and reset state values conditonaly
* In the stateManager util `index`:
   * Edited import/export to reflect default->named export change
* In the `colors` reducer:
   *  Various changes to reflect util changes
   * Utilize `checkIfColorByDiffexpAndResetColors` inside of `set World to current selection` action to reset colors if needed
   * Utilize `resetColors` in `reset colorscale` action
   * Create `clear differential expression` action to reset colors if needed

### DIFF
--- a/client/src/reducers/colors.js
+++ b/client/src/reducers/colors.js
@@ -1,16 +1,4 @@
-import { createColors } from "../util/stateManager";
-
-const resetColors = (prevSharedState, state) => {
-  const { world } = prevSharedState;
-  const { rgb, scale } = createColors(world);
-  return {
-    ...state,
-    colorMode: null,
-    colorAccessor: null,
-    rgb,
-    scale
-  };
-};
+import { ColorHelpers } from "../util/stateManager";
 
 const ColorsReducer = (
   state = {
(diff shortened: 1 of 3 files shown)


client/src/reducers/colors.js
```diff
@@ -1,4 +1,4 @@
-import { createColors } from "../util/stateManager";
+import { ColorHelpers } from "../util/stateManager";
```
bkmartinjr: Optional style suggestion, used in other reducers to make code cleaner: import with a short alias to make code below easier to read.

Ie, `import { ColorHelpers as CH } from ...`

and then `CH.createColors(...)`

See (for example) `reducers/categoricalSelection.sj`

seve: I'm going to actually not go with the abbreviation. I also couldn't find an example in the file you mentioned.  The comprehension cost outweighs the brevity imo

Issue #892, opened 2019-08-21
scatterplot does not show categorical highlighting state


bkmartinjr:
the graph has a category metadata highlighting, triggered by hover-over the category label.  Expression scatterplot does not also signal this, but it should.

Issue #759, opened 2019-05-09
Scaling the dot size based on number of points


gokceneraslan:
It'd be really nice to decrease the dotsize when visualizing large datasets and increase it for smaller ones.

We have a simple scaling in scanpy (https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_anndata.py#L323) which is `120000 / n` where n is the number of cells. This leads to very big dots when dataset is very small, so you might want to limit it via min() or max().

gokceneraslan:
(Discussed with @colinmegill )

colinmegill:
@bkmartinjr @liaprins-czi this will need to take into consideration all three dot sizes: 

* default
* deselected
* hover

colinmegill:
First thought - we might hold `deselected` size constant, at the size it is now, and scale `default` and `hover` together, with a min and a max for both TBD

liaprins-czi:
Yeah I think that ^^^ would be good. It'd be confusing to have the deselected size scale up for a small dataset and end up the same size as a default dot of a large dataset.

bkmartinjr:
Should the scaling be based upon number of points, canvas/window size, zoom or some combination?

Right now, scaling occurs based upon canvas/window size, but not the others (assuming that the canvas has a fixed [-1,1] coordinate space.

liaprins-czi:
I created a Figma board to start loosely prototyping what each of the possible scaling methods would look like that @bkmartinjr mentioned above:
- number of points
- canvas/window size
- zoom level

https://www.figma.com/file/dY5jElMTaYU2XqQLBhi1Et/Scale-dot-size?node-id=0%3A1

I haven't tried combining methods yet. I just wanted to have visuals for how each method might behave on its own, to start with.

liaprins-czi:
Update: Now I have tried out really rough conceptual mockups (in the Figma file linked to above) of "dot density" (number of dots / window size*). This way we account for both at once, which I think makes sense. We can also still apply zoom scaling (or not apply zoom scaling, see mocks of both in the Figma) on top of the dot density approach.

Assuming we wanted to use dot density as the scaling method, we'd then just need to decide:

1. Allow zooming to scale dots or just scale vis as a whole (and maintain size of dots)? There are pros and cons to either.
[...]

PR #893, opened 2019-08-21
Graph/scatterplot improvements

### LINKED ISSUE #722: main embedding pan does not have 1:1 drag with cursor

### LINKED ISSUE #758: center zoom on mouse location
When zooming in, I expect the zoom to be centered on the pointer location (like in Google maps)

**What I did**

1. Switch to zoom/pan mode
2. Move pointer over an "outer" cluster
3. Scroll up to zoom in

**What I expected**

1. Layout would zoom in to let me view the cluster I've hovered over

**What happened**

1. Layout zooms in on center of layout, pushing cluster out of the view

### LINKED ISSUE #882: RAF loop always active in graph pan/zoom mode
The graph component RAF is always active when in pan/zoom mode, which leads to excessive cpu/power consumption (a problem for some usage scenarios, eg mobile/tablet).  We should activate only when required, or move to an explicit repaint callback.

### LINKED ISSUE #886: scatterplot needs refactoring/cleanup
the scatterplot component looks like it was cut&pasted from graph, and it has a lot of unnecessary boilerplate code (eg, a 3d perspective projection, a 3d camera, etc).   Needs cleanup/refactoring along the lines of the graph cleanup

bkmartinjr:
This PR includes a number of enhancements and refactors of both the main graph and scatterplot widgets:
* graph pan now is synchronized with mouse movement. Fixes #722 
* graph zoom centered on mouse location.  Fixes #758 
* graph RAF no longer runs infinite loop, which will save battery/CPU.  Fixes #882 
* significant cleanup/refactoring of both graph and scatterplot rendering. In particular, both move to a 2D model, using a set of simple 2D transforms to control camera & projection.  Fixes #886 
* for both plots, points now stack:  highlighted on top, NaNs/clipped on bottom, all else in middle.
* for both plots, NaN/clipped data will be alpha 0.9, all else 1, improving visibility of non-NaN data.
* a large number of (now) unnecessary JS module dependencies removed (eg, the 3D orbit-camera)

Still to come as future PRs built on top of this, but *not* included in this PR:
* dot rendering improvements, including #759, etc.
* showing hover-over state in scatterplot, #892 
* constraining pan so you can't pan all of the graph off-canvas


liaprins-czi:
This all looks good to me!

PR #897, opened 2019-08-22
merge louvain and prepare extra requires

csweaver:
So one opportunity I saw when separating the install of cellxgene prepare was the ability to merge the special install of the louvain package (currently separated from main install because of license incompatibility) with the special install of the prepare package. Installing `cellxgene[prepare]` should include everything you need to prepare your data.

The problem I'm trying to solve with this is that to get clusters when preparing your data currently you would have to run two special installs
```
pip install cellxgene[prepare]
pip install cellxgene[louvain]
```
OR
```
pip install cellxgene[prepare,louvain]
```

Instead we can merge the two so that a user just has to install
```
pip install cellxgene[prepare]
```

We have moved both to extra requires for different reasons but since they are part of the same command, IMHO they should be installed together.

### DIFF
--- a/server/cli/prepare.py
+++ b/server/cli/prepare.py
@@ -160,14 +160,7 @@ def prepare(
         sc.pp.neighbors(adata)
 
     def run_louvain(adata):
-        try:
-            sc.tl.louvain(adata)
-        except ModuleNotFoundError:
-            click.echo(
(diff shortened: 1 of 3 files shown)


sidneymbell:
For the group of users who I suspect use `prepare`, having an easy installation process + being able to look at cluster labels are both pretty key. 

LGTM!

PR #910, opened 2019-09-06
Fix incorrect gene add error on empty query

### LINKED ISSUE #865: Wait to surface "No matching genes" warning until at least one character is typed
...and if it’s a letter character, there probably _will_ be a gene found.

shouldn't scold me when first input cursor

seve:
This PR will remove the false error given when beginning an add gene query.

image

---

closes #865

### DIFF
--- a/client/src/components/geneExpression/index.js
+++ b/client/src/components/geneExpression/index.js
@@ -75,19 +75,25 @@ class GeneExpression extends React.Component {
     };
   }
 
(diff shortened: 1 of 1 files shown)


bkmartinjr:
> Just curious where we landed on this one... was @colinmegill able to figure out how to entirely disable the flyout from appearing (until there is at least one char typed)?

It seems like a great idea to completely disable if the user has not provided input.

PR #911, opened 2019-09-06
fix long values in sidebar

### LINKED ISSUE #891: color-by mode switch changes left side bar layout unecessarily
If I change the color-by mode (of a gene), it changes the layout of categorical items, causing the left side bar vertical size to change.  This bouncing of size (bigger, smaller, bigger, ...) is super distracting.

root cause seems to be the layout of the continuous occupancy histogram in the categorical items.

without color-by:

clicking color-by on the second gene:

Note the wrapping of text in the louvain category "FCGR3A+ Monocytes".  This causes the entire column to be bigger. 

Ideally, the vertical height of the side bar would not change due to view modes like color-by or what items are selected.

My suggestion is that we clip / truncate the category label names in some way, and perhaps use a tooltip to show their entire name if clipped.

seve:
This PR fixes the issue with longer values wrapping in the sidebar and creating additional lines.  This issue was increased when coloring and adding an occupancy/mini-histogram in the value space.

This is solved by truncating value names by adding ellipses in the center of the value should it be too long.  Full value names are still accessible via tooltip.

image

---

Fixes #891


seve:
@colinmegill @liaprins-czi Not too sure about the tooltip.  I currently have the default delay, I think it could be increased to 1s?

liaprins-czi:
@seve I know Bruce had just added tooltips to the buttons in the top action bar, so can use the same delay they have?

seve:
I decreased the character threshold for truncating to add more padding between the value name and the sparkline.  Also used the global hover delay

seve:
@colinmegill @liaprins-czi bump for reviews

client/src/components/categorical/value.js
```diff
@@ -105,6 +106,22 @@ class CategoryValue extends React.Component {
       categories = schema.annotations.obsByName[colorAccessor]?.categories;
     }
 
+    const longLength = 35;
```
colinmegill: These should be more descriptively named and moved to globals: 

`globals.categoryLabelDisplayStringMaxLength` or something similar

colinmegill:
Locally, I'm seeing baseline misalignment for values: 

image

colinmegill:
Also, I would shorten the tooltip delay — too long as is.

seve:
@colinmegill I was able to merge master into this without the extraneous divs, leaving the `- 130` instead of `- 160`.  From my light testing, it also fixes #927

seve:
I was reading the conversation in #927 and I may have misunderstood the reasoning behind the change from 160 to 130.  Let me know if you want me to switch that back.

PR #919, opened 2019-09-11
Fix bugs surrounding gene adding

### LINKED ISSUE #866: “Bulk add genes” should recognize lowercase inputs
“Bulk add genes” doesn’t recognize lowercase, but “Autosuggest genes” does. We shoul make it consistent, and in favor of the more inclusive experience: “Bulk add genes” should recognize lowercase inputs.
Bulk add genes  doesn't recognize lowercase inputs, but  Autosuggest genes  does

seve:
This PR fixes the following bugs:
* The add button not properly working on `Autosuggest gene`
* `bulk add genes` working case-sensitively

---
Changes:
* `geneExpression` component
   * Add onClick listener to `Add gene` button
   * Refactored `handleBulkAddClick()` to be case-insensitive 

---
closes #866

### DIFF
--- a/client/src/components/geneExpression/index.js
+++ b/client/src/components/geneExpression/index.js
@@ -75,19 +75,25 @@ class GeneExpression extends React.Component {
     };
   }
 
-  _genesToUpper = listGenes => {
-    const upperGenes = new Array(listGenes.length);
-    for (let i = 0, { length } = listGenes; i < length; i += 1) {
-      upperGenes[i] = listGenes[i].toUpperCase();
+  _genesToUpper = col => {
+    // Has to be a Map to preserve index
+    let listGenes = col;
+
+    // Extract the array from column if applicable
+    if (typeof col.asArray === "function") {
+      listGenes = col.asArray();
     }
+
+    const upperGenes = new Map();
+    for (let i = 0, { length } = listGenes; i < length; i += 1) {
+      upperGenes.set(listGenes[i].toUpperCase(), i);
+    }
+
     return upperGenes;
   };
 
   // eslint-disable-next-line react/sort-comp
-  _memoGenesToUpper = memoize(
-    this._genesToUpper,
-    listGenes => listGenes.toString() /* This hash is order specific*/
-  );
+  _memoGenesToUpper = memoize(this._genesToUpper, col => col.__id);
 
   placeholderGeneNames() {
     /*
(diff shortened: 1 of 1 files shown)


client/src/components/geneExpression/index.js
```diff
@@ -108,7 +130,7 @@ class GeneExpression extends React.Component {
   handleClick(g) {
     const { world, dispatch, userDefinedGenes } = this.props;
     const varIndexName = world.schema.annotations.var.index;
-    const gene = g.target;
+    const gene = g?.target;
```
bkmartinjr: this would be much better replaced with a more expliict action, such as:

```
if (!g) return;
```

client/src/components/geneExpression/index.js
```diff
@@ -138,27 +160,43 @@ class GeneExpression extends React.Component {
...
-          }
-          return dispatch(actions.requestUserDefinedGene(gene));
-        })
-      ).then(
+
+      const promises = [];
+      for (let i = 0, { length } = upperGenes; i < length; i += 1) {
+        const upperGene = upperGenes[i];
+
+        if (genes.length === 0) {
```
bkmartinjr: this should not be in the loop - it is a constant every time.   But by putting it in the loop, this necessary check is omitted.   Move outside the loop, eg, immeiately after the `genes` variable is set (line 165).

client/src/components/geneExpression/index.js
```diff
@@ -138,27 +160,43 @@ class GeneExpression extends React.Component {
...
+          /* Because of _.pull() this case is unreachable */
+          return keepAroundErrorToast("Must enter a gene name.");
+        }
+        if (upperUserDefinedGenes.indexOf(upperGene) !== -1) {
+          return keepAroundErrorToast("That gene already exists");
+        }
+
+        const indexOfGene = upperWorldGenes.get(upperGene);
+
+        if (indexOfGene) {
```
bkmartinjr: this logic is inverted - don't you mean `if (!indexOfGene)`?

client/src/components/geneExpression/index.js
```diff
@@ -73,6 +75,26 @@ class GeneExpression extends React.Component {
     };
   }
 
+  _genesToUpper = col => {
+    // Has to be a Map to preserve index
+    let listGenes = col;
+
+    // Extract the array from column if applicable
+    if (typeof col.asArray === "function") {
```
bkmartinjr: recommend deleting this polymorphic arg support, and just make this function _require_ an array as an argument.  Much simpler conceptually.

client/src/components/geneExpression/index.js
```diff
@@ -138,27 +160,43 @@ class GeneExpression extends React.Component {
...
-            return keepAroundErrorToast(
-              `${gene} doesn't appear to be a valid gene name.`
-            );
-          }
-          return dispatch(actions.requestUserDefinedGene(gene));
-        })
-      ).then(
+
+      const promises = [];
+      for (let i = 0, { length } = upperGenes; i < length; i += 1) {
```
bkmartinjr: this is not as clear as the original code which used `.map()`.   Why not simply do `upperGenes.map()`?

client/src/components/geneExpression/index.js
```diff
@@ -234,6 +272,7 @@ class GeneExpression extends React.Component {
                   return "";
                 }}
                 itemListPredicate={filterGenes}
+                onActiveItemChange={item => this.setState({ activeItem: item })}
```
bkmartinjr: best practice to set default state

client/src/components/geneExpression/index.js
seve: upperGenes is a Map so had to pull an array out from keys using [...map.keys()]

PR #921, opened 2019-09-13
Rename "layout" to "embedding" in the UI

### LINKED ISSUE #824: --layout CLI param help confusing
the help text for `--layout` is confusing - it should clearly state that not specifying a layout will make *all* embeddings in the dataset available, and that specifying layout names will restrict options to those specified.

Changes to be made:
* [x] CLI parameter --layout modified to accept N embedding names (rather than the current restriction of one).  Any specified will present in the web UI.   Default, if not specified, is to present all embeddings.
* [x] CLI parameter --layout changes to --embedding. Help and error text in the CLI change corresponding.
* [x] UI text changes from layout to embedding (eg, in the menu used to select an embedding)

bkmartinjr:
Fixes #824 

In the CLI and web UI, rename "layout" to "embedding".


bkmartinjr:
Good catch @liaprins-czi - just made the suggested change per our f2f conversation.

PR #928, opened 2019-09-18
Add obs index to label file 

bkmartinjr:
this adds the obs (row) index to the label file:
- the `obs.index` is written to the file as the index
- upon load, column zero is presumed to be the index.  Additionally, load will fail if the index column values are not unique, and if they do not precisely match the obs.index (using `DataFrame.index.equals()`, which checks for both values and order)

In addition, removed a bunch of debug logging from the annotations reducer.


sidneymbell:
You do have some lint:
>server/app/scanpy_engine/scanpy_engine.py:349:121: E501 line too long (125 > 120 characters)

PR #943, opened 2019-09-21
Add support for anndata backed mode

### LINKED ISSUE #914: Option to load anndata in backed mode
Use case: I have a large anndata h5ad file, that I want to open in cellxgene to view the expression of genes of interest on a 2d embedding, but am unable to do so because cellxgene loads the entire matrix into memory

bkmartinjr:
This PR adds two functional changes:
* a new CLI param `--backed` added to the launch command, causing `cellxgene launch` to load the H5AD using the anndata backed mode
* a new CLI param `--disable-diffexp` which will disable the differential expression UI on the front-end.

#### Backed
 When specified, the H5AD will be opened in file backed mode (using the AnnData option `backed="r"` documented [here](https://anndata.readthedocs.io/en/stable/anndata.read_h5ad.html#anndata.read_h5ad).  This results in the application using less memory and launching/initializing more quickly.  However, various data access operations, most notably reading gene expression data and computing differential expression calculations, will be significantly slower. 

Usage:  `cellxgene launch mydata.h5ad --backed`

The default behavior remains non-backed (fully loaded into memory) mode.  The default mode is recommended for datasets that fit into main memory.  If your dataset is larger than main memory, try `--backed`, but note that computing differential expression is likely to be to slow for interactive use.

Fixes #914 

#### Disable Diffexp

Differential expression requires access to large portions of the X matrix to compute "top N" genes.  For large datasets, this computation can be very slow, and may cause the user's computer to "thrash" if memory / disk IO limits are hit.  Performance may be further affected by other parameters, including X encoding (eg, sparse), computer CPU/RAM/disk and the use of `--backed` mode.

For some deployment scenarios, it is desirable to disable on-demand differential expression calculation.  A new CLI option will disable this feature in the application.  

Usage: `cellxgene launch mydata.h5ad --disable-diffexp`

Diffexp is enabled by default.

#### Other changes in this PR:
* Added heuristic to warn user if diffexp likely to be slow
* --debug, when specified, will disable flask multithreading (allowing use of pdb)
* Introduced a new proxy class for `X` data matrices, including tests for this class
* A special path for processing sparse array indexing is removed, as the scipy sparse indexing was improved dramatically in the 1.3 release.  Updated module dependencies to require scipy>=1.3

Performance note:  as of anndata release 0.6.22-post1, sparse backed data access to X (expression data) is very slow.  Accessing any subset of X will read the entire dataset (for each and every access to X[]).  This is a known issue and will be fixed in a future anndata release.


bkmartinjr:
Re: disabling diff exp:  I'm not sure why restart is any better or worse than disabling, as both require a restart if you need diffexp and you started in --backed mode.   The documentation (eg, help text) makes it very clear what to expect.   The most likely scenario (I think) is users who don't read the docs, and get confused by the interminable compute time.

Also to consider:
* it will likely get somewhat better with the next release of anndata.  But it will never be "as good" as memory cached.   
* it is fine for smaller datasets (hard to say exactly where the cut-off is, but it definitely works fine for small to medium data sets and fast disk.
* even in memory backed mode, it can still be very slow if the submatrix size is big

Options I see:
- leave as is, perhaps with better docs
- add a UI "notice", which informs the user that it will be slow when they run diffexp (eg, a confirmation pop-up or something like that)
- disable, as you suggest
- or, more complex: at startup, or perhaps during prepare, pre-compute data structures that will accelerate diffexp (some of these take extra disk space, but they will dramatically speed it up)

I'll put on the agenda for our next stand-up

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -251,6 +263,10 @@ def _validate_and_initialize(self):
         self._validate_label_data()
         self._create_schema()
 
+        # heuristic
+        if (self.data.shape[0] * self.data.shape[1]) > (100 * 10 ** 6):
```
sidneymbell:
```suggestion
        n_values = self.data.shape[0] * self.data.shape[1]
        if (n_values > 1e8 and self.config['backed']==True) or (n_values > 5e8):
```

sidneymbell: Still a hacky heuristic. But, I found that tmuris (n_values ~= 3e8) was fine in not-backed mode and slow in backed mode. 1.3M dataset (n_values ~= 1e9) was fatally slow in both cases.

bkmartinjr: fine with me - I'll make this change.

bkmartinjr:
@sidneymbell - the three requested changes are pushed.

PR #962, opened 2019-10-01
remove modal for inline edit category name, and autofocus

### LINKED ISSUE #930: Annotations: Inconsistency in method for editing main label name vs sub-label name
Editing the main label’s name pops up a modal with a blank field, but editing the sub-label keeps the original name intact, but allows user to edit it inline.

Recommend: make both editing methods consistent. I prefer the sub-label (inline) method as it’s less heavy-handed than a modal and also keeps the original name intact as a starting point (e.g. maybe you just wanted to fix a typo in the original name… this happens to me all the time).

colinmegill:
Fixes #930

### DIFF
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -237,8 +237,6 @@ class Category extends React.Component {
               <span className="bp3-control-indicator" />
               {""}
             </label>
-            {/* Dialog uses portal, can be anywhere/factored out */}
-
             <span
               data-testid={`category-expand-${metadataField}`}
               style={{
@@ -248,7 +246,7 @@ class Category extends React.Component {
               onClick={() => {
                 const editingCategory =
                   annotations.isEditingCategoryName &&
-                  annotations.categoryEditable === metadataField;
+                  annotations.categoryBeingEdited === metadataField;
                 if (!editingCategory) {
                   this.setState({ isExpanded: !isExpanded });
                 }
@@ -259,7 +257,7 @@ class Category extends React.Component {
               ) : null}
 
               {annotations.isEditingCategoryName &&
-              annotations.categoryEditable === metadataField ? (
+              annotations.categoryBeingEdited === metadataField ? (
                 <form
                   style={{ display: "inline-block" }}
                   onSubmit={e => {
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -238,44 +238,7 @@ class Category extends React.Component {
               {""}
             </label>
             {/* Dialog uses portal, can be anywhere/factored out */}
-            <Dialog
-              icon="tag"
-              title="Edit category name"
-              isOpen={
-                annotations.isEditingCategoryName &&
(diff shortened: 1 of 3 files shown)


client/src/components/categorical/category.js
```diff
@@ -238,58 +238,66 @@ class Category extends React.Component {
               {""}
             </label>
             {/* Dialog uses portal, can be anywhere/factored out */}
```
bkmartinjr: What does this comment mean?

client/src/components/categorical/category.js
```diff
@@ -238,58 +238,66 @@ class Category extends React.Component {
...
+                }
               }}
             >
               {isUserAnno ? (
                 <Icon style={{ marginRight: 5 }} icon="tag" iconSize={16} />
               ) : null}
-              {metadataField}
+
+              {annotations.isEditingCategoryName &&
+              annotations.categoryEditable === metadataField ? (
```
bkmartinjr: the name `categoryEditable` is confusing -- if I read the code right, it contains the name of the category currently being edited, ie, the category in "editable mode".    Consider renaming to `.categoryBeingEdited` or some such.

client/src/components/categorical/category.js
```diff
@@ -238,58 +238,66 @@ class Category extends React.Component {
...
+
             <span
               data-testid={`category-expand-${metadataField}`}
               style={{
                 cursor: "pointer",
                 display: "inline-block"
               }}
               onClick={() => {
-                this.setState({ isExpanded: !isExpanded });
+                const editingCategory =
```
bkmartinjr: What prevents the user from trying to put two category names into edit mode?    A quick skim of the code makes it look possible -- `activateCategoryEditMode()` in `category.js` doesn't seem to check for this.

bkmartinjr: which leads to the question: does label editing state have the same issue?

colinmegill: Let's open a new issue for validation. Issue #930 just concerned changing the form of the input.

bkmartinjr: OK w/ me.  But please file it and cross-link to this PR so we don't lose track of it.

client/src/components/categorical/category.js
colinmegill: See `portalContainer` here: https://blueprintjs.com/docs/#core/components/dialog

It's going to be attached to `document.body`. Similarly, we have componetized our `toasters` and called them in functions. Rendering is independent of the DOM structure, unlike all the other react elements in the component tree.

bkmartinjr: What I'm confused about is that the Dialog was removed by this PR.   So why is the comment still present in the code?

client/src/components/categorical/category.js
colinmegill: Will change

bkmartinjr:
also, please update the PR description to indicate it fixes 930.  TY

colinmegill:
@bkmartinjr it already does — first line, right above the image :)

PR #963, opened 2019-10-01
reduce dot scaling upon zoom

### LINKED ISSUE #958: Dot size scaling issue upon zoom
It should be that as you zoom in, the dot size does increase some, but not as quickly as the zoom rate. A.k.a. the _spacing_ between the dots should _increase_ as you zoom in (which now it is not), but also the dot size should increase some.

I am almost positive you (@bkmartinjr) had set it to be the desired way described above when we worked on the dot scaling a while back, so it may have reverted or been overridden in some way since then (if that helps at all in fixing it).

bkmartinjr:
Fixes #958 

Reduce the rate of point size scaling with zoom in the main graph.   Previously, dot size would scale faster than white space (distance between points).  Now this is reversed, and dot size scales slower, increasing the "spread" between dots as you zoom in.

This PR also removed a bit of obsolete/unused code related to camera setup.

### DIFF
--- a/client/src/components/graph/drawPointsRegl.js
+++ b/client/src/components/graph/drawPointsRegl.js
@@ -31,7 +31,7 @@ export default function(regl) {
       getFlags(flag, isNaN, isSelected, isHighlight);
 
       float size = pointSize(nPoints, minViewportDimension, isSelected, isHighlight);
-      gl_PointSize = size * pow(distance, 1.2);
+      gl_PointSize = size * pow(distance, 0.5);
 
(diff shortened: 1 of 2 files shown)


colinmegill:
I might increase the effect a touch, but this is better and solves the problem!

bkmartinjr:
@liaprins-czi - I plan to merge today, but please let me know if you want to tweak the parameter as Colin has suggested.

liaprins-czi:
@colinmegill When you say:
> I might increase the effect a touch, but this is better and solves the problem!
... do you mean increase the dot sizes as zoom or decrease the dot sizes as zoom?

PR #1015, opened 2019-11-04
add context sensitivity to annotation "label" menu item

### LINKED ISSUE #972: Annotations: option to "Re-label these cells as ..." should be disabled if all of them already are part of that label, or if 0 cells are selected
If I have selected the cells within a given label (selected them via the label's checkmark at least, as in first screenshot) then by definition they are already part of (all of) that label, so the overflow menu item to "Re-label these cells as [insert name of that label]" should be disabled. It is confusing otherwise.

Screen Shot 2019-10-03 at 3 34 37 PM

Same for the _unassigned_ label when it is the only label, and any number of cells are selected, since by definition they are already part of _unassigned_.
- Also, in the overflow menu the word "unassigned" should be italic to make it more associated that it is that "special" type of label that has to always exist.

Screen Shot 2019-10-03 at 3 20 50 PM

Lastly, when there are 0 cells selected, the options to add to any label should be disabled.

Screen Shot 2019-10-03 at 3 43 44 PM

bkmartinjr:
Fixes #972 

Changes requested from UI review:
1. annotation value (label) context menu will enable/disable "re-label" option.  It is disabled if:
    * no cells are selected
    * all selected cells are already labelled with this this value
2. performance improvements to typedCrossfilter to remove jank from this operation (caching of state derived from current selection)
3. the small formatting (italic 'unassigned') change mentioned in #972 
4. removed hardwired `unknown` string and replaced with the globally configured `unassigned` value for a labels.

### DIFF
--- a/client/src/components/categorical/value.js
+++ b/client/src/components/categorical/value.js
@@ -17,13 +17,16 @@ import * as globals from "../../globals";
 import styles from "./categorical.css";
 import { Tooltip } from "@blueprintjs/core";
 
+import { AnnotationsHelpers as AH } from "../../util/stateManager";
+
 @connect(state => ({
   categoricalSelection: state.categoricalSelection,
   annotations: state.annotations,
   colorScale: state.colors.scale,
   colorAccessor: state.colors.colorAccessor,
   schema: state.world?.schema,
-  world: state.world
+  world: state.world,
+  crossfilter: state.crossfilter
 }))
 class CategoryValue extends React.Component {
   constructor(props) {
@@ -119,6 +122,7 @@ class CategoryValue extends React.Component {
     * world state
     * the color accessor (what is currently being colored by)
     * if this catagorical value's selection status has changed
+    * the crossfilter (ie, global selection state)
 
     If and only if true, update the component
     */
@@ -137,12 +141,14 @@ class CategoryValue extends React.Component {
     const worldChange = props.world !== nextProps.world;
     const colorAccessorChange = props.colorAccessor !== nextProps.colorAccessor;
     const annotationsChange = props.annotations !== nextProps.annotations;
+    const crossfilterChange = props.crossfilter !== nextProps.crossfilter;
 
     return (
       valueSelectionChange ||
       worldChange ||
       colorAccessorChange ||
-      annotationsChange
+      annotationsChange ||
+      crossfilterChange
     );
   };
 
@@ -172,6 +178,27 @@ class CategoryValue extends React.Component {
       categoryIndex
     });
   };
+
+  isAddCurrentSelectionDisabled(category, value) {
+    /*
+    disable "add current selection to label", if one of the following is true:
+    1. no cells are selected
+    2. all currently selected cells already have this label, on this category
+    */
+    const { crossfilter, world } = this.props;
+
+    // 1. no cells selected?
+    if (crossfilter.countSelected() == 0) {
+      return true;
+    }
+    // 2. all selected cells already have the label
+    const mask = crossfilter.allSelectedMask();
+    if (AH.allHaveLabelByMask(world.obsAnnotations, category, value, mask)) {
+      return true;
+    }
+    // else, don't disable
+    return false;
+  }
 
   render() {
     const {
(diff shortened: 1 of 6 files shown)


client/src/util/stateManager/annotationsHelpers.js
```diff
@@ -132,6 +132,22 @@ export function setLabelByMask(df, colName, mask, label) {
 	return ndf;
 }
 
+export function allHaveLabelByMask(df, colName, label, mask) {
+	// return true if all rows as indicated by mask have the colname set to label.
+	// False if not.
+	const col = df.col(colName);
+	if (!col) return false;
+	if (df.length != mask.length)
```
colinmegill: Double equals https://stackoverflow.com/questions/1946063/in-javascript-is-same-as, single equals considered unsafe

client/src/components/categorical/value.js
```diff
@@ -17,13 +17,16 @@ import * as globals from "../../globals";
 import styles from "./categorical.css";
 import { Tooltip } from "@blueprintjs/core";
 
+import { AnnotationsHelpers as AH } from "../../util/stateManager";
```
colinmegill: I prefer the longer name here rather than AH, perhaps `AnnoUtil`?

client/src/components/categorical/value.js
```diff
@@ -173,6 +179,27 @@ class CategoryValue extends React.Component {
...
+  isAddCurrentSelectionDisabled(category, value) {
+    /*
+    disable "add current selection to label", if one of the following is true:
+    1. no cells are selected
+    2. all currently selected cells already have this label, on this category
+    */
+    const { crossfilter, world } = this.props;
+
+    // 1. no cells selected?
+    if (crossfilter.countSelected() == 0) {
```
colinmegill: Prefer `===` https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

client/src/components/categorical/value.js
```diff
@@ -173,6 +179,27 @@ class CategoryValue extends React.Component {
...
+      return true;
+    }
+    // 2. all selected cells already have the label
+    const mask = crossfilter.allSelectedMask();
+    if (AH.allHaveLabelByMask(world.obsAnnotations, category, value, mask)) {
```
colinmegill: I would request a name change for this helper...  it's accurate but I think `mask` is a concept internal to the crossfilter. `selection`?

client/src/components/categorical/value.js
bkmartinjr: This is the style I use elsewhere in the app.  I would rather maintain consistency.  Personally, I think the function name contains all the information, and the package name is not super useful (hence the abbreviated form).   However, if you prefer the verbose form, I can make the mod everywhere in the app.   LMK please.

client/src/components/categorical/value.js
bkmartinjr: fully agree!

client/src/components/categorical/value.js
bkmartinjr: mask is not internal - it is a fully supported interface used all over the place, to bridge crossfilter to rendering (graph) and dataframes (subsetting, etc).

Can you suggest a better name?  I spent a lot of time trying to come up with something originally, and failed :-)

client/src/components/categorical/value.js
colinmegill: Took a look... I prefer full length for readability, even if verbose. `AnnotationsHelpers` is ok with me. We don't use it in a ton of places.

bkmartinjr: roger!

client/src/components/categorical/value.js
colinmegill: I think this is worth making an issue for and perhaps discussing in greater length. This code, in `client/src/reducers/universe`, presents a conceptual interface which spans the concept of a `selection` and a `mask`, with little for the programmer to grasp in terms of what handoff is being made between the concepts IMO. 

image

colinmegill: Since it's consistent with the rest of the codebase, I'm ok approving this as 'status quo' so long as an issue is created.

client/src/util/stateManager/annotationsHelpers.js
bkmartinjr: fixed

PR #1019, opened 2019-11-05
Update Dockerfile

prete:
Added python-requests package to the Dockerfile so cellxgene can use URLs as a data source.


bkmartinjr:
@prete - thank you for the PR.  We really appreciate community contributions.

PR #1023, opened 2019-11-06
Warnings on anno textfields

### LINKED ISSUE #970: Annotations: Need warning message if user tries to name/edit name of category or label to be same as an existing one
Nothing happens when I have edited a category name, or edited a label name to match an existing category or label name, respectively. I try to click the check button to save the new name, but nothing happens, because we don't allow two of the same names, which is fine, we just need to communicate why they are not allowed to save it. 

- [x] Warning for _creating_ new *category* with an existing name
- [x] Warning for _creating_ new *label* with an existing name
- [x] Warning for _editing_ new *category* with an existing name
- [x] Warning for _editing_ new *label* with an existing name

When attempt to edit name, and it matches an existing one, could look something like this instead (but for labels, replace "category" with "label", or whatever we call them in the docs/UI...):
 
Screen Shot 2019-10-03 at 3 32 27 PM

Also the checkmark button should appear disabled.

When attempting to create a new category or new label it could be the same idea, with orange outline on form field and inserted helper text explaining the error. For the modal, probably could make it a bit taller at all times by default to better accommodate this potential text, so it doesn't "jump" the height of the modal when it inserts).

### LINKED ISSUE #1017: Annotations: create label named 'unassigned' gives no error
Run with annotations enabled, create a category, and then create a label and name it 'unassigned'

The pop-up prevents the user from submitting this (ie, no error), but gives no message or other information about why this input is invalid.  It would be nice if there was some hint/warning.

colinmegill:
Handles part of #970 related to adding new categories and labels, but not editing in this PR

image

image

Fixes #1017 

image

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -145,7 +145,10 @@ class Categories extends React.Component {
                         color: Colors.ORANGE3
                       }}
                     >
-                      That category already exists
+                      <span style={{ fontStyle: "italic" }}>
+                        {newCategoryText}
+                      </span>{" "}
+                      already exists
                     </p>
                   </div>
 
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -7,7 +7,8 @@ import {
   InputGroup,
   Dialog,
   Classes,
-  MenuItem
+  MenuItem,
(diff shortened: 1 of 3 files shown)


colinmegill:
Label

image

Category

image

PR #1025, opened 2019-11-06
Improve CLI help

### LINKED ISSUE #961: Update + reorganize CLI --help text
- Updated the `--help` text describing the options in the CLI to be more consistent + better explanations in some cases.
- Updated the formatting of the option text itself in some cases
- Changed the order to be alphabetical

Everything is documented in this G-Doc (note that ✅ = no change and 🆕 = there is a change):
https://drive.google.com/open?id=1nmNyGv4BeaPZFNvhshBLAWr5oE-pndksltyGGi58ppI

bkmartinjr:
Fixes #961

Includes:
* refinements to CLI help messages
* --skip-qc/--run-qc changed from a toggle to a flag named --skip-qc (consistency with all other CLI options).
* alpha sort options when printed in help.

### DIFF
--- a/server/cli/cli.py
+++ b/server/cli/cli.py
@@ -4,8 +4,15 @@ from .launch import launch
 from .prepare import prepare
 
 
-@click.group(name="cellxgene", context_settings=dict(max_content_width=85))
-@click.version_option(version="0.12.0", prog_name="cellxgene", message="[%(prog)s] Version %(version)s")
+@click.group(name="cellxgene",
+             context_settings=dict(max_content_width=85,
+                                   help_option_names=['-h', '--help']))
+@click.help_option("--help", "-h", help="Show this message and exit.")
+@click.version_option(
+    version="0.12.0",
+    prog_name="cellxgene",
+    message="[%(prog)s] Version %(version)s",
+    help="Show the software version and exit.")
 def cli():
     pass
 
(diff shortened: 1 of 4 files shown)


bkmartinjr:
@liaprins-czi - some of these are easy, some are likely not possible without significant effort (we are using a third-party package to manage arguments/options, and it has opinions about some of this stuff).   I'll fix what I can, summarize the end-result, and ping you for another review when available for review.   Thanks!

bkmartinjr:
Update / response to the punch list above:

> It can be hard to visually line up the option with its description

This would require that we overrise some of the core parts of the library we use for argument processing (output formatting).   The library we use conforms to the standard CLI formatting (no extra space) which is designed to condense output -- see (for example) `ls --help`.   It is possible to make the change, but would add complexity and maintenance effort.   Can you confirm this is important enough to warrant the work?

> in launch --help the -s, --scripts option has default listed as [default: ]

fixed

> In prepare --help the first sentence "Preprocess data for use with cellxgene." Is separate

fixed

> I missed this one when working on the G-Doc: I think --sparse (within prepare --help) 

fixed

> The Usage: section of cellxgene --help is fine that [OPTIONS] COMMAND [ARGS] 

I changed these to the following (I think this was per your request -- LMK if I misunderstood):
* for cellxgene --help:   `Usage: cellxgene <options> COMMAND <args>`
* for launch --help: `Usage: cellxgene launch <options> <path to data file>`
* for prepare --help: `Usage: cellxgene prepare <options> <path to data file>`

PR #1027, opened 2019-11-07
Console error on create category submit

### LINKED ISSUE #1010: keyboard [enter] when creating new category throws error
Using keyboard Enter key to submit a new category create causes the action to be dispatched with a null string category name, resulting in an Exception "user annotations require string name" thrown by the an error check in the Universe state management code.

Root cause:  in categorical.js, Categories.onSubmit() assumes it will get an onBlur (and save the name to the component state variable newCateogryText) *before* the submit event is received.

This is not true when the user hits [Enter] -- no blur is received.

To reproduce:
1. start with annotations enabled
2. click Create new Category
3. Type a name *and hit Enter*  (do not click the submit button or otherwise use the mouse -- ie, no blur event)

colinmegill:
Fixes #1010

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -76,26 +76,30 @@ class Categories extends React.Component {
       >
         {/* READ ONLY CATEGORICAL FIELDS */}
         {/* this is duplicative but flat, could be abstracted */}
-        {_.map(allCategoryNames, catName =>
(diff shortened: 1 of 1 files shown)


colinmegill:
A note for posterity: this could have been a Blueprint form handling bug we just worked around, and if they fix it, may need to revisit.

PR #1048, opened 2019-11-15
Anno editing label/category error states

### LINKED ISSUE #970: Annotations: Need warning message if user tries to name/edit name of category or label to be same as an existing one

colinmegill:
Fixes #970


client/src/components/categorical/category.js
```diff
@@ -334,7 +347,11 @@ class Category extends React.Component {
                     rightElement={
                       <Button
                         minimal
-                        disabled={newCategoryText.length === 0}
+                        disabled={
+                          (allCategoryNames.indexOf(newCategoryText) > -1 &&
```
bkmartinjr: I think you should factor this out into a helper function, much like we did with labelError() and labelErrorMessage().

client/src/components/categorical/category.js
```diff
@@ -344,6 +361,23 @@ class Category extends React.Component {
...
+                    <span
+                      style={{
+                        display: "block",
+                        fontStyle: "italic",
+                        fontSize: 12,
+                        marginTop: 5,
+                        color: Colors.ORANGE3
+                      }}
+                    >
+                      {newCategoryText === ""
```
bkmartinjr: See comment on line 351

client/src/components/categorical/category.js
```diff
@@ -344,6 +361,23 @@ class Category extends React.Component {
                       />
                     }
                   />
+                  {(allCategoryNames.indexOf(newCategoryText) > -1 &&
```
bkmartinjr: see comment on line 351

client/src/components/categorical/value.js
```diff
@@ -32,7 +33,11 @@ class CategoryValue extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      editedLabelText: ""
+      editedLabelText: String(
+        props.categoricalSelection[props.metadataField].categoryValues[
```
bkmartinjr: nit - this lookup occurs all over the file, and probably should be a helper function (for example, see the same thing in handleDeleteValue())

client/src/components/categorical/value.js
```diff
@@ -348,6 +356,13 @@ class CategoryValue extends React.Component {
               <form
                 onSubmit={e => {
                   e.preventDefault();
+                  if (
+                    (category.categoryValues.indexOf(editedLabelText) > -1 &&
```
bkmartinjr: much like we did in categorical.js and category.js, this error check,  and the corresponding error message (line 416) should be factored into a helper function for clarity.

client/src/components/categorical/value.js
```diff
@@ -358,6 +373,13 @@ class CategoryValue extends React.Component {
                   }}
                   small
                   autoFocus
+                  intent={
+                    (category.categoryValues.indexOf(editedLabelText) > -1 &&
```
bkmartinjr: see comment on line 360

client/src/components/categorical/value.js
```diff
@@ -374,6 +402,22 @@ class CategoryValue extends React.Component {
                     />
                   }
                 />
+                {(category.categoryValues.indexOf(editedLabelText) > -1 &&
```
bkmartinjr: see comment on line 360

client/src/components/categorical/category.js
```diff
@@ -159,22 +159,16 @@ class Category extends React.Component {
...
-    const { metadataField, universe } = this.props;
-    const { obsByName } = universe.schema.annotations;
-
-    if (obsByName[metadataField].categories.indexOf(name) !== -1) {
-      error = "duplicate";
-    }
-
-    if (!AnnotationsHelpers.isLegalAnnotationName(name)) {
-      error = "characters";
+    if (name) {
```
colinmegill: This probably could be moved outside the function itself, depending on how many times this function is called. Imagining: 

`name && errorCheck()`

colinmegill: That doesn't add much to the verbosity of the inline check

client/src/components/categorical/category.js
```diff
@@ -239,6 +231,20 @@ class Category extends React.Component {
           {`${newCategoryText} already exists`}
         </span>
       );
+    } else if (err === "characters") {
```
colinmegill: This may be two subcomponents later on.

client/src/components/categorical/category.js
bkmartinjr: it all depends on what you want the default behavior to be.  If you look at line 510, you see that we occasionally say `!name && errorCheck()`, so you have to pick a default.  In this case, most of the invocations preferred that and error only occurs when name is a non-zero length string.

I guess we could also pass in the desired behavior as an arg, eg, `errorCheck(val, nullIsAnError=true)`

I'm not dogmatic on this - fix to match your own taste...

client/src/components/categorical/category.js
```diff
@@ -185,6 +196,83 @@ class Category extends React.Component {
...
+        <span
+          style={{
+            display: "block",
+            fontStyle: "italic",
+            fontSize: 12,
+            marginTop: 5,
+            color: Colors.ORANGE3
+          }}
+        >
+          {`${newCategoryText} already exists`}
```
bkmartinjr: error messages sometimes contain the entered text (`newCategoryText`), sometimes do not.   Given that this is an inline component, I'm not sure it adds much value to replicate it.  And space is a premium, which also argues for omitting the user text in favor of a more verbose error message.

client/src/components/categorical/category.js
```diff
@@ -185,6 +196,83 @@ class Category extends React.Component {
...
+        <span
+          style={{
+            display: "block",
+            fontStyle: "italic",
+            fontSize: 12,
+            marginTop: 5,
+            color: Colors.ORANGE3
+          }}
+        >
+          {`${newCategoryText} contains illegal characters`}
```
bkmartinjr: see above comment about including the user input in the message.

client/src/components/categorical/value.js
```diff
@@ -70,23 +72,90 @@ class CategoryValue extends React.Component {
...
+      markup = (
+        <span
+          style={{
+            fontStyle: "italic",
+            fontSize: 12,
+            marginTop: 5,
+            color: Colors.ORANGE3
+          }}
+        >
+          {`${editedLabelText} already exists`}
```
bkmartinjr: not sure we should be including the user text entry (`editedLabelText`) in the error message given that this is an inline edit.

client/src/components/categorical/value.js
bkmartinjr: I think you missed this check - shouldn't this call the category error check?

client/src/components/categorical/value.js
bkmartinjr: I think that doing something like I suggested would help the code clarity throughout these files.

bkmartinjr: @colinmegill has elected to leave this for now, and handle in a future refactor.

PR #1053, opened 2019-11-19
[WIP] JS lint and dead code removal

### LINKED ISSUE #1044: chore: eslint before ship
We are gathering some lint - do a full round of delinting pre-0.13.

bkmartinjr:
lint, cleanup and dead code removal before our next release

there are no intentional functional changes in this PR

Fixes #1044


client/src/components/continuous/continuous.js
```diff
@@ -84,7 +85,7 @@ class Continuous extends React.Component {
                     isObs
                     zebra={zebra % 2 === 0}
                     ranges={summary}
-                    handleColorAction={this.handleColorAction(key).bind(this)}
+                    handleColorAction={this.handleColorAction(key)}
```
colinmegill: Though lint will complain about `.bind(this)`, if a function is being passed into a subcomponent, this is still necessary. `handleColorAction` calls `this.props` expecting parent scope.

bkmartinjr: true, but see change on line 28.  It will be bound to the scope because it is an assigned property.

See the note about property initializers here: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#es6-classes

PR #1070, opened 2019-12-19
Makefile modularity, test targets, and auto-formatting

mweiden:
This is a collection of some "prim and proper" improvements to cellxgene's build, test, and deploy `make` commands to ease development:
1. Add ~[`yapf`](https://github.com/google/yapf)~ `black` as a python auto-formatter. It and flake8 are configured together such that if you run `make fmt` in the `server` module, the code will be formatted to avoid most flake8 errors.
1. Create `unit-test` and `smoke-test` make targets. This makes running tests faster and easier to remember.
1. Modularize the root `Makefile`: move `client` and `server`-specific make commands to makefiles in the modules themselves. This has two benefits:
    1. It removes implementation logic like `npm`/`python` commands from the root `Makefile` that only the modules themselves need to know about. This makes the root `Makefile` more readable.
    1. It provides ways of templating sub-tasks like `unit-tests` that can then be tied together in the root and called in one go with a command like `make unit-test`.

**NOTE**: If you want to run the server unit tests from the root directory, you should now do this with `make unit-test-server`.


mweiden:
@bkmartinjr The one other goodness that could result from this--perhaps in a separate PR--is to make JS linting part of the CI/CD pipeline. In this new make setup, all you'd have to do is to add `lint-client` to the `lint` target in the root `Makefile`.

server/app/util/fbs/NetEncoding/Column.py
```diff
@@ -4,6 +4,7 @@
 
 import flatbuffers
```
bkmartinjr: Everything in this file is machine generated and should not be reformatted or edited in any way.   With flake8's config, we ignore this directory.  Need to match that.

server/app/util/fbs/NetEncoding/Float32Array.py
```diff
@@ -4,6 +4,7 @@
```
bkmartinjr: Everything in this file is machine generated and should not be reformatted or edited in any way.

server/app/util/fbs/NetEncoding/Float64Array.py
```diff
@@ -4,6 +4,7 @@
```
bkmartinjr: Everything in this file is machine generated and should not be reformatted or edited in any way.

server/.style.yapf
```diff
@@ -0,0 +1,2 @@
+[style]
```
bkmartinjr: We have previously decided on a 120 character column size.  Is the revert to 80 characters intentional?  I'd rather not change this unless there is a good reason (the argument for 120 is that it matches the working style & large format editors most of us work in today).

server/app/util/fbs/NetEncoding/Column.py
mweiden: Missed this. I've made sure the exclusion goes is used in the auto-formatting.

server/.style.yapf
mweiden: I missed this. I've adjusted it back to 120.

mweiden:
@bkmartinjr replying inline.

> * Please don't reformat machine-generated files (server/app/util/fbs/NetEncoding/*) and the others listed in setup.cfg.
Pardon, I missed this. I've added adjusted the black configuration to address this.

> * The move from 120 to 80 character column limit for Python is undesirable.  Was this deliberate?
Nope, just missed the `flake8` configuration in `setup.cfg`. I've changed this back to 120 characters in `black`.

> * Why yapf and not black?  We previously had a plan to use black at some point, as it has slightly better features (eg, string quoting).
I didn't have a strong opinion on this. Happy to use `black`. It does seem to work a little bit better than `yapf` out of the box.

dev_docs/developer_guidelines.md
```diff
@@ -21,20 +53,24 @@
 If you install cellxgene using `make install-dev` the server will be restarted every time you make changes on the server code. If changes affects the client, the browser must be reloaded.
 
 ### Linter
-We use `flake8` to lint code. Travis CI runs `flake8 server`.
+
+We use `yapf` to auto-format and lint python.
```
bkmartinjr: s/yapf/black/g

bkmartinjr: also, unless I'm misreading the Makefile, we use black to auto-format, and flake8 to lint.

mweiden: Good catch 👍 Fixed this.

mweiden:
> 1. There are still a few mentions of yapf in the .md (docs) and PR description.

Good catch. Updated these.

> 2. You may want to note in the PR description that running Python unit tests no longer works from the top-level directory (you changed the paths locating test data).

You can still run the python unit tests from the root with `make unit-test-server`. You're right, though, that cellxgene devs may not be familiar with this. I've updated the PR description and the dev guidelines with more detailed instructions.

> LMK if you want another task and I'll start assigning you bugs! :-)

I'm definitely interested in more bugs. I will see if I can pick one of the ones off the backlog. If you have specific recommendations for a cxg newb, send them my way.

👍

PR #1074, opened 2019-12-27
allow hyphen in annotation file name

### LINKED ISSUE #1072: Getting 400 error when updating existing annotations file
I am able to open existing annotations with the following command:

> ~/code/nsnx/cellxgene-gateway/.cellxgene-gateway/bin/cellxgene launch ../cellxgene_data/somedir/pbmc3k.h5ad --port 8002 --host 127.0.0.1 --experimental-annotations --experimental-annotations-file ../cellxgene_data/somedir/pbmc3k_annotations/somedirAnno1-MW36MWBW.csv

However, if I edit the annotations, I get a series of 400 errors in the Chrome Developer Tools. The request is a PUT to http://localhost:8002/api/v0.2/annotations/obs?annotation-collection-name=somedirAnno1-MW36MWBW. The error message is:

> Error, bad annotation collection name

bkmartinjr:
Fixes #1072 

Annotations file name now allows hyphen (in total, [0-9][a-z][A-Z]\-\_)


server/app/rest_api/rest.py
```diff
@@ -221,7 +221,7 @@ def is_safe_collection_name(name):
     """
     if name is None:
         return False
-    return re.match(r"^\w+$", name) is not None
+    return re.match(r"^[\w\-]+$", name) is not None
```
mweiden:
```suggestion
    return re.match(r"^(?!\-)[\w\-]+(?<!\-)$", name) is not None
```

We should probably guard against the name starting or ending in a hyphen, right? Fix for this included.

bkmartinjr: Why do we need to include this additional restriction?    -foobar is an entirely legal filename on any system I can think of.

mweiden: Fair enough. Resolving.

Issue #344, opened 2018-10-18
fonts are not locally served


bkmartinjr:
the CLI aspires to bundle and serve all front-end assets locally.  However, there are currently fonts (Roboto) pulled from a CDN.   At some point, need to install and bundle.

neuromusic:
When implementing this, we'll need to pay attention to licensing, too.

bkmartinjr:
@csweaver - this will need to be fixed for the native (stand-alone) UI packaging.

PR #1077, opened 2019-12-31
Serve `Roboto Condensed` font from the app directly

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/344


bkmartinjr:
Reminder to test the production build to ensure that it works correctly (given that there are separate webpack configs for prod & dev)

mweiden:
@bkmartinjr I've removed a reference to google fonts that I missed and tested my changes with the `dev` webpack build. As far as I can tell, it works.

Issue #683, opened 2019-04-01
provide notification when there is a newer version of cellxgene available


neuromusic:
If a newer version of cellxgene has been released, give the user a gentle notification that there is a new version of cellxgene available, with a link to update instructions.

colinmegill:
I don't think this update notice should appear in a hosted context — ie., user seeing the message can't take any action. @liaprins-czi any ideas?

liaprins-czi:
@colinmegill Are you talking about the publication sort of usecase? If so, then I agree. Would there be a way to prevent these users from getting the update message?

seve:
With `prepare` being proposed as our fully supported method of hosting, could we just disable the checks when it is in the static mode?

bkmartinjr:
Suggest that we have a CLI param that disables this feature.    There are other features that the user may wish to disable in a hosted context (eg, diffexp).

PR #1078, opened 2019-12-31
Notify users of new versions of cellxgene

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/683

Behavior:
```bash
0 cellxgene> time cellxgene
Usage: cellxgene <options> COMMAND <args>

Options:
  -h, --help                      Show this message and exit.
  --version                       Show the software version and exit.
  --upgrade-check / --no-upgrade-check
                                  Check for release upgrades on start.  [default:
                                  True]

Commands:
  launch   Launch the cellxgene data viewer. Run `cellxgene launch --help` for more
           information.
  prepare  Preprocess data for use with cellxgene. Run `cellxgene prepare --help`
           for more information.

real	0m1.083s
user	0m1.426s
sys	0m0.406s
```
```bash
0 cellxgene> time cellxgene launch
There's a new version of cellxgene available (0.13.0)!
To upgrade, run the following: pip install --upgrade cellxgene

Usage: cellxgene launch <options> <path to data file>

Error: Missing argument "<path to data file>".

real	0m1.700s
user	0m1.442s
sys	0m0.407s
```
```bash
2 cellxgene> time cellxgene --no-upgrade-check launch
Usage: cellxgene launch <options> <path to data file>

Error: Missing argument "<path to data file>".

real	0m1.032s
user	0m1.385s
sys	0m0.412s
```
```bash
0 cellxgene> cellxgene launch example-dataset/pbmc3k.h5ad 
There's a new version of cellxgene available (0.13.0)!
To upgrade, run the following: pip install --upgrade cellxgene

[cellxgene] Loading data from pbmc3k.h5ad.
[cellxgene] Launching! Please go to http://127.0.0.1:5005 in your browser.
[cellxgene] Type CTRL-C at any time to exit.
```

### DIFF
--- a/server/cli/upgrade.py
+++ b/server/cli/upgrade.py
@@ -41,10 +41,12 @@ def _request_cellxgene_releases():
             raise RateLimitException
     url = "https://api.github.com/repos/chanzuckerberg/cellxgene/releases"
     res = requests.get(url)
+    raise_on_rate_limit(res)
     for release in res.json():
         yield release
     while 'next' in res.links.keys():
         res = requests.get(res.links['next']['url'])
+        raise_on_rate_limit(res)
         for release in res.json():
             yield release
 
@@ -58,8 +60,8 @@ def validate_version_str(version_str, release_only=True):
     """
     match = SEMVER_FORMAT.match(version_str)
     has_match = match is not None
-    if release_only:
-        return has_match and not match.group("prerelease")
+    if has_match and release_only:
+        return not match.group("prerelease")
     return has_match
 
 
--- a/server/cli/upgrade.py
+++ b/server/cli/upgrade.py
@@ -1,7 +1,7 @@
 import click
 import re
+import requests
 
-from github import Github, RateLimitExceededException
 from requests.exceptions import ConnectionError
 from .. import __version__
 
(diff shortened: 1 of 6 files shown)


mweiden:
@bkmartinjr note, I tested this by setting `__version__ = "0.12.0"` by hand in `server/__init__.py`.

server/cli/upgrade.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class RateLimitException(Exception):
+    """
+    Github API Rate Limit Exception
+    """
+
+
+def _request_cellxgene_releases():
+    def raise_on_rate_limit(response):
```
bkmartinjr: where is this called?

server/cli/upgrade.py
```diff
@@ -0,0 +1,82 @@
...
+        yield release
+    while 'next' in res.links.keys():
+        res = requests.get(res.links['next']['url'])
+        for release in res.json():
+            yield release
+
+
+def validate_version_str(version_str, release_only=True):
+    """
+    Test if a string conforms to SemVer format (https://semver.org/)
```
bkmartinjr: The regex can also throw an AttributeError, right?  Should it return False in that case or let the exception propagate?

bkmartinjr: and if the exception is part of the API, add a test

server/cli/upgrade.py
mweiden: Good catch. I forgot to call it!

server/cli/upgrade.py
mweiden: I've rewritten the function such that the call that would cause the `AttributeError` (`match.group("prerelease")`) will only be called in the case that there _is_ a match. In this case, no `AttributeError` can be raised.

PR #1102, opened 2020-01-08
Fix label sorting bugs

### LINKED ISSUE #1091: Swatch mismatch
I have not reproduced this error, and it is likely connected to #1090. 

When hovering over a duplicated label, let's say this is a new category duplicated from mouse gender and the hovered value is female and the swatch is purple. Green dots lit up in the graph. Likewise, hovering on orange lit up purple. 

This resolved on refresh, once unassigned was present. It seems more likely this is connected to unassigned being missing than to the hovering code which has been solid.

### LINKED ISSUE #1095: Labels + their assigned colors should have a more meaningful sort order
- [ ] When a category is toggled open, the labels within it should appear in alphabetical and/or numeric order (alphabetical if letter characters, numeric if number characters) similar to how they are today, but with one update: Any label called `nan` (or `Nan`, `NaN`, etc) _or_ called `Unassigned` should go down to the very bottom. (Currently the `nan` labels are sandhiched between label starting with an "m" and label starting with "o".)

- [ ] Also, the colors assigned to labels should follow this sort order, so they look like they are in a rainbow spectrum from top to bottom.

bkmartinjr:
Fixes #1091 
Fixes #1095 

Changes:
* adds the special handling for 'nan' labels
* allocates colors in a manner preserving the default rainbow "order"
* fixes several bugs which caused loss of sort order:  top N truncation, and adding a user specified label

In addition the special "category sort" algo was promoted to a common utility (previously was buried in the category component).


client/src/util/stateManager/schemaHelpers.js
```diff
@@ -30,6 +31,15 @@ export function indexEntireSchema(schema) {
 	return schema;
 }
 
+export function sortAllCategorical(schema) {
+	/* UI relies on OBS annotation categories being in presentation sort order */
+	schema.annotations.obs.columns.forEach(c => {
+		if (c.categories) {
+			c.categories = catLabelSort(c.writable, c.categories);
```
colinmegill: Spacing looks off here

client/src/util/stateManager/schemaHelpers.js
```diff
@@ -90,7 +100,11 @@ export function addObsAnnoCategory(schema, name, category) {
 
 	const newSchema = _reindex(_copy(schema));
 
-	/* remove category */
-	newSchema.annotations.obsByName[name].categories.push(category);
+	/* add category, retaining presentation sort order */
+	const catAnno = newSchema.annotations.obsByName[name];
+	catAnno.categories = catLabelSort(catAnno.writable, [
+		...catAnno.categories,
```
colinmegill: Spacing

client/src/util/stateManager/schemaHelpers.js
bkmartinjr: the whole file was tab indented.  Fixed now.

client/src/util/stateManager/schemaHelpers.js
bkmartinjr: the whole file was tab indented. Fixed now.

PR #1103, opened 2020-01-09
Rename Makefile target `build-server` to `build-cli`

mweiden:
The `build-server` makefile target does not actually build the server
module, rather it builds the CLI. Rename this target to make it less
confusing.


mweiden:
@seve @sidneymbell @colinmegill flagging you all on this in case you use this target.

Issue #648, opened 2019-03-15
remove d3 dependency from crossfilter


bkmartinjr:
crossfilter currently uses a d3 implementation of point-in-polygon:
* method https://wrf.ecse.rpi.edu//Research/Short_Notes/pnpoly.html
* repo: https://github.com/substack/point-in-polygon/blob/master/index.js

Break the dependency on d3 by incorporating the algo directly in crossfilter.withinPolygon()

PR #1108, opened 2020-01-12
Remove d3 dependency in crossfilter

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/648

### DIFF
--- a/client/src/util/typedCrossfilter/crossfilter.js
+++ b/client/src/util/typedCrossfilter/crossfilter.js
@@ -1,5 +1,3 @@
-import { polygonContains } from "d3";
-
 import PositiveIntervals from "./positiveIntervals";
 import BitArray from "./bitArray";
 import {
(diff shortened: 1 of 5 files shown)


client/src/util/typedCrossfilter/crossfilter.js
```diff
@@ -11,6 +9,8 @@ import {
 } from "./sort";
 import { makeSortIndex } from "./util";
 
+const pointInPolygon = require("point-in-polygon");
```
colinmegill: `require` --> `import`

colinmegill:
Hi @mweiden! 

Could you tell me a little more about this PR? I would expect we would move off of the d3 dependency to another package if we were trying to get off of d3 altogether in the app (I don't think that's a requirement as of now), or if this package is significantly faster.

Otherwise, I would prefer we not merge this, and remain on the d3 version.

colinmegill:
(Otherwise,  we're just adding a different dependency, which — I'm unsure if that was @bkmartinjr's intention)

mweiden:
@colinmegill pardon, I misunderstood the ask in the issue. @bkmartinjr and I just sync'ed and came up with the following plan:

1. Lift [the code referenced in the issue](https://github.com/substack/point-in-polygon/blob/master/index.js) directly into the module (with attribution); add tests.
1. Merge this PR (approval and coarse performance check pending).
1. At a later date, optimize the performance of the code, if necessary.

Thoughts?

colinmegill:
That's what I thought he meant, sounds great. So long as there's no regression in performance from d3, sounds good to me.

PR #1110, opened 2020-01-15
Ontologies

### LINKED ISSUE #996: Autosuggest ontology terms for user-created annotations
Great idea from @aopisco: 

When a user is creating a new categorical annotation, allow them to indicate if it's a cell type label. If so, autosuggest terms from the [cell ontology](http://www.obofoundry.org/ontology/cl.html) to standardize vocabulary. This will be important for enabling reuse of these annotations (e.g., for training classifiers later on).

### LINKED ISSUE #1036: Escape should exit annotation modes
Exiting some annotation modes, like editing a category name, will need hotkeys manually wired up. Blueprint has a helper for this:

https://blueprintjs.com/docs/#core/components/hotkeys.key-combos

### LINKED ISSUE #1050: New label dialog resizing when entering text
Empty textbox:
image

After entering text:
image

image

### LINKED ISSUE #1055: annotations edit - no way to cancel
When I edit a category or label name (user annotations), there is no way to cancel the edit.   There must be a cancel in addition to submit/accept.

### LINKED ISSUE #1057: Annotations: "Create new category" button in modal should be disabled when text field is blank or characters don't meet req's
Currently, the "Create new _label_" button within the modal behaves as expected: 
- disabled at start, before type any characters
- disables when delete all characters I had typed, so it is an empty field again
- disables when I have a types a  duplicate name of an existing label

However, the "Create new _category_" button within its modal does not disable ever.

Also, both buttons should disable if I type an un-allowed character (e.g. too many spaces in a row, etc); not sure if this is the current behavior because I'm waiting to see that PR.

label

category

colinmegill:
Primarily fixes #996, allowing annotations to autosuggest annotation labels given an existing ontology. 

Also fixes #1057 && fixes #1050 && fixes #1036 && fixes #1055 

Jan-14-2020 16-50-05


sidneymbell:
Running list of minor bugs (will edit):
- Pin the "create label not in ontology" to the top of the list
- Button for "create label and assign selected cells" should not appear if I have no selected cells
- New category names should also have suggestions from ontology

client/src/components/categorical/categorical.js
```diff
@@ -27,7 +21,8 @@ class Categories extends React.Component {
     this.state = {
       createAnnoModeActive: false,
       newCategoryText: "",
-      categoryToDuplicate: null
+      categoryToDuplicate: null,
+      activeSuggestItem: null
```
bkmartinjr: this state field seems unused.  Dead code?

client/src/components/categorical/category.js
```diff
@@ -8,15 +8,16 @@ import {
   Tooltip,
   InputGroup,
```
bkmartinjr: Several unused imports - dead code?   InputGroup, AnnoSelect, etc.

client/src/components/categorical/value.js
```diff
@@ -11,11 +11,15 @@ import {
   Position,
   PopoverInteractionKind,
   Tooltip,
-  Colors
+  Colors,
+  Dialog,
+  Classes
```
bkmartinjr: there are a number of unused imports - dead code?

client/src/components/categorical/value.js
```diff
@@ -26,12 +30,16 @@ import { AnnotationsHelpers } from "../../util/stateManager";
...
+  crossfilter: state.crossfilter,
+  ontology: state.ontology,
+  ontologyLoading: state.ontology?.loading,
+  ontologyEnabled: state.ontology?.enabled
 }))
 class CategoryValue extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      activeSuggestItem: "",
```
bkmartinjr: this state value seems unused?

client/src/components/categorical/value.js
```diff
@@ -83,6 +83,20 @@ class CategoryValue extends React.Component {
...
+    const label = this.getLabel();
+
+    dispatch({
+      type: "annotation: label edited",
+      metadataField,
+      editedLabel: editedLabelTextNotInOntology,
+      categoryIndex,
+      label
+    });
+    // this.setState({ newLabelText: "" });
```
bkmartinjr: cleanup

client/src/components/categorical/categorical.js
```diff
@@ -127,119 +135,65 @@ class Categories extends React.Component {
...
-        {_.map(allCategoryNames, catName =>
-          !schema.annotations.obsByName[catName].writable ? (
-            <Category
-              key={catName}
-              metadataField={catName}
-              createAnnoModeActive={createAnnoModeActive}
-              isUserAnno={false}
-            />
-          ) : null
+        {_.map(
```
bkmartinjr: we have moved away from lodash where native JS replacements exist.  In this case, please replace with:  `allCategoryNames.map(...)`

bkmartinjr: and remove the lodash import from this file.

client/src/components/categorical/categorical.js
```diff
@@ -127,119 +135,65 @@ class Categories extends React.Component {
...
-        {_.map(allCategoryNames, catName =>
-          schema.annotations.obsByName[catName].writable ? (
-            <Category
-              key={catName}
-              metadataField={catName}
-              createAnnoModeActive={createAnnoModeActive}
-              isUserAnno
-            />
-          ) : null
+        {_.map(
```
bkmartinjr: use native Array.map()

client/src/reducers/world.js
```diff
@@ -246,21 +261,25 @@ const WorldReducer = (
...
+      return {
+        ...state,
+        ...setLabelOnCurrentSelection(state, crossfilter, metadataField, label)
       };
-      const obsAnnotations = state.obsAnnotations.replaceColData(
-        metadataField,
-        unclipped.obsAnnotations.col(metadataField).asArray()
-      );
-      return { ...state, obsAnnotations, unclipped };
+      // const mask = crossfilter.allSelectedMask();
```
bkmartinjr: Oops, I think I left this dead code.  Please remove the commented section as it is detritus.

client/src/reducers/ontology.js
```diff
@@ -0,0 +1,26 @@
...
+    enabled: false, // are ontology terms enabled?
+    terms: null, // an array of term names, eg, ['cell', 'lung cell', ...]
+    loading: true
+  },
+  action
+) => {
+  switch (action.type) {
+    case "configuration load complete":
+      const enabled =
+        action.config?.parameters?.annotations_cell_ontology_enabled;
```
bkmartinjr: this may set `enabled` to null (rather than false) which is unintended.   Please replace with:
```
const enabled =
        action.config?.parameters?.annotations_cell_ontology_enabled ?? false;
```

(added the nullish coallescing op & false value at end)

client/src/components/categorical/categorical.js
colinmegill: Yes and no... AnnoInputs passes a function here: 

`handleItemChange={this.handleSuggestActiveItemChange}`

and we store the value, but do not currently use it. Conceivable that we would. Remove?

client/src/components/categorical/value.js
colinmegill: Yes and no... AnnoInputs passes a function here:

handleItemChange={this.handleSuggestActiveItemChange}

and we store the value, but do not currently use it. Conceivable that we would. Remove?

client/src/components/categorical/categorical.js
bkmartinjr: I am not a fan of dead code.  If you are going to use the feature in the immediate future it is fine to leave it.  But if it is just speculative, *YAGNI*

client/src/components/categorical/value.js
bkmartinjr: same as above: keep it clean....

PR #1114, opened 2020-01-16
fix incomplete user category name creation

### LINKED ISSUE #1112: user anno - new category names must validate against schema
See categorical.js::categoryNameError() - the user entered category name is validated for uniqueness against the category names in the categoricalSelection summary object.   This object is only guaranteed to have a _subset_ of annotation category names in the schema, so it not detect all collisions.   Collisions can occur with continuous annotations, the index, or other categorical annotations not displayed for some other reason (see controlsHelpers.js::selectableCategoryNames() for other logic).

We need to test for name uniqueness in the obs annotation *schema*, not in the display categories.

bkmartinjr:
User created categories must have a unique name.   The validation code was only checking that the names were unique in the *categorical* obs annotations, and would incorrectly allow a collision with other annotations (eg, continuous annotations).  This would later lead to a crash/error and loss of data.

This PR expands the check to require uniqueness across all obs annotation names.

Fixes #1112

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -71,8 +71,12 @@ class Categories extends React.Component {
     /* allow empty string */
     if (name === "") return false;
 
-    const { categoricalSelection } = this.props;
-    const allCategoryNames = Object.keys(categoricalSelection);
+    /* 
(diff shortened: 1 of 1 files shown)


colinmegill:
Good catch, looks good!

PR #1115, opened 2020-01-16
disable magical nan parsing in back-end CSV handler

### LINKED ISSUE #1111: issue with creating user annotation values named `nan`
The following behavior is seen on user-created annotation values named `nan` both manually created and copied from read-only categories:
1. nan -> null upon refresh
2. Manually naming a label nan and then refreshing also changes it to null
3. nan works like a normal label, but null has several bugs, a couple listed below:
4. Hovering on null enlarges dots as expected, but turning those cells on with the checkbox interaction does not enlarge dots (but it should)
5. Selecting just a null label for diff exp, nothing happens when click to make it the first diff exp group, but other labels try to pull in the cells that are in null when I try to add them to diff exp selection

---
HT: @liaprins-czi for the list of behaviors

bkmartinjr:
Fixes #1111 

We use the Pandas read_csv implementation, which includes all sorts of "na" handling for series data.  This fix forces all labels to be treated as generic strings, consistent with the remainder of our label handling (agnostic to any label or cat name semantics).

### DIFF
--- a/server/app/scanpy_engine/labels.py
+++ b/server/app/scanpy_engine/labels.py
@@ -9,7 +9,7 @@ import pandas as pd
 
 def read_labels(fname):
     if fname is not None and os.path.exists(fname) and os.path.getsize(fname) > 0:
-        return pd.read_csv(fname, dtype="category", index_col=0, header=0, comment="#")
+        return pd.read_csv(fname, dtype="category", index_col=0, header=0, comment="#", keep_default_na=False)
     else:
         return pd.DataFrame()
 


sidneymbell:
Hmm. Perhaps we should instead do a `df.fill_na('unassigned')`? 

Fair to assume that if a value is `NaN`, it should be treated like all the other missing values rather than having two different labels that are both, effectively, unassigned.

bkmartinjr:
how would fill_na() solve the problem?

The core issue boils down to:
* the user entered a label named 'nan' (a string, not a magic value)
* we use Pandas which has special treatment for this string, for categorical series.  We don't have those semantics, and don't want them, as our abstraction is "dumb CSV"

These are not "real" floating point NaNs...they are a real label name the user typed in. 

Principle of minimum surprise clearly points toward "do no unexpected transformations"

PR #1119, opened 2020-01-21
Update legal characters for labels

### LINKED ISSUE #1113: reconcile OBO/OWL legal term names with our current label name restrictions
Current code does not allow user to select a (legal) ontology term if it contains various non-alphanumeric characters (eg, a comma, dash, etc).   Anything in the ontology file should be legal (which means any unicode point).   In addition, this will require validation that the CSV escaping is correctly handled when saving/loading.

Notes from discussion on desired approach:
1 - We still validate user input when they enter a label (i.e., not an ontology term), and still restrict to a subset of all characters.
2 - When they enter an ontology term, we use the exact term, even if it contains characters not in the set used to validate de novo terms in #1.
3 - To make these behaviors as similar as possible, we will add ( ) - to the allowed characters list in user-entered labels legal characters (see #1).
4 - CSV escaping will be properly implemented for load and save

End result:
* any term in the ontology will be selectable, loadable and savable.
* de novo terms entered by the user, not in the ontology, will be restricted to a sub-set of ASCII `A-Za-z0-9_-()`

@sidneymbell - please edit if I missed something.

sidneymbell:
This PR makes two changes:
* custom annotations category names and labels may contain additional characters `/^(\w|[ .()-])+$/`
* if ontology suggest is enabled, allow use of any term in the ontology even if it has characters not legal in user-specified labels

Fixes #1113 

Open issue:

- [ ] **We still validate user input when they enter a label (i.e., not an ontology term), and still restrict to a subset of all characters.**
There is an outstanding bug on @colinmegill's annotation branch where no labels are being checked if ontologies are enabled. Will wait to merge until this is fixed.

### DIFF
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -200,6 +200,9 @@ class Category extends React.Component {
     /* allow empty string */
     if (name === "") return false;
 
+    /* allow any term in the ontology */
+    if (this.props.ontology.terms.includes(name)) return false;
+
     /* check for label syntax errors */
     const error = AnnotationsHelpers.annotationNameIsErroneous(name);
     if (error) return error;
@@ -257,7 +260,7 @@ class Category extends React.Component {
       "empty-string": "Blank names not allowed",
       duplicate: "Category name must be unique",
       "trim-spaces": "Leading and trailing spaces not allowed",
-      "illegal-characters": "Only alphanumeric, underscore and period allowed",
+      "illegal-characters": "Only alphanumeric and special characters (-_.) allowed",
       "multi-space-run": "Multiple consecutive spaces not allowed"
     };
     const errorMessage = errorMessageMap[err] ?? "error";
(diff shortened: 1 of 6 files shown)


client/src/components/categorical/category.js
```diff
@@ -200,6 +200,9 @@ class Category extends React.Component {
     /* allow empty string */
     if (name === "") return false;
 
+    /* allow any term in the ontology */
+    if (this.props.ontology.terms.includes(name)) return false;
```
bkmartinjr: if the ontology is large, this is going to be slow/laggy (remember, it gets evaluated upon each keystroke event).

I suggest changing the component to have a derived property which is a Set object containing the ontology terms, and testing for the input value presence in the set.

bkmartinjr: Second thought - you will need this same query in categorical.js.  So you might as well create a single shared Set object in the reducer, and store in the main app state.  That way the two components do not create duplicate objects.

bkmartinjr:
> Category name validation should also allow existing ontology terms as a legal name, right? If so, you are missing a check in categorical.js::categoryNameError().

From talking to Colin, I now understand the team decided to omit ontology suggest from category name for this release.   Ignore this review comment.

PR #1127, opened 2020-01-24
Add --server-shutdown flag to enable user to terminate web interface.

blankenberg:


blankenberg:
Adds a power button when `--server-shutdown` is passed (outlined in red below)

power

bgruening:
Cool! That would be nice to have!

bkmartinjr:
Hi @blankenberg - thanks for the PR!   Could you help by providing a user story or rationale for the feature?   I am trying to understand the use case it fulfills.  Thank you.

blankenberg:
Hi @bkmartinjr thanks for looking! 

The basic idea is to give a user the ability to shutdown a running server without requiring access to or knowledge of the underlying infrastructure. Specifically, my use-case revolves around cellxgene being available as a Galaxy (https://galaxyproject.org/) tool.

A simplified outline:
- user adds datasets to their Galaxy history (upload their own datasets, access from public repositories, etc.)
- user runs some set of tools/operations to qc/prepare/analyze their data, resulting in a AnnData `h5ad` dataset.
- user configures and runs cellxgene as a Galaxy tool
- Galaxy schedules and executes cellxgene *somewhere*
- Galaxy proxies the web interface of and access to this isolated web application to the user
- User does stuff in cellxgene, e.g. creates custom annotations.
- User clicks power button. cellxgene terminates gracefully
- user created files (e.g. annotations) are processed by Galaxy and end up back inside of the user’s history/workspace -- these new datasets can be used as input to another run of cellxgene or in a downstream operation

Presence of REST endpoint enables additional integration hooks through (Galaxy), e.g. save and exit gracefully via shutdown in a list of User’s currently running tools.

Please let me know if I can provide any additional information.

ambrosejcarr:
Hi @blankenberg Thanks for the use case! 

Could you talk a bit more about why you decided to design cellxgene-based shutdown instead of controlling the process from the galaxy application? I don't know enough about Galaxy to understand the trade-offs.

mweiden:
Hi @blankenberg. We haven't heard back from you in a while and the branch has gone a bit stale, so I'm going to close this PR.

If you think we should reconsider this modification, please resubmit with an updated PR. That said, it seems like this use case could be fulfilled with a kill signal sent from galaxy (or other software wrapper) to cellxgene, when you want it to shut down.

bgruening:
@mweiden we do have this wrapper, but from a user point of view it makes a lot of sense to "close" an application.

A user open cellxgene with the Galaxy infrastructure, means Galaxy will schedule the application with its normal job submission. This means we can handle 1000s of applications and serve 1000s of users with its "own" celxgene instance in a very accessible way. No need to deal with clouds or CLI and so on. But this also means that the user sees cellxgene as kind of an App - and here a "close" option in the fronted you hugely benefit the UI experience.

blankenberg:
No worries about closing, sorry for letting this PR go stale. 

We have had a working cellxgene integration in Galaxy for a while now, but this was one of those missing user experience pieces. I still think it would be generally useful to optionally allow this control directly from cellxgene for Galaxy and otherwise, but I did not have funding (unfortunately still do not) to work on this further.

Issue #1123, opened 2020-01-22
resolve pinned anndata package requirement


bkmartinjr:
AnnData 0.7* is not compatible with cellxgene at the moment. The 0.14 release will pin the dependency to the 0.6.22post1 version of anndata.

We need to resolve the incompatibilities and then remove the pinned dependency.

Related: #917

alokito:
This bit me this morning when we redeployed a server... between this and the h5py issue, might be worth pinning all dependencies, at least for released versions.

PR #1132, opened 2020-01-27
[WIP] Patch for anndata 0.7

sidneymbell:
JSON encoding error caused by anndata 0.7.* was due to a change to `type(anndata.__version__)`. 

A bit tricky to find (thanks for the pointer, @bmccandless!), but a simple one-line fix. 

Reverts the temporary pin in the requirements back to `anndata>=0.6.2`.

### DIFF
--- a/server/app/rest_api/rest.py
+++ b/server/app/rest_api/rest.py
@@ -45,7 +45,7 @@ class ConfigAPI(Resource):
                 },
                 "links": {"about-dataset": current_app.config["ABOUT_DATASET"]},
                 "parameters": {**current_app.data.get_config_parameters(uid=cxguid, collection=anno_collection)},
-                "library_versions": {"cellxgene": cellxgene_version, "anndata": anndata_version},
+                "library_versions": {"cellxgene": cellxgene_version, "anndata": str(anndata_version)},
             }
         }
 
--- a/server/app/util/matrix_proxy.py
+++ b/server/app/util/matrix_proxy.py
@@ -73,6 +73,8 @@ class MatrixProxy(_ArrayProxyBase):
         "numpy.ndarray": True,
         "scipy.sparse.csc.csc_matrix": True,
         "scipy.sparse.csr.csr_matrix": True,
+        "anndata._core.sparse_dataset.backed_csr_matrix": "MatrixProxy_anndata_h5py",
+        "anndata._core.sparse_dataset.backed_csc_matrix": "MatrixProxy_anndata_h5py",
     }
     proxy_registry = None
     last_cache_token = None
--- a/server/app/scanpy_engine/matrix_proxy.py
+++ b/server/app/scanpy_engine/matrix_proxy.py
@@ -38,6 +38,8 @@ class MatrixProxy_anndata_h5py(MatrixProxyView):
             "anndata.h5py.h5sparse.backed_csc_matrix",
             "anndata.h5py.h5sparse.backed_csr_matrix",
             "h5py._hl.dataset.Dataset",
+            "anndata._core.sparse_dataset.backed_csr_matrix",
+            "anndata._core.sparse_dataset.backed_csc_matrix",
         )
 
     @classmethod
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -1,5 +1,4 @@
-# TEMP anndata 0.7 breaks cellxgene - workaround
-anndata==0.6.22post1
+anndata>=0.6.2
 click>=6.7
 fastobo>=0.6.1
 Flask>=1.0.2


sidneymbell:
I found another issue: anndata now wraps the scipy sparse matrix datatypes in their own anndata-specific classes. 

I've updated the allowed input types for the proxy matrix, but this involves anndata's private classes so may not be the desired solution. @bmccandless -- can you take a look and weigh in on this? 

All tests pass and it works as expected on my box, but good to have another set of eyes smoketest, as well.

bkmartinjr:
@sidneymbell - does this fix #1123 or is it still unresolved?

ivirshup:
Hey! Just thought I'd weigh in a few points:

> I found another issue: anndata now wraps the scipy sparse matrix datatypes in their own anndata-specific classes.

We should only do this in backed mode.

> I've updated the allowed input types for the proxy matrix,  but this involves anndata's private classes so may not be the desired solution.

I think we'd be amenable to exporting `SparseDataset`. Thoughts @flying-sheep?

server/app/util/matrix_proxy.py
```diff
@@ -73,6 +73,8 @@ class MatrixProxy(_ArrayProxyBase):
         "numpy.ndarray": True,
         "scipy.sparse.csc.csc_matrix": True,
         "scipy.sparse.csr.csr_matrix": True,
+        "anndata._core.sparse_dataset.backed_csr_matrix": "MatrixProxy_anndata_h5py",
+        "anndata._core.sparse_dataset.backed_csc_matrix": "MatrixProxy_anndata_h5py",
```
flying-sheep: I don’t understand why mentioning our classes here is necessary: They are subclasses of `scipy.sparse.csc.csc_matrix`/`csr_matrix` and should therefore hit the cases that are already there. So I assume a better fix is to make your registry work with subclasses.

If I’m wrong, that’s because I didn’t look into what your MatrixProxy does.

server/app/rest_api/rest.py
```diff
@@ -45,7 +45,7 @@ def get(self):
                 },
                 "links": {"about-dataset": current_app.config["ABOUT_DATASET"]},
                 "parameters": {**current_app.data.get_config_parameters(uid=cxguid, collection=anno_collection)},
-                "library_versions": {"cellxgene": cellxgene_version, "anndata": anndata_version},
+                "library_versions": {"cellxgene": cellxgene_version, "anndata": str(anndata_version)},
```
flying-sheep: uh, this looks like a bug on our part, `__version__` should always be a string I think!

server/app/util/matrix_proxy.py
ivirshup: @flying-sheep, we've stopped having these be user visible, `adata.X` should be a `SparseDataset` which returns in memory sparse matrices on subsetting.  Users shouldn't see a something that subclasses from a scipy sparse matrix in `0.7`.

flying-sheep: Current master says otherwise or am I missing something? https://github.com/theislab/anndata/blob/c09124375ac2828bc325081d5efe630a3ab87e82/anndata/_core/sparse_dataset.py#L117

ivirshup: `adata.X` should return a `SparseDataset` (similar to dense arrays and `h5py.Dataset`). The backed matrix classes are an implementation detail.

server/app/util/matrix_proxy.py
```diff
@@ -73,6 +73,8 @@ class MatrixProxy(_ArrayProxyBase):
         "numpy.ndarray": True,
         "scipy.sparse.csc.csc_matrix": True,
         "scipy.sparse.csr.csr_matrix": True,
+        "anndata._core.sparse_dataset.backed_csr_matrix": "MatrixProxy_anndata_h5py",
```
bkmartinjr: These do not belong in this (general purpose) file.  All ScanPy/AnnData driver code is in the scanpy_engine module.

server/app/util/matrix_proxy.py
bkmartinjr: I will have a look and see if we can accomplish our goal in the way @flying-sheep suggests.  Happy to make it cleaner.

All that the "MatrixProxy" is doing is providing a guaranteed set of interfaces to the underlying AnnData data access classes.  This allows the rest of our code to ignore the many underlying implementations of a "matrix".   In various older versions of AnnData, we had various problems with what adata.X or adata.X[] might return:
* lack of function, eg, transposition and column iteration were not supported by some classes
* different semantics ,eg, the dimension dropping in numpy but not scipy sparse
* performance bugs in backed mode which required us to build our own "view" object

The code above is simply trying to sniff the object returned by either adata.X or adata.X[], and make a determination if we need to wrap it in a proxy.   Ideally, this would be based upon stable & public types, or attributes of the object, which reliably indicate behavior.

For 0.6.*, using csr_matrix above, as suggested by @flying-sheep, won't tell us if the matrix is backed or not.   The issue we are working around is the performance bug in 0.6 where any sub-slice of a backed sparse array reads the entire array into memory.    The actual API is fine, but the implementation of that API in backed mode caused failures.

I'm happy to simply update the cellxgene requirement to anndata>=0.7 and put a cleaner set of tests in.  I'd be even happier if I could remove the Matrix entirely :-)

Appreciate advice....

bkmartinjr:
@sidneymbell - I will start another PR to continue this work.  Lets leave this closed & merged for now, as it is not causing any instability in master.

server/app/util/matrix_proxy.py
flying-sheep: Ah, now I understand, thank you. About your questions

- You should always slice the adata object, not the `.X`, that way you can be sure you always have a relatively predictable 2D array – relatively because it can still be a `scipy.sparse.cs[rc]_matrix` or np.ndarray
- To test if an object is currenly backed, use `adata.isbacked` instead of implementation details
- AnnData 0.7 should fix quite a few bugs, did you try if it fixes theislab/anndata#79 or your use case here?

Generally, please don’t use the undocumented, private API, it can and will break without prior warning (like in any other Python project). If you want things that aren’t part of the public API, please tell us and we’ll help you get it implemented ASAP!

bkmartinjr: Thanks for the feedback.  I will need to look more at the latest version to understand slicing on adata vs. X. In prior versions, this did not meet our needs.   It would be fantastic if that has changed.

Can you give me any insights into the indexing behavior on adata?  I have noticed on the 0.7.1 release, for example, that slicing adata is significantly slower than X, for a large in-memory datasets.

Example, on a 1Mx1K dense matrix (the old mouse neuron dataset from 10X), loaded into memory (backed=None):
* row slicing, eg, `adata[0:10, :].X` vs. `adata.X[0:10, :]` is over 100,000X slower
* col slicing, eg, `adata[:, [3, 77, 900]].X` vs `adata.X[:, [3, 77, 900]]` is roughly 20X slower

My naive test code:
```
import anndata
import timeit

adata = anndata.read_h5ad('10x_mouse_13MM_processed.h5ad', backed=None)

def go(f):
    t = timeit.Timer(f).autorange()
    return t[1] / t[0]

print('row slices')
t1 = go(lambda: adata[0:10, :].X)
t2 = go(lambda: adata.X[0:10, :])
print(f'{t1:g}, {t2:g}, ratio: {t1/t2:g}')

print('col slices')
t1 = go(lambda: adata[:, [3, 77, 900]].X)
t2 = go(lambda: adata.X[:, [3, 77, 900]])
print(f'{t1:g}, {t2:g}, ratio: {t1/t2:g}')
```

Running it on my laptop, I get:
```
row slices
5.90577, 9.49219e-07, ratio: 6.22171e+06
col slices
0.999165, 0.0456957, ratio: 21.8657
```

If I load the data with `backed='r'`, the difference is not quite as stark, but it is still much slower:
```
row slices
6.04593, 0.0508553, ratio: 118.885
col slices
2.1715, 1.11327, ratio: 1.95057
```

This level of overhead is quite significant to our use case.   Am I incorrectly using the adata slicing?

flying-sheep: You’re not, this seems like a performance bug. Did you file it?

bkmartinjr: I did not yet file it.  I will.   Literally just found it before I sent this comment.

In general, I have not yet had a chance to poke around with 0.7 and really validate how well it works for us.   I will try to make time for that over the next few weeks, possibly as part of exploring your proposed slicing interface change.

ivirshup: Could you show us the repr of the object? Could you also share the file?

ivirshup: Ah, I have a hunch about what's happening here now. I think this is expected, though we are working to cut down on the discrepancy. Have you checked if this is a regression?

The short version of what's happening is that taking a view of an AnnData object is much more complicated than taking a view of a numpy array, and I think it's always going to be a few orders of magnitude slower. However, there are two particular issues – how neighbor graphs are stored and `raw` – which make views of `AnnData` objects particularly slow. We have a fix for the neighbor graph storage (they should go in `obsp` now, but we have to update `scanpy`'s `neighbors` method to use that). `raw` would have to have views implemented.

flying-sheep: Ah, so the problem is that they’re slicing a lot of tiny slices other than a handful … in that case, yes, it’s always going to be slower.

bkmartinjr: I'll drop the repr and other info into theislab/anndata#317.

And yes, we frequently take lots of small slices while incrementally fetching data for the UI tier. Example: fetch X for a single gene, to build a histogram of expression. The overhead in the current full adata slice would be a significant hit to our interactive performance.

No idea if it is a regression.  We have always sliced `X` directly (at least as long as I've worked on the project - 2ish years).    This is the first time - AFAIK - where we have been advised to switch to using the top-level adata slicing, and consequently this is the first exploration of the tradeoffs.  Previously we used patterns found in ScanPy, primarily because they were more performant.

Couple of thoughts:
* what we really want, and why we implemented "MatrixProxy", is a very fast and light weight way to slice X, without materializing it, and return a consistent ndarray/spmatrix/DataFrame subarray of X.   In principle, full adata slicing is a great pattern, and would allow us to completely remove MatrixProxy (win!).   But it needs to be roughly the same speed as slicing X - ie, it should act more like a completely lazy view, and then only slice out the data requested (X in our case, could be other stuff in other use cases).
* The ScanPy code seems to break the same rules we did, probably for the same reason.  For example, see scanpy/tools/_rank_genes_groups.py where X is pulled from the adata, and directly sliced.

Is there a path to making `adata[...].X` roughly on par with `adata.X[...]`?

flying-sheep: OK, let’s continue in theislab/anndata#317.

bkmartinjr: @ivirshup - I'm working through your recommendation that we slice `adata`, not `adata.X`, and finding it is not compatible with 0.6.22post1, with at least some datasets.

In this case, I have a copy of the pbmc3k dataset, re-encoded as CSC sparse (written using 0.6.22post1), and opened in backed mode.  Any slicing of adata fails.  Slicing of adata.X does not fail.

```
# python
Python 3.7.6 (default, Dec 19 2019, 23:49:42) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import anndata
>>> anndata.__version__
'0.6.22.post1'
>>> adata = anndata.read_h5ad('pbmc3k-CSC-gz.h5ad', backed='r')
>>> adata
AnnData object with n_obs × n_vars = 2638 × 1838 backed at 'server/test/test_datasets/pbmc3k-CSC-gz.h5ad'
    obs: 'n_genes', 'percent_mito', 'n_counts', 'louvain'
    var: 'n_cells'
    uns: 'draw_graph', 'louvain', 'louvain_colors', 'neighbors', 'pca', 'rank_genes_groups'
    obsm: 'X_pca', 'X_tsne', 'X_umap', 'X_draw_graph_fr'
    varm: 'PCs'
>>> adata[0,0].X
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/bruce/projects/cellxgene/oldad-venv/lib/python3.7/site-packages/anndata/core/anndata.py", line 1230, in __getitem__
    return self._getitem_view(index)
  File "/mnt/c/Users/bruce/projects/cellxgene/oldad-venv/lib/python3.7/site-packages/anndata/core/anndata.py", line 1234, in _getitem_view
    return AnnData(self, oidx=oidx, vidx=vidx, asview=True)
  File "/mnt/c/Users/bruce/projects/cellxgene/oldad-venv/lib/python3.7/site-packages/anndata/core/anndata.py", line 561, in __init__
    self._init_as_view(X, oidx, vidx)
  File "/mnt/c/Users/bruce/projects/cellxgene/oldad-venv/lib/python3.7/site-packages/anndata/core/anndata.py", line 633, in _init_as_view
    self._raw = adata_ref.raw[oidx]
  File "/mnt/c/Users/bruce/projects/cellxgene/oldad-venv/lib/python3.7/site-packages/anndata/core/anndata.py", line 336, in __getitem__
    if self._adata is not None or not self._adata.isbacked: X = self._X[oidx, vidx]
TypeError: 'NoneType' object is not subscriptable
>>> adata.X[0,0]
-0.17146951
>>> adata.X
<HDF5 sparse dataset: format 'csc', shape (2638, 1838), type '<f4'>
```

This seems to be related: theislab/anndata#234

Is there a method that will work with both 0.6.* and 0.7.* anndata?

ivirshup: > I'm working through your recommendation that we slice adata, not adata.X, and finding it is not compatible with 0.6.22post1, with at least some datasets.

 sure I'd say always subset the AnnData object. I think it's likely you'll get better performance from indexing into the sub-element directly (especially if you only subset once).

> In this case, I have a copy of the pbmc3k dataset, re-encoded as CSC sparse (written using 0.6.22post1), and opened in backed mode. Any slicing of adata fails. Slicing of adata.X does not fail.

I think I recognize this bug. It has to do with subsetting along var when `raw` is defined.

Does this work if the file was written with 0.6.22post1 but read with anndata 0.7? If so, could pinning anndata above version 0.7 be a solution?

bkmartinjr: Pinning the dependency fixes it, but will force all of our users to update.  Our team is reluctant to force an update, out of concern for users who don't use a venv, and would potentially have other apps/code destabilized by the update.

Trying to reduce this to the core issue -- I think it comes down to the "right" way to index when all you want is a slice of X.   @ivirshup noted:

> sure I'd say always subset the AnnData object. I think it's likely you'll get better performance from indexing into the sub-element directly (especially if you only subset once).

I had been following @flying-sheep's recommendation, above, which I read as indicating that slicing the sub-object was unsupported API when in backed mode:

> You should always slice the adata object, not the .X, that way you can be sure you always have a relatively predictable 2D array – relatively because it can still be a scipy.sparse.cs[rc]_matrix or np.ndarray

Is use of `adata.X[...]` supported, or not, when the dataset is loaded `backed='r'`?    And will it always return these expected types?    Anecdotally, it seems to work., and if that is true, it has the benefit of working with both 0.6 and 0.7 generations.

Would appreciate consensus advice on this -- ideally I'd like to align with your intended API usage patterns, and avoid future problems :-)

ivirshup: > Trying to reduce this to the core issue -- I think it comes down to the "right" way to index when all you want is a slice of X.

My perspective would be that both are correct. However, we don't have as much control over the results of `adata.X[...]`, so it can be inconsistent. For example, we can promise that `adata[obsidx, varidx].X.shape == adata[obsidx, varidx].shape`. We also support more forms of indexing when you subset `adata` instead of `X`. For example, there are cases where `adata.X[obs_idx, var_idx] != adata.X[obs_idx, :][:, var_idx]`.

For a novice user, I'd probably recommend just subsetting the AnnData object. For more experienced developers I'd say dealing with the underlying arrays directly avoids some machinery, and can be more efficient. However, that machinery is there to prevent footguns and give consistent behaviour across array types.

> Is use of adata.X[...] supported, or not, when the dataset is loaded backed='r'? And will it always return these expected types? Anecdotally, it seems to work., and if that is true, it has the benefit of working with both 0.6 and 0.7 generations.

My expectation is that `adata.X[...]` will work in all released versions (bar-ing bugs), in that X should always be array-ish. We don't have control of what `adata.X[...]` does unless we've defined the class of `X`, which is the case for `SparseDatasets`.

@flying-sheep, would you agree with that assessment?

---------------------------------------------

Separately, @bkmartinjr fair point about versioning. However, what's the time and memory overhead on `adata[0,0].X`? For me, it's equivalent to reading the matrix into memory, due to the performance bugs with backed mode in that release. Given this, I don't think I'd recommend supporting backed mode for pre 0.7 releases.

flying-sheep: Slicing the sub-object isn’t unsupported, you just have less reliable output. Slicing an AnnData object means that X will afterwards always be a 2D array.

Maybe we should implement `AnnData.A` which will always return a 2D numpy array. Then people could do `adata[idx].A` and then call `.squeeze()` or `.flatten()` on the result and it’d always work, no matter what’s in there.

bkmartinjr: > My expectation is that adata.X[...] will work in all released versions (bar-ing bugs), in that X should always be array-ish. 

As long as there is a reasonable subset of slicing/indexing operations available in the array-ish return value, _or_ a way to detect which are supported (eg, based upon public typing), I think you have proposed an excellent candidate solution:
* use `adata.X[...]`, which works for anndata 0.6 & 0.7
* strongly discourage/disable use of backed when anndata<=0.7

Currently we use simple scalar slicing only, eg, `adata.X[:, [0, 99]]`, which seems simple enough to be supported by most array-ish types.

flying-sheep: Yeah, this will actually give you a 2D array-ish object even when indexing with an 1 element list. If you however do `adata.X[:, 5]`, you get either a 1D numpy array or a 2D scipy sparse matrix.

bkmartinjr: Thanks @flying-sheep & @ivirshup.  I think I have 99% of this cleaned up on our end, thanks to your help.

One more small question:  is there a reliable & fast way to detect if the X matrix is sparse/dense when the adata is opened with `backed="r"`?   In other words, equivalent of `adata.issparse()`. 

I have hacky ways to do it, but am not sure if there is a safe/good way:
* slice it and check type, eg, `scipy.sparse.isspmatrix(adata.X[0, 0:1])`
* see if it "looks like" a duck:  `getattr(adata.X, 'format_str', False)`, or look at `type(adata.X)`

Ideally I'd like something that works on anndata>=0.6.*, but could live with 0.7.

flying-sheep: That’s exactly the kind of feature request we want to see over at anndata’s issue tracker!

We’re very happy to make peoples’ lives easier by implementing new supported ways of doing things. The process is simple:

1. open feature request
2. discuss how to optimally do things
3. implement in your project:

   ```py
   if new_way_is_supported():
       new_way_that_is_officially_supported()
   else:
       hacky_way_that_works_with_old_versions()
   ```

bkmartinjr: happy to file it!

PR #1143, opened 2020-02-05
update anndata dependency requirement; remove h5py pinned version

### LINKED ISSUE #917: remove temp pin of h5py when regression fixed
Remove the temporary pinning of h5py module when theislab/scanpy#832 is resolved.

PR that made the change: #916.

bkmartinjr:
Remove the temporary pinned dependency on h5py.  It is believed the incompatibility is resolved by the 0.7 anndata release.

This also bumps the minimum requirement for anndata to >=0.7.1, as the combination of anndata <= 0.7 and h5py 2.10 will trigger the compat bug.

Fixes #917

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -12,6 +12,4 @@ numpy>=1.15.2
 pandas>=0.24.2
 scipy>=1.3.0
 tables==3.5.1
-# TEMP workaround for https://github.com/theislab/scanpy/issues/832 aka h5py regression
-h5py==2.9.0
 requests>=2.22.0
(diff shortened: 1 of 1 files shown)


bkmartinjr:
If anndata 0.7 has a bug, *please* file it ASAP with them.

Can you elaborate on your claim that most users will not run in a virtual environment?

If the above bug in anndata is fixed, do you see any concerns upping the min dependency?   My other conversation with the Anndata team suggests that it may be required in a future cellxgene release for other reasons.

bkmartinjr:
Closing this in favor of #1157

PR #1145, opened 2020-02-05
js package dep update

bkmartinjr:
update to resolve various security bulletins & vulns (in our dev tools chain)

Note that this PR leaves start-server-and-test at 1.10.6.    The most recent 1.10.8 updates its own internal dependencies and fails.


seve:
LGTM

PR #1147, opened 2020-02-05
optimize centroid util

seve:
This PR refactors the scope of several variables in the centroid utility, removing repeated evaluations from the centroid coordinate calculation loop and unnecessary member accessing.

### DIFF
--- a/client/src/util/centroid.js
+++ b/client/src/util/centroid.js
@@ -6,11 +6,87 @@ import { unassignedCategoryLabel } from "../globals";
   Centroid coordinate calculation
 */
 
-/* 
-  calcMedianCentroid goes through a given metadata category
-  fetches each cell's coordinates grouping by category value.
+// Generates a mapping of categorical values to data needed to calculate centroids
+// categoricalValue -> {
+//  size: int,
+//  holdsFinite: Boolean,
+//  xCoordinates: Float32Array,
+//  yCoordinates: F32 Array
+// }
+const getCoordinatesByCategoricalValues = (
+  obsAnnotations,
+  categoryName,
+  obsLayout,
+  layoutDimNames,
+  categoricalSelection,
+  schemaObsByName
+) => {
+  console.time("create scratch pad");
+  const categoryArray = obsAnnotations.col(categoryName).asArray();
 
-  It then calculates the median value and puts that in the array
+  const layoutXArray = obsLayout.col(layoutDimNames[0]).asArray();
+  const layoutYArray = obsLayout.col(layoutDimNames[1]).asArray();
+
+  const scratchPad = new Map();
+
+  const { categoryValueIndices, categoryValueCounts } = categoricalSelection[
+    categoryName
+  ];
+
+  // Check to see if the current category is a user created annotation
+  const isUserAnno = schemaObsByName[categoryName].writable;
+
+  // Iterate over all cells
+  for (let i = 0, len = categoryArray.length; i < len; i += 1) {
+    // Fetch the categorical value of the current cell
+    const categoryValue = categoryArray[i];
+
+    // Get the index of the categoryValue within the category
+    const categoryValueIndex = categoryValueIndices.get(categoryValue);
+
+    // If the category is truncated and this value is removed,
+    //  it will not be assigned a category value and will not be
+    //  labeled on the graph
+    // If the user created this category,
+    //  do not create a label for the `unassigned` value
+    if (
+      categoryValueIndex !== undefined &&
+      !(isUserAnno && categoryValue === unassignedCategoryLabel)
+    ) {
+      // Create/fetch the scratchpad value
+      let value = scratchPad.get(categoryValue);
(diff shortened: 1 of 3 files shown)


seve:
Benchmark findings:

|  | Master | This Branch |
|---------------------------------|--------------------|--------------------|
| Fetching Cell Embeddings | 92.746826171875ms | 85.798095703125ms |
| Calculating Every Median | 276.736083984375ms | 277.915771484375ms |
| Entirety of centroid utility | 277.09814453125ms | 364.6201171875ms |
| Centroid Button Press to Render | 398.28515625ms | 387.950927734375ms |

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
...
 
-/* 
-  calcMedianCentroid goes through a given metadata category
-  fetches each cell's coordinates grouping by category value.
-
-  It then calculates the median value and puts that in the array
+/*
+Generates a mapping of categorical values to data needed to calculate centroids
+categoricalValue -> {
+ size: int,
```
bkmartinjr: comment is wrong - should be 'length'

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
...
+ holdsFinite: Boolean,
+ xCoordinates: Float32Array,
+ yCoordinates: F32 Array
```
bkmartinjr: Float32Array

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
   Centroid coordinate calculation
```
bkmartinjr: I would add a prominent comment that this calculates centroids for a truncated set of labels.  Ie, it follows the display logic of truncating categories that have excessive labels.

it is not a general purpose centroid calculator, as it will ignore labels in those categorical annotations that have more than the current system truncation setting.

(this is a feature, but also means that this code is only useful for display purposes that want the centroids for the list of labels we will show the user in the LSB)

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
...
-const calcMedianCentroid = (
+const getCoordinatesByCategoricalValues = (
   obsAnnotations,
   obsLayout,
   categoryName,
   layoutDimNames,
   categoricalSelection,
   schemaObsByName
 ) => {
+  const scratchPad = new Map();
```
bkmartinjr: pick a better name - this name suggests that the contents are not results!

how about:  coordinatesByCategory?

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
...
-      valueArray[3][index] = layoutYArray[i];
+      const coordinatesLength = value.length;
 
-      coordinates.set(categoryValue, valueArray);
+      value.xCoordinates[coordinatesLength] = layoutXArray[i];
+      value.yCoordinates[coordinatesLength] = layoutYArray[i];
+
+      value.length = coordinatesLength + 1;
+
+      scratchPad.set(categoryValue, value);
```
bkmartinjr: why do the redundant set() on each value?   If the `value` object is already in the Map(), you don't need to add it again.

Suggestion:
* when you first initialize the `value` object, at line 62, perform the set().
* remove the set() on each iteration through the loop.

in other words,
* you should do a set once per category label value
* you should do a get on each iteration.

client/src/util/centroid.js
```diff
@@ -6,100 +6,131 @@ import { unassignedCategoryLabel } from "../globals";
...
-        new Float32Array(numInCategoryValue), // x coordinates
-        new Float32Array(numInCategoryValue) // y coordinates
-      ];
-      const index = valueArray[1];
-      let hasFinite = valueArray[0];
-
-      hasFinite =
-        hasFinite ||
+      // Create/fetch the scratchpad value
+      let value = scratchPad.get(categoryValue);
```
bkmartinjr: calling this variable `value` is confusing given all the other uses of the term.

perhaps:
* scratchpad becomes `coordsByCategoryLabel`
* value becomes `coords`
Or something like that.

client/src/util/centroid.js
seve: I'm a little confused here, if there value object is being mutated and then we are inserting the updated objected via `set()`  is that redundant? Since the object returned by `get()` is a copy, any mutations will be lost once we exit scope if we don't insert.

bkmartinjr: get does not return a copy - it returns the actual object

seve: You're right, when I tested it earlier I was just using primitives as the values in the hashmap and wasn't getting the expected values *facepalm*

seve: fix commited

PR #1148, opened 2020-02-05
Collect all env vars in one, easy-to-find place

mweiden:
Past state:
* Default environement variables were stored in both client/package.json
and client/__tests__/e2e/config.js
* Constants that should have been linked with a variable--like the cellxgene server port
during testing--were repeated.

With this commit:
* All environment variables are parameterized
* All environment variables are packaged in default env files

HEADS UP: _This PR will require you to set an environment before you start developing!_ I've included `common.mk` to warn you, if you haven't loaded an environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ install:
   - set -eo pipefail
   - pip install flake8
   - make pydist install-dist dev-env
+  - source environment.ci
 
 jobs:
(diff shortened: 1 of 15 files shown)


mweiden:
@bkmartinjr taking this out of PR and bringing it back when it's cleaned up.

PR #1149, opened 2020-02-08
Collect all env vars in one, easy-to-find place

mweiden:
Past state:
* Default environement variables were stored in `client/package.json`, `client/__tests__/e2e/config.js`, `client/server/development.js`, `client/src/globals.js`, and `server/cli/launch.py`
* Constants that should have been linked--like the cellxgene server port during testing--were repeated.

With this commit:
* All environment variables are parameterized
* All environment variables are packaged in default env files
* `make` targets (replacing npm scripts) keep everything locked together

NOTE: _These changes change development flow. FE developers will have to use `make` commands instead of `npm` commands._ To ease the transition most commands are the same, all you have to do is use `make <command>` instead of `npm run <command>`. For more, see the [developer scripts](https://github.com/chanzuckerberg/cellxgene/blob/a5d1b6463d2c35af32ec84404b8f70ecb687ae25/dev_docs/developer_scripts.md) documentation.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -25,7 +25,7 @@ build:
 
 .PHONY: server-requirements
 server-requirements:
-	virtualenv -p python3.6 ../venv
+	virtualenv -p python3 ../venv
 	source ../venv/bin/activate && \
 		pip install -r ../server/requirements-dev.txt && \
 		yes | pip uninstall cellxgene || true && \
@@ -35,10 +35,13 @@ server-requirements:
 start-frontend:
 	node server/development.js
 
+.PHONY: start-server
+start-server:
+	cellxgene launch -p $(CXG_SERVER_PORT) $(CXG_OPTIONS) $(DATASET)
+
 .PHONY: backend-dev
 backend-dev: server-requirements
-	source ../venv/bin/activate && \
-		cellxgene launch -p $(CXG_SERVER_PORT) $(CXG_OPTIONS) $(DATASET)
+	source ../venv/bin/activate && $(MAKE) start-server
 
 .PHONY: backend-dev-anno
 backend-dev-anno: server-requirements
@@ -62,10 +65,7 @@ test:
 
 .PHONY: smoke-test
 smoke-test:
-	start_server_and_test \
-		'cellxgene launch -p $(CXG_SERVER_PORT) $(DATASET)' \
-		$(CXG_SERVER_PORT) \
-		'$(MAKE) e2e'
+	start_server_and_test '$(MAKE) start-server' $(CXG_SERVER_PORT) '$(MAKE) e2e'
 
 .PHONY: unit-test
 unit-test:
@@ -74,4 +74,3 @@ unit-test:
 # pass remaining commands through to npm run
 %:
 	npm run $(*)
-
--- a/client/Makefile
+++ b/client/Makefile
@@ -1,3 +1,5 @@
+include ../common.mk
+
 .PHONY: clean
 clean:
 	rm -rf node_modules
--- a/client/Makefile
+++ b/client/Makefile
@@ -1,5 +1,9 @@
 include ../common.mk
 
+DATASET := $(if $(DATASET),$(DATASET),../example-dataset/pbmc3k.h5ad)
+
+# Packaging
+
 .PHONY: clean
 clean:
 	rm -rf node_modules
@@ -15,6 +19,57 @@ install:
 .PHONY: build
 build:
 	npm run build
+
+
+# Development convenience methods
+
+.PHONY: server-requirements
+server-requirements:
+	virtualenv -p python3.6 ../venv
+	source ../venv/bin/activate && \
+		pip install -r ../server/requirements-dev.txt && \
+		yes | pip uninstall cellxgene || true && \
+		pip install -e ..
+
+.PHONY: start-frontend
+start-frontend:
+	node server/development.js
+
+.PHONY: backend-dev
+backend-dev: server-requirements
+	source ../venv/bin/activate && \
+		cellxgene launch -p $(CXG_SERVER_PORT) $(CXG_OPTIONS) $(DATASET)
+
+.PHONY: backend-dev-anno
+backend-dev-anno: server-requirements
+	CXG_OPTIONS='--experimental-annotations' $(MAKE) backend-dev
+
+.PHONY: backend-dev-anno-ontology
+backend-dev-anno-ontology: server-requirements
+	CXG_OPTIONS='--experimental-annotations --experimental-annotations-ontology' \
+		$(MAKE) backend-dev
+
+.PHONY: e2e
+e2e:
+	node node_modules/jest/bin/jest.js \
+		--verbose false \
+		--config __tests__/e2e/e2eJestConfig.json \
+		e2e/e2e.test.js
+
+.PHONY: test
+test:
+	node node_modules/jest/bin/jest.js
+
+.PHONY: smoke-test
+smoke-test:
+	start_server_and_test \
+		'cellxgene launch -p $(CXG_SERVER_PORT) $(DATASET)' \
+		$(CXG_SERVER_PORT) \
+		'$(MAKE) e2e'
+
+.PHONY: unit-test
+unit-test:
+	node node_modules/jest/bin/jest.js --testPathIgnorePatterns e2e
 
 # pass remaining commands through to npm run
 %:
--- a/common.mk
+++ b/common.mk
@@ -1,7 +1,7 @@
 PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
 PATH := $(PATH):$(PROJECT_ROOT)/scripts
 
-SHELL := env PATH=$(PATH) /bin/bash
+SHELL := env PATH='$(PATH)' /bin/bash
 
 # get_or_else_dev_env_default
 # - If a variable is defined, return its value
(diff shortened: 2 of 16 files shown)


bkmartinjr:
Adding @seve and @colinmegill as reviewers, as this has direct impact on their FE development workflow.

common.mk
```diff
@@ -0,0 +1,19 @@
+PROJECT_ROOT := $(shell git rev-parse --show-toplevel)
+PATH := $(PATH):$(PROJECT_ROOT)/scripts
+
+SHELL := env PATH=$(PATH) /bin/bash
```
bkmartinjr: this will fail if there is a space in any path in $PATH, due to lack of escaping.  Suggest:
> SHELL := env PATH='$(PATH)' /bin/bash

environment.default
```diff
@@ -0,0 +1,6 @@
+# Default environment variables for cellxgene development
+set -a
+CXG_SERVER_PORT=5000
```
bkmartinjr: this used to be `5005`.  Is there a good reason to change the default port number?  I'm just worried about unintended downstream impact on other dependent repos.

server/cli/launch.py
```diff
@@ -19,6 +19,7 @@
 
 # anything bigger than this will generate a special message
 BIG_FILE_SIZE_THRESHOLD = 100 * 2 ** 20  # 100MB
+DEFAULT_SERVER_PORT = int(environ.get('CXG_SERVER_PORT')) if environ.get('CXG_SERVER_PORT') else 5005
```
bkmartinjr: I would use the more concise:   `int(environ.get('CXG_SERVER_PORT', '5005'))`

scripts/await_port
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+import argparse
+import errno
+import socket
+import sys
+import time
+
```
bkmartinjr: please add brief summary of interface & use

dev_docs/developer_guidelines.md
```diff
@@ -9,6 +9,12 @@
...
 **All instructions are expected to be run from the top level cellxgene directory unless otherwise specified.**
 
+### Environment
+
+For all `make` commands, `common.mk` automatically checks whether required environment variables are set and, if they are not set, assigns them default values from `environment.default`.
+
+You can set these environment variables manually with the `export` shell command, as in `export JEST_ENV=debug`.
+
 ## Running test suite
 Client and server tests run on Travis CI for every push, PR, and commit to master on github. End to end tests run nightly on master only.
```
bkmartinjr: it would be nice to add a note on how to run client unit-tests (for front-end devs)

environment.default
bkmartinjr: Actually, further testing reveals that this is a bug.  It prevents the front-end hot loading server from working.   (ie, run `make start-frontend`, and the JS code won't find the cellxgene backend).  Definitely needs to be changed to 5005.

environment.default
```diff
@@ -0,0 +1,6 @@
+# Default environment variables for cellxgene development
+set -a
```
bkmartinjr: I'm unsure why you need these `set` mode changes?   As far as I can tell, you don't rely on this file being executed by bash (which would understand them) or imported into a Makefile except via the `sed` command in common.mk (which ignores them).

So they seem unnecessary.  If they are necessary, the need and purpose is somewhat opaque and needs a comment.

dev_docs/developer_scripts.md
```diff
@@ -118,3 +70,65 @@ gui-spec-windows - creates the initial spec file for windows, do not run unless
...
+
+**Requires**
+* The server to be running. Best way to do this is with `make backend-dev`
+* `make ci` to install the necessary node modules
+
+**Usage** `make start-frontend`
+
+#### backend-dev
+
+**About** This script enables FE developers to run the REST API necessary to back the development server for the front end. It is intended to ensure that the FE developer gets the current version of the backend with a single command and no knowledge of python necessary. It creates and activates a virtual environment and installs cellxgene from the current branch.
```
bkmartinjr: suggest documenting the use of CXG_OPTIONS and provide an example.  Without this, we are going to have a explosion of additional targets like `backend-dev-anno`, `backend-dev-anno-ontology`, etc.

dev_docs/developer_scripts.md
```diff
@@ -118,3 +70,65 @@ gui-spec-windows - creates the initial spec file for windows, do not run unless
...
+**Tips** Developers will probably want to run this in parallel with the node dev server. You can either do this by running each in a separate termanial window or by running `backend-dev` in the background (add an `&` at the end of the command to run in the background: `DATASET=<dataset> make backend-dev &`).
+
+**Breakdown**
+
+| command                                  | purpose                                                   |
+| ---------------------------------------- | --------------------------------------------------------- |
+| python3.6 -m venv cellxgene              | creates cellxgene virtual environment                     |
+| source cellxgene/bin/activate            | activates virtual environment                             |
+| yes \| pip uninstall cellxgene \|\| true | uninstalls cellxgene (if installed)                       |
+| pip install -e ..                        | installs current local version of cellxgene               |
```
bkmartinjr: is the double dot `..` correct?  I thought it was `pip install -e .`?

dev_docs/developer_scripts.md
```diff
@@ -118,3 +70,65 @@ gui-spec-windows - creates the initial spec file for windows, do not run unless
...
+| cellxgene launch                         | launches cellxgene (must supply dataset as last parameter |
```
bkmartinjr: missing closing paren

client/Makefile
```diff
@@ -14,6 +20,57 @@ install:
 build:
 	npm run build
 
+
+# Development convenience methods
+
+.PHONY: server-requirements
+server-requirements:
+	virtualenv -p python3.6 ../venv
```
seve: When running `backend-dev` I get the error message:
> virtualenv -p python3.6 ../venv
The path python3.6 (from --python=python3.6) does not exist

Do I need to downgrade from python 3.7 to 3.6 or is this a different issue?

dev_docs/developer_scripts.md
```diff
@@ -118,3 +70,65 @@ gui-spec-windows - creates the initial spec file for windows, do not run unless
...
+
+### Runner scripts
+
+#### start-frontend
+
+**About** Serve the current client javascript independently from the `server` code.
+
+**Requires**
+* The server to be running. Best way to do this is with `make backend-dev`
+* `make ci` to install the necessary node modules
```
seve: `make ci` is a wipe and install of all required node modules, should it instead be `npm install`?

environment.default
mweiden: The ideas here are the following:
1. If you were to rely on `source`-ing the environment directly, `set -a` will propagate theses variables to subprocesses. The variables in the environment file are then authoritative.
2. If people use this file as a template and create their own environments--say `environment.colin`-- they could `source` it and the changes would be propagated as suggested in (1)

dev_docs/developer_scripts.md
mweiden: True. You're right. It was incorrect in the [original](https://github.com/chanzuckerberg/cellxgene/blame/master/dev_docs/developer_scripts.md#L20). Fixing it here.

environment.default
bkmartinjr: I agree that is what happens when you have bash execute the file.  However, you are not doing that -- you are "grepping" through the file from common.mk.   Unless I'm completely confused, these two lines have zero effect on the execution environment, and can be safely removed.

client/Makefile
```diff
@@ -14,6 +20,57 @@ install:
...
+	source ../venv/bin/activate && \
+		pip install -r ../server/requirements-dev.txt && \
+		yes | pip uninstall cellxgene || true && \
+		pip install -e ..
```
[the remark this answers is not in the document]
bkmartinjr: if you don't use the `npm run` commands, eg, `backend-dev`, you can safely ignore these Makefile targets, and keep doing what you already doing.

client/Makefile
bkmartinjr: different issue.  You can safely keep using Py 3.6, 3.7 or 3.8.

However, if you do use these Makefile targets, you will need to have python 3.6 installed.

client/Makefile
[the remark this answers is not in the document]
mweiden: This replicates the logic of the current [`npm run backend dev`](https://github.com/chanzuckerberg/cellxgene/blob/master/client/package.json#L8) script. So the answer to your question is "Yes," if you choose to use this.

client/Makefile
mweiden: Good catch. Fixing this.

dev_docs/developer_scripts.md
mweiden: The idea here is that you don't always want to update your `package-lock.json` when developing and definitely not while you're running CI. For more see https://docs.npmjs.com/cli/ci.html

environment.default
mweiden: I guess the primary purpose here was to set an example for other people who were creating their own environment files. In this case you need it, since `common.mk` pulls from `environment.default` directly. Point taken, though. I will remove them.

dev_docs/developer_scripts.md
mweiden: True. Added.

Issue #1153, opened 2020-02-13
Annotations have unpredictable undo/redo behavior


mweiden:
### Expected
Expected steps:
1. Delete a category
1. Undo button is active (indicates last action is undoable)
1. Click Undo

Expected results:
* Category is restored

### Current
Steps to reproduce bug:
1. Start from a refreshed page
1. Delete a category

Results:
* ~~Undo button is not active~~
* Category is restored

### Expected
Expected steps:
1. Hover on category options
1. Hover on options of a label
1. Edit label name
1. Submit change
1. Undo button is active (indicates last action is undoable)
1. Click undo

Expected results:
* No edit dialog
* Label is renamed

### Current
Steps to reproduce bug:
1. Hover on category options
1. Hover on options of a label
1. Edit label name
1. Submit change
1. Undo button is active (indicates last action is undoable)
1. Click undo

Results:
* Edit label dialog opens with label name from before edit

### Expected
Expected steps:
1. Hover on category options
1. Add a label
1. Submit label name
1. Undo button is active (indicates last action is undoable)
1. Click undo

Expected results:
* Label is removed
* No new label dialog
* Is redoable

### Current
Steps to reproduce bug:
1. Hover on category options
1. Add a label
1. Submit label name
1. Undo button is active (indicates last action is undoable)
1. Click undo

Results:
* Label is removed
* New label dialog opens
* Is not redoable

Issue #969, opened 2019-10-02
manual labels/annotations need e2e smoke tests


bkmartinjr:
Implement e2e (puppeteer) smoke tests for all new UI features.

colinmegill:
- [ ] make sure the system works with anno turned off (experimental labels flag off)
- [ ] with flags on, do the normal tests pass
- [ ] create a couple category 
- [ ] some labels
- [ ] assign cells to labels and make sure numbers are right
- [ ] delete labels 
- [ ] delete categories
- [ ] rename labels
- [ ] rename categories
- [ ] if a label has cells, select a different set and assign it
- [ ] make sure that the stuff saved in the file is what you expect, was everything saved
- [ ] checks that you load the file correctly on startup 
- [ ] all the above tests, but throw in some undo redo, (add a category, undo, then redo)
- [ ] subset the world, all the same tests, but where universe is not equal to world

mweiden:
Update: I just discovered that annotations features do not play nicely with undo/redo. I've documented the issues here https://github.com/chanzuckerberg/cellxgene/issues/1153

PR #1154, opened 2020-02-13
Add frontend smoke tests for annotations

mweiden:
Partially completes https://github.com/chanzuckerberg/cellxgene/issues/969

This PR covers the following test cases (basically all the frontend tests):
- make sure the system works with anno turned off (experimental labels flag off) — package.json smoke-test-without-annotation
- with flags on, do the normal tests pass — package.json smoke-test-with-annotation
- create a category
- create a label
-  assign cells to labels and make sure numbers are right
- delete a label
- delete categories
- rename a label
- rename a category
- if a label has cells, select a different set and assign it
- subset the world, all the same tests, but where universe is not equal to world

It does not cover the following test cases from https://github.com/chanzuckerberg/cellxgene/issues/969:
- make sure that the stuff saved in the file is what you expect, was everything saved (pytest)
- checks that you load the file correctly on startup (pytest)
- all the above tests, but throw in some undo redo, (add a category, undo, then redo)

Note the undo/redo test is blocked by https://github.com/chanzuckerberg/cellxgene/issues/1153


mweiden:
@bkmartinjr when you review this, please note f0bd970. Motivation: when running the test the annotations fixture can update, which causes the test to brake.

bkmartinjr:
> @bkmartinjr when you review this, please note [f0bd970](https://github.com/chanzuckerberg/cellxgene/commit/f0bd970bb2b9b7a3ec6a5c15bebcab9e7f2684a5). Motivation: when running the test, autosave can update the annotations fixture, which causes the subsequent runs of the test to break.


@mweiden - what do you mean by "fixture"?   Also, down the road, if saving is completely disabled, how will you do a full test that the annotations make it into the server-side correctly (ie, are saved into the CSV)?

mweiden:
> what do you mean by "fixture"?

@bkmartinjr I'd call `example-dataset/pbmc3k-annotations.csv` and `example-dataset/pbmc3k.h5ad` "test fixtures" since they are both starting datasets used in testing.

> if saving is completely disabled, how will you do a full test that the annotations make it into the server-side correctly (ie, are saved into the CSV)?

This is a good question and I share the concern. Something here needs to bend: even if we construct the test categories/labels before each test, reproducibility will be a problem since the server will periodically save state.

One option: have the script that runs the test reset any altered test fixtures or remove any saved annotations state (in the case of constructing them on the fly).

bkmartinjr:
> One option: have the script that runs the test reset (git checkout ...) any altered test fixtures or remove any saved annotations state (in the case of constructing them on the fly).

Another option: have the test make a copy of the base file to /tmp, and start the server pointing at the temp file.  Probably need to generate a unique tmpfile name in case the test gets run more than once, concurrently.

mweiden:
@bkmartinjr I like it. I'll give that a shot.

client/src/globals.js
```diff
@@ -1,6 +1,9 @@
 import { Colors } from "@blueprintjs/core";
 import { dispatchNetworkErrorMessageToUser } from "./util/actionHelpers";
 
+/* flags for testing */
+export const autosaveEnabled = process.env.AUTOSAVE_ENABLED ? process.env.AUTOSAVE_ENABLED === "true" : true;
```
bkmartinjr: One last-minute thought:  I think we will need to remove this at some point, as I can't see a world where we (literally) don't save a user's annotations.

Can you give it a name more suggestive of its temporary "work around" nature?

Eg, "tmpWorkAroundAutosaveEnabled"?

mweiden: I'm just going to implement the `/tmp` file idea and push it to this PR here in a bit.

PR #1157, opened 2020-02-16
anndata X indexing & version compatibility improvements

### LINKED ISSUE #917: remove temp pin of h5py when regression fixed

### LINKED ISSUE #1142: AnnData 0.7
Hello!

I've been running into some trouble with using cellxgene along with AnnData 0.7. I'm a little surprised by this since I didn't seem to run into issues using pre-release version of anndata, and files it wrote. Issues often look like:

```
[2020-02-05 15:13:31,778] ERROR in app: Exception on /api/v0.2/config [GET]
TypeError: Object of type Version is not JSON serializable
```

* Any idea what's up? Did we break something?

bkmartinjr:
This PR corrects our historical and incorrect use of the anndata 'X' parameter. Discussion in #1142 and theislab/anndata#317.  Briefly,
* adata.X is not guaranteed to be array-ish, but any slice of adata[].X or adata.X[] is array-ish
* when anndata<=0.7, opening an h5ad with sparse X, in backed='r' mode, has performance problems on any slice (will read the entire X, rather than just the slice)

As a consequence, the backed mode was not particularly useful for a number of large, sparse datasets.   Prior to this PR, we had a MatrixProxy that attempted to work around the backed performance problem, but at the cost of utilizing private interfaces.   With the advent of anndata==0.7.*, the backed performance problem is largely resolved (caveat: off-axis slicing is still slow).

This PR makes the following changes:
* The MatrixProxy is removed, and all use of adata.X is  replaced with standard X indexing (adata.X[...]).
* h5py is unpinned, as 0.7 is now compatible with h5py 2.10.0
* A warning is emitted if the user uses backed mode with anndata<=0.7, and recommends an update

In addition, this PR corrects two other issues found during compatibility testing:
* if the dataset location is specified with a URL, rather than a local file name, anndata==0.7 `read_h5ad()` will fail because cellxgene did not preserve the .h5ad file suffix (other versions of anndata appear to work fine without the suffix).  This file suffix is now preserved, working around this anndata 0.7 quirk.
* the minimum anndata dependency is now >=0.6.20 (it was 0.6.2).  This resolves a long-standing incompatibility with modern scipy and anndata (fixed in anndata==0.6.20).   cellxgene hasn't worked with 0.6.2 for at least a year.

Note: h5py==2.10.0 and anndata<=0.7 remain incompatible, and this is unlikely to be fixed. However, if a user has updated their h5py to the most recent version, it seems reasonable to also require that they update the anndata module.

Unit tests manually validated with anndata 0.6.22post1, 0.7.1, 0.7 and 0.6.20.

Fixes #917 
Fixes #1142


mweiden:
@bkmartinjr re. the below:

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -370,7 +382,12 @@ def _is_valid_layout(self, arr):
 
     @requires_data
     def _validate_data_types(self):
-        if sparse.isspmatrix(self.data.X) and not sparse.isspmatrix_csc(self.data.X):
+        # The backed API does not support interogation of the underlying sparsity or sparse matrix type
```
mweiden:
```suggestion
        # The backed API does not support interrogation of the underlying sparsity or sparse matrix type
```

server/app/scanpy_engine/scanpy_engine.py
```diff
@@ -370,7 +382,12 @@ def _is_valid_layout(self, arr):
...
+        # Fake it by asking for a small subarray and testing it.   NOTE: if the user has ignored our
+        # anndata <= 0.7 warning, and run in backed mode with a large sparse dataset, this "small"
```
mweiden: `and run in backend mode` doesn't quite parse

server/app/util/matrix_proxy.py
```diff
@@ -1,425 +0,0 @@
-import abc
```
mweiden: Fantastic that you can remove all this code!

server/app/scanpy_engine/scanpy_engine.py
bkmartinjr: it is 'backed', not 'backend'.    In the anndata API, `backed` is short-hand for file-backed. 

I'll expand on the wording.

Issue #1155, opened 2020-02-14
CI improvements


bkmartinjr:
Suggested improvements to our CI tests, giving us a broader set of compatibility tests _and_ faster per-commit tests.

- [x] don't run all the compatibility tests on each commit, thereby giving a faster "quick test".   For example, we don't need Python 3.6 and 3.7 on every commit.
- [x] Add a nightly set of tests, for broader compatibility, both unit & smoke tests.   Suggestions:
   - [x] Python 3.6. 3.7, 3.8
   - [x] commonly used anndata released versions:  0.6.22post1, 0.7.1
   - [x] docker build
- [x] Tests to ensure ecosystem incompatibilities are detected before they are released:
   - [x] most recent cellxgene release vs. anndata master
   - [x] cellxgene master vs. anndata master

We may want to expand the compatibility matrix over time.

PR #1158, opened 2020-02-18
Simplify CI checks on git push, add nightly checks on cron

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1155

## Need

`cellxgene` developers need:
1. CI checks that run after a `git push` to run quickly
1. A broader set of compatibility checks (for now, we need to test different versions of `anndata` and `python`)

## Approach

* Move CI tests from Travis to Github Actions
* Split the test suites into tests on `push` and compatibility tests that run ~nightly~ Tuesdays at 8 UTC, 12AM PT
* Added dependency caching to tests on `push`
* Add tests for different versions of `anndata` including foward-looking tests that test the current version of `cellxgene` against the `HEAD` of `anndata` and `cellxgene` `HEAD` against `anndata` `HEAD` to give advanced warning of upcoming release incompatibilities

## Results

* Tests on push run ~33% faster
* I _think_ the ~nightly~ compatibility test configurations are correct; so the tests indicate that we are currently broken WRT newer versions of `anndata`


.github/workflows/nightly_tests.yml
```diff
@@ -0,0 +1,99 @@
+name: Nightly Tests
+
+on: [push]
+#  schedule:
+#    cron: '* 8 * * * *'
```
mweiden: Note, `push` was needed to debug here. I'll re-enable the cron before we merge this.

bkmartinjr:
@mweiden - when the cron jobs are enabled, how will we detect that they have failed?  Do they notify the core team, is there a badge, or?

mweiden:
@bkmartinjr I believe the commit will show up as red after the nightly test runs and emails are sent on test failure, if you're subscribed to github notifications. Also I've included badges https://github.com/chanzuckerberg/cellxgene/pull/1158/commits/d4dd7e6ad7cd1c83acf54218c217cfa16b409f35

PR #1175, opened 2020-02-25
Add undo/redo smoke tests for annotations

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/969

... also refactor the tests for DRY.

Covers the remaining tests outlined in #969:
* undo/redo category creation
* undo/redo category deletion
* undo/redo category edit
* undo/redo label creation
* undo/redo label deletion
* undo/redo label edit


mweiden:
Clearly `codecov` is not picking up our smoke tests 🤦‍♂

bkmartinjr:
> Probably because the code is compiled and run through the development server during smoke tests...

Why not use the prod (CLI) server?  That would certainly be a better test, as it would not suffer from differences in backend prod/dev behavior.

mweiden:
> Why not use the prod (CLI) server? That would certainly be a better test, as it would not suffer from differences in backend prod/dev behavior.

Issue #1135, opened 2020-01-29
Opening a dataset on hosted should default to coloring by the first category in LSB


jenzopr:
Hi cellxgene-Team!

A small, but effective addition would be to allow a cell label to be colored by default. This would aid a user that visits the (hosted) instance for the first time by directly pointing him/her to an important aspect of the data (e.g. coloring by a _celltype_ label). 

Along this line, the toggle to show labels (btw an amazing addition from `0.14`!) could be allowed to be switched on by default as well.

Let me know what you think! 

Best,
Jens

colinmegill:
Love this idea!

bkmartinjr:
Would this be better solved by picking a better start-up default for the UI?  In other words, rather than starting the app with "nothing selected, no labels shown", start the app with "color by first category, with labels shown"?

What heuristic could be used to pick the category used?  (first in the category list?)

I'm reluctant to adding a CLI parameter to do this, as it is just another hard-to-discover feature that many users will not use.

jenzopr:
@bkmartinjr that's a good idea! In Python, the column order of `adata.obs` can be specified - so colouring by the first column in the list would be fine for me (reminder: there needs to be some logic for a numeric column being first in the list).

bkmartinjr:
@mweiden - please chat with @liaprins-czi on UX design before we implement

liaprins-czi:
In the initial request, it sounded like the idea was to be able to color a single celltype label, rather than coloring by the entire category. Just curious if coloring by an entire category, wherein each label gets its own color, meets the original need? If not, can you explain the need for a single celltype being colored alone, @jenzopr ? Thank you!

mweiden:
@bkmartinjr definitely

colinmegill:
@liaprins-czi @jenzopr I interpreted the request as, in visual form: "one of the paint swatches / color by category buttons should be 'on' by default when a user loads the app up, and I want to be able to specify which one."

![image](https://user-images.githubusercontent.com/1770265/73677390-fe96e100-466a-11ea-9240-69ef26e80bdd.png)

jenzopr:
Pardon my "semantic" inaccuracy @liaprins-czi. I indeed meant coloring by an entire category by default. Thanks for clarification!

liaprins-czi:
Thanks @jenzopr ! I just wanted to be sure :)

liaprins-czi:
[...]

PR #1177, opened 2020-02-26
Color the scatterplot with the top left sidebar category by default

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1135

Also
* Clear up some unused variables
* Remove an unused parameter from `selectableCategoryNames`

### DIFF
--- a/client/__tests__/e2e/e2eAnnotations.test.js
+++ b/client/__tests__/e2e/e2eAnnotations.test.js
@@ -200,7 +200,7 @@ describe.each([
 
   async function assertCategoryExists(categoryName) {
     const result = await utils.waitByID(`${categoryName}:category-expand`);
-    expect(await result.evaluate(node => node.innerText)).toBe(categoryName);
+    expect(result).not.toBeNull()
   }
 
   async function assertCategoryDoesNotExist(categoryName) {
@@ -213,7 +213,7 @@ describe.each([
     expect(category).not.toBeNull();
     await cxgActions.expandCategory(categoryName);
     const previous = await utils.waitByID(`categorical-value-${categoryName}-${labelName}`);
-    expect(await previous.evaluate(node => node.innerText)).toBe(labelName);
+    expect(previous).not.toBeNull()
   }
 
   async function assertLabelDoesNotExist(categoryName, labelName) {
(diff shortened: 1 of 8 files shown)


client/__tests__/e2e/e2eAnnotations.test.js
```diff
@@ -85,11 +85,11 @@ describe.each([
   test("create a label", async () => {
     await utils.hoverOn("cluster-test:see-actions");
     await utils.clickOn("cluster-test:add-new-label-to-category");
-    await utils.typeInto("cluster-test:new-label-name", "test-label-name");
+    await utils.typeInto("cluster-test:new-label-name", "test-label");
     await utils.clickOn("cluster-test:submit-label");
     await cxgActions.expandCategory("cluster-test");
-    const result = await utils.waitByID("categorical-value-cluster-test-test-label-name");
-    expect(await result.evaluate(node => node.innerText)).toBe("test-label-name");
```
mweiden: Adjusting this because the length of the label name caused it to be condensed in the sidebar and the equality test assertion to fail.

client/src/actions/index.js
```diff
@@ -135,7 +135,15 @@ const doInitialDataLoad = () =>
       Step 3 - load everything else
       */
       await obsAnnotationFetchAndLoad(dispatch, schema, universe);
-
+      const firstCategory = (schema.schema.annotations?.obs?.columns || []).find(
```
bkmartinjr: I don't think this is an appropriate test to find "first", as it does not use the same logic as the category component.  It will miss non-categorical annotations which are displayed as categorical (ie, string, bool), and there is currently no guarantee that the schema order will match the render order.

Please see src/components/categorical/categorical.js for the actual logic:
```
    /* all names, sorted in display order.  Will be rendered in this order */
    const allCategoryNames = ControlsHelpers.selectableCategoryNames(
      schema,
      ControlsHelpers.maxCategoryItems(config)
    ).sort();
```

mweiden: True. I realized this after putting up the PR and then tagged it with `[WIP]`. Will figure this out in a bit.

mweiden:
We decided during the 2020-03-02 planning meeting to hold off implementing this feature this way. We'll revisit this as part of implementing hosted cellxgene, where server configuration or query string parameters will make implementing the feature more natural.

PR #1179, opened 2020-02-26
Improve label picking

### LINKED ISSUE #1130: Ontology auto-suggest enhancements
The workaround in `ontology-patch-1` is fine for a temporary workaround, but we should not lose track of the ideal design.

- User needs to easily switch to creating their own label if they start out wanting an ontology label but can’t find the one they want.
- Also, editing a label that was originally set as an ontology should allow the user to choose a different ontology, or to name it manually whatever they want (and same for originally-manually-named labels).

Reference from #996 for the design:

Here is a doc of the guidelines @sidneymbell and I have been thinking for this work:
https://drive.google.com/open?id=1-7zeBmMzjfZA0bSZVv2-D_AWbvvVEuorsID65-2UFGw

Here is a link to the Figma with explorations:
https://www.figma.com/file/j0OrP1bd1LQbZOWPbKSMlI/Annotations-ontologies?node-id=40%3A0

"Minimal" prototype without hierarchy:
https://www.figma.com/proto/j0OrP1bd1LQbZOWPbKSMlI/Annotations-ontologies?node-id=9%3A0&viewport=-87%2C671%2C0.1471354216337204&scaling=min-zoom

More "robust" prototype with hierarchy:
https://www.figma.com/proto/j0OrP1bd1LQbZOWPbKSMlI/Annotations-ontologies?node-id=45%3A3&viewport=-9075%2C401%2C0.5&scaling=min-zoom

Here are some Blueprint components which might be helpful for surfacing hierarchy within the ontology suggestions:
- [Tree](https://blueprintjs.com/docs/#core/components/tree)
- [Breadcrumbs](https://blueprintjs.com/docs/#core/components/breadcrumbs)
- [Overflow list](https://blueprintjs.com/docs/#core/components/overflow-list)
- [Divider](https://blueprintjs.com/docs/#core/components/divider)

bkmartinjr:
Unify annotation label/category name entry when ontologies are enabled.   This provides ontology term suggestions in both category and category label prompts (both new and edit).

Changes in this PR:
* New component, `LabelInput`, lets user enter a category or label name.   When ontologies enabled, it provides suggestions from the ontology terms; when disabled, it is a standard input component.
* Error prompting for label input validation revamped to be compatible with the drop-down suggestion popover used by LabelInput
* All dead code, related to the separate implementation "pick label from ontology" removed.
* Ontology-based suggestions are available from label add, label edit, category add, category edit, with sublime-like fuzzy search through the ontology terms.
* General lint pass while I'm in the files...

image

image

Fixes #1130


client/src/components/categorical/labelInput.js
```diff
@@ -0,0 +1,176 @@
...
+      const { newLabelMessage } = this.props;
+      return (
+        <MenuItem
+          icon="flag"
+          active={modifiers.active}
+          disabled={modifiers.disabled}
+          key={queryResult.target}
+          onClick={handleClick}
+          text={<em>{queryResult.target}</em>}
+          label={newLabelMessage || "New label"}
```
colinmegill: This needs to be set conditionally (seems like perhaps another file is missing passing down `newLabelMessage`) — `New label` --> `Category not in ontology` or `New category`

image

bkmartinjr: good catch - change pushed.

Issue #1171, opened 2020-02-24
Don't remove genes from workspace upon reset


sidneymbell:
cc @aopisco 
"STOP LOSING THE GENE LIST IF I RESET" (currently using 'undo undo undo' to get back to `universe` instead of using `reset` because of this)

This is consistent with the proposed gene sets work in #1069

aopisco:
this would rather be an enhancement but eventually would be good to be able to select which genes to maintain. the scenario here is we want to keep all the manually inputted genes but DGE genes might no longer be useful; on the first version reset could remove all DGE (particularly when those start being ~100) and keep the manually inputted genes; on the second version the user could select any DGE of interest to keep on its "gene exploratory analysis" list

seve:
@colinmegill Should we consider persisting the redux store in local storage as a solution for this until we address gene sets?

bkmartinjr:
> @colinmegill Should we consider persisting the redux store in local storage as a solution for this until we address gene sets?

The right solution would be to refine the UX design of reset so it has a better set of side effects.   There is no need to hack around it - it is our current UI design that is deliberately deleting the gene list - we can just change it to do what we want.

colinmegill:
The reason we did this, initially, is that once the user has subset, `cell set 1` and `cell set 2` for diffexp may no longer reflect cells on screen in `world` 

The fix for this is to remove `cell set 1` and `cell set 2` in favor of accomplishing diffexp via label selection. 

This is captured in #729 and also facilitates #99

bkmartinjr:
@colinmegill - are you advocating for not doing a short term fix? 

I find the current behavior _worse_ than having "disconnected genes" as it results in information loss.

colinmegill:
I'm fine with not clearing diffexp as a stopgap

mweiden:
I've spoken with Lia about this and we came up with the following desired behavior:
1. If you add a gene and then press "Undo" that removes the gene. If you then press "Redo" the gene is added back.
2. If you add a gene and then press "Reset," the gene is not removed.

Notice that both mirror the undo/redo/reset behavior of adding a category on the left-hand side.

PR #1183, opened 2020-02-29
Undo selection appends genes from differential expression to user gene list

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1171

## Need
When a user performs a differential expression from within a sub-selection (world) of the data and then resets the selection to all cells (universe), the differential expression results are no longer valid.

## Approach
* When the selection is reset, move the top (maxUserDefinedGenes - len(userDefinedGenes) from the differential expression results to the
list of user defined genes
* Raise maxUserDefinedGenes to 25 to give users more room and accommodate the extra genes transferred in from differential expression


mweiden:
Second option for button icons:

Screen Shot 2020-02-28 at 11 17 55 PM

colinmegill:
@mweiden HA! Excellent set of second button choices :) significant improvement on the first.

client/src/components/brushableHistogram/index.js
```diff
@@ -523,6 +523,7 @@ class HistogramBrush extends React.PureComponent {
           </span>
           <span
             data-testclass="brushable-histogram-field-name"
+            data-testid={`${field}:field-name`}
```
bkmartinjr: `:field-name` is seems overly generic - we may have other fields in the future.  Maybe something more descriptive?

client/src/components/menubar/index.js
```diff
@@ -362,30 +316,43 @@ class MenuBar extends React.Component {
...
+          <Tooltip
+            content="Subset to currently selected cells and associated metadata"
+            position="bottom"
+            hoverOpenDelay={globals.tooltipHoverOpenDelay}
           >
-            <Icon icon="double-chevron-down" />
-          </AnchorButton>
-        </Tooltip>
-        <ButtonGroup style={{ marginRight: "10px" }}>
+            <AnchorButton
```
bkmartinjr: suggestion - factor this out into a sub-component ans file like Clip or InformationMenu

client/src/reducers/controls.js
```diff
@@ -74,9 +75,12 @@ const Controls = (
       };
     }
     case "reset World to eq Universe": {
+      const { userDefinedGenes, diffexpGenes } = state;
       return {
         ...state,
-        resettingInterface: false
+        resettingInterface: false,
+        userDefinedGenes: [].concat(userDefinedGenes, diffexpGenes).slice(0,globals.maxUserDefinedGenes),
```
bkmartinjr: this removes the most recently added genes -- is that intentional?   I would have imagined we would remove the least recently added genes.

@liaprins-czi - thoughts?

bkmartinjr:
> I think they'd still expect the genes to stay no matter what they do.

No question you are right.

@mweiden - how about changing the logic:
* you can't manually add more user defined genes when total number exceeds the global limit
* but...you can still add them via subset/reset

This leaves a loophole, where a clever user could continuously add genes, but I don't think we really care.  Worst case, they run out of memory in their browser. 

Alternatively, up the global limit to something much larger, etc, 100.

client/src/reducers/controls.js
liaprins-czi: Yes I agree it should keep the most recently added and sacrifice the "oldest" genes instead... good catch

client/src/components/brushableHistogram/index.js
mweiden: Ah! Turns out I didn't use this anyway. Deleting it.

client/src/reducers/controls.js
mweiden: Good points. @bkmartinjr I've implemented the solution you mentioned below: don't throw out any user-added genes and have a max gene limit that only slices off diffexp genes.

mweiden:
> - you can't manually add more user defined genes when total number exceeds the global limit
> - but...you can still add them via subset/reset
>
> Alternatively, up the global limit to something much larger, etc, 100.

sidneymbell:
This looks awesome!

Re: number of genes, this relates to the request from multiple users to be able to specify the number of differentially expressed genes returned. Is there a strong reason to cap the number of genes so tightly? I'd be in favor of Bruce's suggestion to increase it to, say, 100. 

If the user hits the limit, yes, removing the "oldest" genes first makes sense to me.

mweiden:
@sidneymbell responses inline below:

> Is there a strong reason to cap the number of genes so tightly? I'd be in favor of Bruce's suggestion to increase it to, say, 100.
I think you're conflating number of genes returned by differential expression with the number of genes displayed on the sidebar. I did implement a limit on the total number of genes that can be displayed on the sidebar (`globals.maxGenes`=100). Further, I increased the number of genes users can add manually to the sidebar (`globals.maxUserDefinedGenes`=25). However, the number of results returned from each run of differential expression is still 10.

> [Users request] to be able to specify the number of differentially expressed genes returned
Agree that this should be an option. Let's follow up on this as part of the sidebar refactor.

> If the user hits the limit, yes, removing the "oldest" genes first makes sense to me.
With the changes in 7e2ce04, no genes manually added by the user will be removed.

PR #1186, opened 2020-02-29
experimental re-embedding

### LINKED ISSUE #1047: re-projecting cells after subsetting
I have been using cellxgene for the TabulaSapiens dataset. I am having difficulty balancing having all the cells in one object v.s. being able to visualize small subsets of cells. Distinct clusters can be squished together on 2D map as an artifact of too much other variation. If I supply cellxgene with the full latent space (10-30 dimensional), then cellxgene could in theory re-runs UMAP to generate 2D maps that only contain the cells that are subsetted, and the visualization can make much better use of the space available on the projection. 

I would really appreciate this functionality if possible!

### LINKED ISSUE #1150: feature: re-embed selected cells
Subset of #280 

Add a feature to allow users to create a new embedding/layout on a subset of the cells in a dataset.   Basic story:
* user selects sub-population and does a subset operation in the UI
* user says "make me a new embedding, of type umap" (may be more options in future to pick algo and params, but initially just super simple)
* new embedding created on backend (may be async, ie, may take a while).  Error generated if it takes too long.
* new embedding given an automatically generated name (eg, "umap_20200214"), downloaded and made visible.  Embedding will now be in the embedding menu.

Open issues/work:
* We will need to assign some default coordinate to cells not in the subset, in case the user resets back to all cells, and wants to display the embedding.   (eg., all cells not in the subset are assigned an embedding position if [-1,-1]).
* We will also need to identify the default algorithm and parameters.   @sidneymbell @aopisco it would be great if you could collaborate on a notebook for this.

---

From @aopisco:
> A couple of thoughts here:
> 1. The most common user case will be the user wanting to subset a group of cells.
> 2. The user selects the cells they're interested in and click to subset. That should trigger re-embedding by running sc.tl.umap(adata_subset).
> 3. If the user is doing annotations using cellxgene, then the user can select a smaller group within the subset and annotate those and so one
> 4. At the moment because there is no re-embedding after subsetting it's impossible in many applications (like when there are T cells and NK cells) to get a finer look at the data which is limiting the potential of the subsetting function

bkmartinjr:
This PR adds an _experimental_ feature allowing the user to request re-embeddings of any subset of the dataset.   To use:
* must have scanpy module installed (caveat: only tested with latest, v 1.4.5.1)
* run `cellxgene launch --experimental-enable-reembedding mydata.h5ad`
* select a subset of the cells, and click the subset button
* click the re-embed button
* when notified that the embedding is ready, you can select it on the embeddings drop-down menu
* when you reset the data subsetting, all re-embeddings of the subset data are deleted

Important caveats:
* requires manual installation of scanpy - this module is not a cellxgene dependency, and the feature will not work without it.
* the feature is incompatible with `--backed` mode (and an error will be raised if both are specified)
* scanpy generates a lot of noisy, trash warnings on the console which can be ignored
* it is slow, and gets rapidly slower as the number of selected cells expands

Importantly, as with all experimental features, this may not be added as a supported feature, and may be removed in a future release.

Fixes #1150 
Fixes #1047 

Other minor fixes in this PR:
* the /config route was not correctly returning availability of many routes.
* the subset/unsubset buttons in the menu bar incorrectly set active state rather than disabled state, and were inconsistent with the rest of the toolbar.


server/cli/launch.py
```diff
@@ -317,6 +325,9 @@ def launch(
         except DatasetAccessError as e:
             raise click.ClickException(str(e))
 
+        if experimental_enable_reembedding and matrix_data_loader.matrix_data_type() != MatrixDataType.H5AD:
```
bmccandless: Not important for this PR, but I think it would be quite easy to get this to work with TileDb.   Anndata has a constructor that takes in X, obs and var arrays.   So, it's just a matter of how you initialize the temporary AnnData object before you compute the umap.

bkmartinjr: Agree - I have an alternate thought on how to do it, but I agree it should be easily in principle (I'm not sure we want or need to use ScanPy/Anndata - I think we just integrate the underlying embedding module Scanpy wraps).

server/compute/scanpy.py
```diff
@@ -0,0 +1,49 @@
...
+    adata = adata[obs_mask, :].copy()
+
+    for k in list(adata.obsm.keys()):
+        del adata.obsm[k]
+    for k in list(adata.uns.keys()):
+        del adata.uns[k]
+
+    sc.pp.pca(adata, zero_center=None, n_comps=min(adata.n_obs - 1, 50), **pca_options)
+    sc.pp.neighbors(adata, **neighbors_options)
+    sc.tl.umap(adata, **umap_options)
```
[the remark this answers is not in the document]
bkmartinjr: Given the code on line 40-41, why is this change needed?

moritzschaefer:
It seems this feature got eliminated from the code base at some point. Is there merely a new name and I don't find it or is this feature not supported? (I am especially interested in the feature in the explorer app (`cellxgene launch`). 
Thanks for any support

PR #1189, opened 2020-03-02
smoke-test centroids + graph overlay layer

seve:
This PR adds two tests to verify that centroid labels are generated/displayed and that the graph overlay layer correctly transforms the labels

### DIFF
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -281,3 +281,20 @@ describe("ui elements don't error", () => {
     await page.evaluate("window.scrollBy(0, 1000);");
   });
 });
+
+describe("centroid labels", () => {
+  test.only("labels are created", async () => {
+    console.log(await utils.clickOn("centroid-label-toggle"));
+    Object.keys(data.categorical).forEach(async label => {
+      console.log("button", await utils.clickOn(`colorby-${label}`));
+      // console.log(button);
+      const labels = await utils.getAllByClass("centroid-label");
+      console.log(labels);
+
+      // expect(labels).toHaveLength(Object.keys(data.categorical[label]).length);
+      console.log(labels.length);
+
+      expect(labels).toHaveLength(2);
+    });
+  });
+});
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -316,8 +316,6 @@ describe("graph overlay", () => {
     const initialCoordinates = await utils.getElementCoordinates(
       `${categoryValue}-centroid-label`
     );
-    console.log("panCoords", panCoords);
-    console.log("initialCoordinates", initialCoordinates);
     await cxgActions.drag(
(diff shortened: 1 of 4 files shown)


mweiden:
@seve you may need to rebase your branch against master and resolve merge conflicts ⬇️

client/__tests__/e2e/e2e.test.js
```diff
@@ -309,6 +309,58 @@ describe("ui elements don't error", () => {
...
+  });
+});
+
+describe("centroid labels", () => {
+  test("labels are created", async () => {
+    await utils.clickOn("centroid-label-toggle");
+    const labels = Object.keys(data.categorical);
+    /* eslint-disable no-await-in-loop */
+    // Toggle colorby for each category and check to see if labels are generated
+    for (let i = 0, { length } = labels; i < length; i += 1) {
```
mweiden: nit: I think here you can use a [`for-of`](https://stackoverflow.com/questions/3010840/loop-through-an-array-in-javascript) statement and make things a little cleaner

client/__tests__/e2e/puppeteerUtils.js
```diff
@@ -231,53 +256,82 @@ export const cellxgeneActions = puppeteerPage => ({
...
+    await puppeteerUtils(puppeteerPage).clickOn(
+      `${categoryName}:delete-category`
+    );
   },
 
   async createLabel(categoryName, labelName) {
     await puppeteerUtils(puppeteerPage).clickOn(`${categoryName}:see-actions`);
-    await puppeteerUtils(puppeteerPage).clickOn(`${categoryName}:add-new-label-to-category`);
-    await puppeteerUtils(puppeteerPage).typeInto(`${categoryName}:new-label-name`, labelName);
+    await puppeteerUtils(puppeteerPage).clickOn(
```
mweiden: What line width are you using in Prettier? In the python code, we've been using 120 characters.

seve: AFAIK we use the default prettier settings which I see as 80.  Maybe in the future we can add a prettier config file

client/__tests__/e2e/e2e.test.js
seve: IIRC: `for of`'s don't work nicely with async/await

bkmartinjr: they are also heinously slow.   We strongly dislike for..or in any situation where performance matters even a bit.

This code matches our current conventions.

client/__tests__/e2e/puppeteerUtils.js
bkmartinjr: AFAIK, we use the default prettier config (i,e, we implicitly have a config - the default).   I believe 80 is the default line length.

client/__tests__/e2e/e2e.test.js
mweiden: Ah! Ok, good to know.

PR #1190, opened 2020-03-03
 annotations cli updates

### LINKED ISSUE #1066: Tracking issue: annotations CLI config
__For discussion. cc @liaprins-czi__

## Disable annotations flag
- [x] Swap logic of the `--annotations` flag  
Current behavior:  
`--experimental-annotations      Enable user annotation of data.  [default: False]`
Desired behavior:  
`--disable-annotations      Disable in-app annotation of data. [default: False]`  

## Naming  
- [x] Rename `--annotations-file` to `--annotations-input-file`  
- [x] Remove `experimental` from all args 

## Defaults
- [x] If no input file is specified, prompt the user for a name on startup (including their unique hash)
- [ ] If no output dir is specified, default to (1) h5ad directory if local and available; (2) cwd if not

## Misc
- [x] Make sure the docs are updated 
- [x] Update tests!

sidneymbell:
Fixes #1066


docs/posts/annotations.md
```diff
@@ -36,10 +35,7 @@ There are two options for updating draft annotations.
 
 ### Autodetect annotations csv
 
-Cellxgene can automatically find and reload your draft annotations in editable mode like so:  
-```
-annotations-directory$ cellxgene launch mydata.h5ad --experimental-annotations
-```
+Cellxgene will automatically find and reload your draft annotations in editable mode like so:
```
mweiden:
```suggestion
Cellxgene will automatically find and reload your draft annotations in editable mode.
```
No example given. Or did you mean to add one?

docs/posts/annotations.md
```diff
@@ -13,16 +13,15 @@ We are _piloting_ a new feature in cellxgene that enables users to create and ed
...
 ```
 
 To preserve data provenance, **`cellxgene` does not alter the input h5ad file**.  Rather, newly-created annotations are saved in a specified CSV file:  
 ```
 annotations-directory/name-########.csv
 ```
 
-- The default `annotations-directory` is  your current working directory (i.e., the directory you were in when you started cellxgene).
-
+- The default `annotations-directory` is the directory where `mydata.h5ad` is saved; if we can't find that directory (e.g., you're loading an h5ad from a URL), it falls back to your current working directory (i.e., the directory you were in when you started cellxgene).
```
bkmartinjr: I would also point out that it will default to CWD if it can't write to the location of the H5AD.

Eg., `(e.g., you're loading an h5ad from a URL, or the location of the h5ad is not writable)`

docs/posts/annotations.md
```diff
@@ -36,10 +35,7 @@ There are two options for updating draft annotations.
 
 ### Autodetect annotations csv
 
-Cellxgene can automatically find and reload your draft annotations in editable mode like so:  
-```
-annotations-directory$ cellxgene launch mydata.h5ad --experimental-annotations
-```
+Cellxgene will automatically find and reload your draft annotations in editable mode.
```
bkmartinjr: you should describe the search path precedence.   remember that we look in multiple places if --annotations-input-file or --anotations-output-dir is not specified, which means that we may find _multiple_ annotations files.  Which is used?

docs/posts/annotations.md
sidneymbell:
```suggestion
- The default `annotations-directory` is the directory where `mydata.h5ad` is saved; if we can't find that directory (e.g., you're loading an h5ad from a URL or the location of the h5ad is not writable), it falls back to your current working directory (i.e., the directory you were in when you started cellxgene).
```

server/cli/launch.py
```diff
@@ -26,14 +26,14 @@
...
+        "--disable-annotations",
         is_flag=True,
         default=False,
         show_default=True,
-        help="Enable user annotation of data.",
+        help="Disable user annotation of data.",
     )
     @click.option(
-        "--experimental-annotations-file",
+        "--annotations-input-file",
```
bkmartinjr: The name is misleading - it is both the input and output file.  It will be overwritten.

I think the original name `--annotations-file` is clearer.  And definitely more accurate.

CC: @liaprins-czi

server/cli/launch.py
```diff
@@ -42,7 +42,7 @@ def annotation_args(func):
         "Incompatible with --annotations-output-dir.",
     )
     @click.option(
-        "--experimental-annotations-output-dir",
+        "--annotations-output-dir",
```
bkmartinjr: Similar to above, given that we also _load_ data from this directory, I suggest it be renamed `--annotations-directory`

CC: @liaprins-czi

server/cli/launch.py
```diff
@@ -51,14 +51,14 @@ def annotation_args(func):
         "Incompatible with --annotations-input-file.",
     )
     @click.option(
-        "--experimental-annotations-ontology",
+        "--annotations-ontology",
```
bkmartinjr: This should remain experimental, as we discussed and decided, it is not yet ready for non-experimental status.

server/cli/launch.py
```diff
@@ -51,14 +51,14 @@ def annotation_args(func):
...
         is_flag=True,
         default=False,
         show_default=True,
         help="When creating annotations, optionally autocomplete names from ontology terms.",
     )
     @click.option(
-        "--experimental-annotations-ontology-obo",
+        "--annotations-ontology-obo",
```
bkmartinjr: should remain experimental.

server/cli/launch.py
```diff
@@ -271,13 +271,13 @@ def launch(
...
-    experimental_annotations_file,
-    experimental_annotations_output_dir,
+    disable_annotations,
+    annotations_input_file,
+    annotations_output_dir,
     backed,
     disable_diffexp,
-    experimental_annotations_ontology,
-    experimental_annotations_ontology_obo,
+    annotations_ontology,
```
bkmartinjr: should remain experimental.

server/cli/launch.py
```diff
@@ -342,32 +342,32 @@ def launch(
...
-            click.echo("Warning: --experimental-annotations-file ignored as --annotations not enabled.")
-        if experimental_annotations_output_dir is not None:
-            click.echo("Warning: --experimental-annotations-output-dir ignored as --annotations not enabled.")
-        if experimental_annotations_ontology:
-            click.echo("Warning: --experimental-annotations-ontology ignored as --annotations not enabled.")
-        if experimental_annotations_ontology_obo is not None:
-            click.echo("Warning: --experimental-annotations-ontology-obo ignored as --annotations not enabled.")
+    if disable_annotations:
+        if annotations_input_file is not None:
+            click.echo("Warning: --annotations-input-file ignored as annotations are disabled.")
```
bkmartinjr: I think these should be hard errors, not warnings, as they imply a misconfiguration or other serious misunderstanding.  Warnings are easily missed in all the startup output.

server/cli/launch.py
```diff
@@ -26,14 +26,14 @@
 
 def annotation_args(func):
     @click.option(
-        "--experimental-annotations",
+        "--disable-annotations",
```
bkmartinjr: I can't find a test case to test this feature (required for non-experimental features).  Please add at least a simple test that the system works correctly with annotations disabled.

server/cli/launch.py
sidneymbell: My understanding from our discussion was that the new ontologies module resolved the issues identified as blockers for removing the experimental tag. Apologies if I've misunderstood. @colinmegill @liaprins-czi -- What are your feelings on this?

server/cli/launch.py
sidneymbell: Will do, thanks for the catch

server/cli/launch.py
sidneymbell: @liaprins-czi -- I'll defer to you on these two CLI args. Just let me know what you'd like and I'll update.

server/cli/launch.py
bkmartinjr: Definitely not - we explicitly discussed the need to look at hierarchies and other features previously outlined by Lia before we move to fully supported.   I also don't see the harm in leaving as is given that it is not impeding our pilot user.

server/cli/launch.py
liaprins-czi: I think Bruce's rationale for `--annotations-file` makes a lot of sense, and it is shorter too, so I would go with that.

server/cli/launch.py
liaprins-czi: That makes sense; I agree with it being `--annotations-directory` for that reason

server/cli/launch.py
liaprins-czi: I think regular annotations is ready to leave the experimental realm, but I think there's more to do to make ontologies good enough (for example exploring more around the concept of revealing hierarchies and whether to/how to limit the label autosuggestions to children of the category parent, etc).

sidneymbell: Cool, thanks for clarifying. Will fix :)

server/cli/launch.py
sidneymbell: Kk, will update

docs/posts/annotations.md
bkmartinjr: this is no longer relevant given our decision to default to a single location.   Am resolving this conversation.

client/Makefile
```diff
@@ -73,15 +69,18 @@ e2e-annotations:
 
 .PHONY: smoke-test
 smoke-test:
-	start_server_and_test '$(MAKE) start-server' $(CXG_SERVER_PORT) '$(MAKE) e2e'
+	start_server_and_test \
```
bkmartinjr: super tiny nit: I would make the `smoke-test` target test the default configuration (ie, annotations on), and change the other target to `smoke-test-disabled-annotations`

CC: @mweiden 

IMHO, this can be done in another PR unless it is easy/quick.

docs/posts/annotations.md
bkmartinjr: Nit:  what is `editable mode`?   Maybe "If annotations are enabled, cellxgene will automatically find and reload your draft annotations"?

Also, this is only true if the user has not specified an alternative annotations file or directory.

bkmartinjr: ps. I'm fine if this is refined post-PR

server/cli/launch.py
```diff
@@ -26,29 +26,29 @@
...
         show_default=True,
         multiple=False,
         metavar="<path>",
         help="CSV file to initialize editing of existing annotations; will be altered in-place. "
-        "Incompatible with --annotations-output-dir.",
+        "Incompatible with --annotations-dir.",
     )
     @click.option(
-        "--experimental-annotations-output-dir",
+        "--annotations-dir",
```
bkmartinjr: Ideally we would also have test cases that execise --annotations-dir.    Also fine to leave for future work.

bkmartinjr:
I just noticed the unit tests are failing - please make sure they pass before merging.

client/Makefile
mweiden: Yeah, we probably need to chase this in a separate PR as making `smoke-test` have all default tests doesn't mesh well with the way the smoke tests tests are currently packaged.

PR #1191, opened 2020-03-03
Disable ColorBy button for truncated categories

### LINKED ISSUE #1156: color-by on truncated category causes browser hang
If you load a dataset with a very large obs categorical annotation (eg, immune_bone_marrow test dataset, cell_names category), and click the "droplet" color-by icon, it will hang the browser.

It is likely creating a separate color for each category.  It should bail out due to the excessive size of this category (just like the categorical menu in the LSB does)

Suggested fix - one of:
a)  if a category has more than some cut-off number, disable color-by
b)  or, just color-by the same list of categories shown in the truncated drop-down list

mweiden:
Fixes #1156

For categories that have more than 100 labels we truncate the labels in the UI, but still allowed users to ColorBy these categories. Coloring by these categories can cause browsers to get bogged down.

This commit disables ColorBy for truncated categories.

### DIFF
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -231,10 +231,11 @@ class Category extends React.Component {
           <Tooltip
             content={
               isTruncated
-              ? `Can't color by this category, too many label values (max ${globals.maxCategoricalOptionsToDisplay})`
+              ? `This category exceeds the label limit (${globals.maxCategoricalOptionsToDisplay}) and can't be colored by`
               : "Use as color scale"
             }
             position="bottom"
+            usePortal={false}
             hoverOpenDelay={globals.tooltipHoverOpenDelay}
           >
             <Button
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -231,8 +231,8 @@ class Category extends React.Component {
           <Tooltip
             content={
               isTruncated
-              ? `This category exceeds the label limit (${globals.maxCategoricalOptionsToDisplay}) and can't be colored by`
-              : "Use as color scale"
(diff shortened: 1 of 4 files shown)


mweiden:
Preview:
Screen Shot 2020-03-03 at 8 33 20 PM
Screen Shot 2020-03-03 at 8 33 11 PM

colinmegill:
Prefer: `Coloring by fooCategory is disabled, as it exceeds the limit of 100 labels`, and if we're doing this we might bump the limit to 150 or 200. I think we're trying to catch 1000, or 10,000 ie., one for every cell. I don't think we're trying to catch a large ontology.

bkmartinjr:
> bump the limit to 150 or 200. I think we're trying to catch 1000, or 10,000

we do create quite a large number of DOM objects per label, so I wouldn't go to 1000.  But @colinmegill 's suggestion seems reasonable to me.

mweiden:
@colinmegill 
Screen Shot 2020-03-04 at 9 08 15 AM

colinmegill:
@mweiden The button in your image looks active, not disabled, unless that's intentional https://blueprintjs.com/docs/#core/components/button

image

liaprins-czi:
Good catch @colinmegill, it does look not completely disabled. I just inspected the code and it does say it has class of `bp3-active`, but it's easy to fix by changing that to `bp3-disabled`, and it will look like this:
Screen Shot 2020-03-04 at 9 45 10 AM

colinmegill:
Nice catch @liaprins-czi! 

This should be something like: 

`<Button disabled={doesExceedMaxAllowedLabels}`

mweiden:
@colinmegill @liaprins-czi Ah! Didn't realize. OK. Submitting a separate PR to fix this.

PR #1195, opened 2020-03-04
Make smoke tests faster, more stable

mweiden:
## Need
Primary goal of this PR is to fix [a heisenbug that occurs only in CI](https://github.com/chanzuckerberg/cellxgene/pull/1191/checks?check_run_id=485516637).

## Approach
Through the investigation I made the following changes/refactors (some of which are tangential to the original issue):
* Bulk add genes in the `undo selection appends the top diff exp genes to user defined genes` and `subset selection appends the top diff exp genes to user defined genes` tests
* Refactor some of the methods in `puppeteerUtils` and `cellxgeneActions` to have better functional style
* Better modularize test helpers by moving `cellxgeneActions` and `setupTestBrowser` out of `puppeteerUtils.js` and into their own files.

## Benefits
1. It's not 100% clear to me that I have fixed the original issue in CI. All the runs of CI that I've triggered as part of this PR have passed. It's possible that there was a hiccup in the GithubActions runner... 🤞
1. The test helpers are a little cleaner and more modular.

### DIFF
--- a/client/__tests__/e2e/cellxgeneActions.js
+++ b/client/__tests__/e2e/cellxgeneActions.js
@@ -1,0 +1,196 @@
+import { strict as assert } from "assert";
+
+export const cellxgeneActions = (page, utils) => ({
+
+  async drag(testId, start, end, lasso = false) {
+    const layout = await utils.waitByID(testId);
+    const elBox = await layout.boxModel();
+    const x1 = elBox.content[0].x + start.x;
+    const x2 = elBox.content[0].x + end.x;
+    const y1 = elBox.content[0].y + start.y;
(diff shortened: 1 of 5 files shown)


mweiden:
@bkmartinjr I think it might have something to do with adding genes individually to the search bar and then waiting for them to render. This is why I tried to add genes via `actions.bulkAddGenes`. There were/are calls to `util.waitForAllByIds` in there, but I'm still a bit dubious of whether the rendering lag and timeouts are all working well together.

bkmartinjr:
that does make sense, as there is an autosuggest with a mind of its own on the other UI.

PR #1198, opened 2020-03-05
Do not over-prune var data cache

mweiden:
There is a bug in how the universe and world gene sets are constructed and passed to `ControlsHelpers.pruneVarDataCache` that causes the var data cache to be over-pruned. This commit fixes the issue.

Consider the following example from the node console that mirrors the bug in the code:

```
❯ node
Welcome to Node.js v13.5.0.
Type ".help" for more information.
> new Set([1], [2], [3])
Set(1) { 1 }
```

What we really want is the set `Set(3) { 1, 2, 3 }`, which can be constructed as:

```
> new Set([].concat([1], [2], [3]))
Set(3) { 1, 2, 3 }
```

### DIFF
--- a/client/src/reducers/universe.js
+++ b/client/src/reducers/universe.js
@@ -62,9 +62,7 @@ const Universe = (state = null, action, nextSharedState, prevSharedState) => {
       const { userDefinedGenes, diffexpGenes } = prevSharedState;
       const allTheGenesWeNeed = [
         ...new Set(
-          userDefinedGenes,
-          diffexpGenes,
-          Object.keys(action.expressionData)
+          [userDefinedGenes, diffexpGenes, Object.keys(action.expressionData)].filter(ele => ele).flat()
         )
       ];
       varData = ControlsHelpers.pruneVarDataCache(varData, allTheGenesWeNeed);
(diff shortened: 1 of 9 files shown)


client/src/reducers/universe.js
```diff
@@ -61,11 +61,11 @@ const Universe = (state = null, action, nextSharedState, prevSharedState) => {
       //
       const { userDefinedGenes, diffexpGenes } = prevSharedState;
       const allTheGenesWeNeed = [
-        ...new Set(
+        ...new Set([].concat(
```
bkmartinjr: IMHO, a more readable syntax (and much more modern):
```
   new Set([
        ...userDefinedGenes,
        ...diffexpGenes,
        ...Object.keys(action.expressionData)
    ])

File: client/src/reducers/world.js [hunk 2]
```diff
@@ -100,11 +100,11 @@ const WorldReducer = (
       //
       const { userDefinedGenes, diffexpGenes } = prevSharedState;
       const allTheGenesWeNeed = [
-        ...new Set(
+        ...new Set([].concat(
```
bkmartinjr: suggest using spread operator

Issue #38, opened 2018-05-08
One Million Cells: Tracking Issue


csweaver:
App is "usable" on a modern laptop (eg, macbook pro) with 1M cells loaded, eg, speed and interactive performance are reasonable.   On more typical data sets (eg, 250K cells), speed & interactive performance should be excellent (no jank, etc.)

csweaver:
@bkmartinjr I know you had a third idea, but I don't remember it. Could you create that issue?

PR #1200, opened 2020-03-07
hosted cellxgene (#38)

bmccandless:
In this PR, contains scripts and instructions for deploying cellxgene
for AWS elastic beanstalk.  It supports the multi-dataset option.

The Makefile in the server/eb directory creates an artifact.zip
file, which can be deploy at AWS EB.

The server/eb directory contains:
  app.py - flask app to run the server
  Makefile - which creates an artifact.zip file which can be deployed.
  README.md - instructions for setting up and deploying the eb app.


scripts/copy_client_assets.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
bkmartinjr: Tiny nit:  given that this script is only invoked from inside Makefiles, why not simply define it as a Makefile target (eg, put it in common.mk or something similar)?

server/eb/README.md
```diff
@@ -0,0 +1,65 @@
+# AWS Elastic Beanstalk
```
bkmartinjr: Please add a large, bold caveat to the effect that this is an experimental, incomplete feature, subject to change.  And that we would love feedback :-)

server/eb/app.py
```diff
@@ -0,0 +1,53 @@
...
+    logging.exception("Exception importing server modules")
+    sys.exit(1)
+
+try:
+    dataroot = os.getenv("CXG_DATAROOT")
+    if dataroot is None:
+        logging.error("CXG_DATAROOT environment variable must be set")
+        sys.exit(1)
+
+    app_config = AppConfig(dataroot=dataroot)
```
bkmartinjr: For this interim PR (ie, prior to having a more robust configuration mechanism), I would add explicit setting of any configuration where we must have a certain value and the default might change.   Specifically, I would explicitly disable annotations.

server/eb/app.py
```diff
@@ -0,0 +1,53 @@
+"""cellxgene AWS elastic beanstalk application"""
+
+import sys
+import os
+import logging
+
+logging.basicConfig(
+    filename="/opt/python/log/app.log",
```
bkmartinjr: Question on logging:  will this also put the info into the web logs so it is visible from the EB/AWS console?  Or does this only drop it into the file?   Naively, I assumed we would want to see the logs in the default console/tooling.

server/eb/README.md
```diff
@@ -0,0 +1,65 @@
+# AWS Elastic Beanstalk 
+
+This directory contains script to aid in creating and deploying cellxgene on
+an AWS Elastic Beanstalk instance.
```
bkmartinjr: It might be good to add a quick summary of what this will provide, once deployed.  For example:

> This will result in a variant of cellxgene, running on AWS EC2 instances, serving data from S3.   All datasets must be in the new CXG (tiledb) format - see the converter script `cxgtool.py` located ??? - and located in a single S3 prefix, which is accessible to the instance.   In the current incarnation, no access control or authentication support is available (outside of anything you configure yourself), so this is most appropriate for public datasets.

> This is early development work, and will change significantly in the near future.   We would love feedback on it, but please assume it will change.

server/eb/app.py
bmccandless: Yes agreed.   I'll fill in all the config parameters.
annotations are already turned off a few lines down.
What about diffexp? Should that be enabled here or disabled?

server/eb/app.py
bmccandless: Logging goes to both /opt/python/log/app.log and stdout.
The next statement down has the handling for stdout.
/opt/python/log/app.log  is the magic location that makes it visible (e.g. 'eb logs' will find it).
I'll add a check that /opt/python/log exists before adding this line, so that you can run the app locally without problems.

server/eb/README.md
bmccandless: Looks good, I'll add that.

server/eb/app.py
bkmartinjr: for now, I think it is fine to enable diffexp.   When we do a more robust configuration mechanism, it can be refined.

server/eb/app.py
bkmartinjr: Out of curiosity, what is the value of also logging into a file?  Will deployments need to add log rotation, etc?

I'm fine leavings as-is for this interim PR.

scripts/copy_client_assets.sh
bmccandless: Done.

server/eb/app.py
bmccandless: I am doing it both to a file and stdout for some convenience.  I was able to test my application locally and catch error more quickly that way when I could see the logs in stdout.  But for running in EB, the logs need to go to that file.    We will probably want to add log rotation too.

bkmartinjr: sounds good.  I suspect much of this other configuration & deployment tooling belongs outside the cellxgene repo.  Thanks for the detailed response.

PR #1206, opened 2020-03-10
Conditionally truncate category string

colinmegill:
Fixes ##1161

### DIFF
--- a/client/__tests__/e2e/e2eAnnotations.test.js
+++ b/client/__tests__/e2e/e2eAnnotations.test.js
@@ -183,7 +183,11 @@ describe.each([
   });
 
   async function assertCategoryExists(categoryName) {
-    await utils.waitByID(`${categoryName}:category-expand`);
+    const handle = await utils.waitByID(`${categoryName}:category-expand`);
+    const result = await handle.evaluate(node => node.innerText);
+    // slice beginning and end of category name result to account for truncation of long names
+    expect(result.slice(0, 10)).toBe(categoryName.slice(0, 10));
+    expect(result.slice(-10)).toBe(categoryName.slice(-10));
   }
 
   async function assertCategoryDoesNotExist(categoryName) {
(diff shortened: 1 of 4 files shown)


client/__tests__/e2e/e2eAnnotations.test.js
```diff
@@ -183,8 +183,7 @@ describe.each([
   });
 
   async function assertCategoryExists(categoryName) {
-    const result = await utils.waitByID(`${categoryName}:category-expand`);
-    expect(await result.evaluate(node => node.innerText)).toBe(categoryName);
+    await utils.waitByID(`${categoryName}:category-expand`);
```
bkmartinjr: rather than deleting the `expect`, can we do a weaker test?  For example, test that the first 10 and last 10 characters match, or something like that?

bkmartinjr: What I had in mind (sorry if it contains typos):

```
const foundCatName = await result.evaluate(node => node.innerText);
expect(foundCatName.slice(0, 10)).toBe(categoryName.slice(0, 10))
expect(foundCatname.slice(-10).toBe(categoryName.slice(-10))
```

bkmartinjr: or pick a number (instead of 10) that matches closer to what you expect.

mweiden: @colinmegill sounds like our second solution might be worthwhile

bkmartinjr: @mweiden - if I'm mistaken, and this _is_ being tested elsewhere, feel free to land the PR.  It just seemed easy to make the test almost as robust as it was previously.  LMK if I'm off base!

mweiden: @bkmartinjr slight misunderstanding here: `waitByID` throws an exception if it doesn't find the element it is looking for, so the test would fail.

PR #1035, opened 2019-11-12
Add color mapping to the bar chart.

Donaldpherman:
enhancement for #867

PR #1218, opened 2020-03-12
Add color mapping to the bar chart

mweiden:
Finishing up https://github.com/chanzuckerberg/cellxgene/pull/1035

Changes made on top of @Donaldpherman's PR:
* Only color histograms that are selected for colorby
* Reuse binning functions from `util/dataframe/histogram.js`; this fixes an issue with d3 generating near-zero width bins
* Fix/simplify histogram color mapping functions
* Do not attempt to plot bins if the calculated binWidth is zero; this can happen if all dataframe column values are the same
* Refactor the function that draws the histogram a bit

Illustration of fixes to color mapping are below. Each shows the legend scaled against the colored histogram bins.

Before (source @liaprins-czi):
Screen Shot 2020-03-11 at 11 32 26 AM

After (source this PR):
Screen Shot 2020-03-11 at 10 38 15 PM

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -334,15 +334,11 @@ class HistogramBrush extends React.PureComponent {
       .attr("class", "histogram-container")
       .attr("transform", `translate(${this.marginLeft},${this.marginTop})`);
 
-    const yMax = d3.max(bins, function (d) { return d.length });
-    const yMin = d3.min(bins, function (d) { return d.length });
-
-    const colorScale = d3.scaleSequential(interpolateCool)
-      .domain([yMin, yMax]);
+    const colorScale = d3.scaleSequential(interpolateCool).domain([0, bins.length]);
 
     const histogramScale = d3
       .scaleLinear()
-      .range([yMin, yMax])
+      .range([1, this.width + this.marginLeft + this.marginRight])
       .domain([
         colorScale.domain()[1],
         colorScale.domain()[0]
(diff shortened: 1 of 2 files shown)


client/src/components/brushableHistogram/index.js
```diff
@@ -61,17 +61,14 @@ class HistogramBrush extends React.PureComponent {
...
+    histogramCache.binWidth = (domainMax - domainMin) / numBins;
 
-    histogramCache.bins = d3
-      .histogram()
-      .domain(histogramCache.x.domain())
-      .thresholds(histThresholds)(values);
+    histogramCache.binStart = i => domainMin + i * histogramCache.binWidth;
+    histogramCache.binEnd = i => domainMin + (i + 1) * histogramCache.binWidth;
+
+    const yMax = Math.max(...histogramCache.bins);
```
bkmartinjr: Style nit - while it is fine in this case because numBins is locally defined and small, be careful using the spread operator this way.  It will push every array element onto the call stack (ie, it duplicates the entire array just for the single function call).   In many cases, a call to reduce() is going to be faster, and will avoid the stack overflow exception if the array gets too big.

mweiden: Fixed. 👍

client/src/components/brushableHistogram/index.js
```diff
@@ -93,67 +90,54 @@ class HistogramBrush extends React.PureComponent {
...
-      if (brushXselection) {
-        const selection = d3.brushSelection(brushXselection.node());
-        if (!range && selection) {
-          /* no active selection - clear brush */
-          brushXselection.call(brushX.move, null);
-        } else if (range && !selection) {
-          /* there is an active selection, but no brush - set the brush */
-          const x0 = x(range[0]);
-          const x1 = x(range[1]);
+    if ((dfChanged || rangeChanged) && brushXselection) {
```
bkmartinjr: this should be

```
    if ((dfChanged || rangeChanged || colorSelectionChanged) && brushXselection) {
```

PR #1227, opened 2020-03-15
fixes for hosted cellxgene

bmccandless:
* fixes for --dataroot and favicon
 * fixes for static assets with hosted cellxgene

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -89,10 +89,12 @@ def dataroot_index():
     # FIXME with a splash screen that includes a listing of all the datasets.
     # or perhaps a login screen if this is a hosted environment,
     # or have a configuration option to redirect to a user specified page.
-    data = '<!doctype html><html lang="en"><head>'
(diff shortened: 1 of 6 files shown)


server/app/app.py
```diff
@@ -85,15 +85,12 @@ def static_redirect(dataset, therest):
...
-    return redirect("/static/favicon.png", code=301)
-
-
 def dataroot_index():
     # FIXME with a splash screen that includes a listing of all the datasets.
-    # or perhaps a login screen if this is a hosted environment
-    data = "<H1>Welcome to cellxgene</H1>"
+    # or perhaps a login screen if this is a hosted environment,
+    # or have a configuration option to redirect to a user specified page.
+    data = '<!doctype html><html lang="en"><head>'
```
bkmartinjr: this doesn't look right - there is no closing tag for <head>.

If it is empty, you can remove it, or use <head/>

bmccandless: Thanks, I just pushed a fix for <head>

PR #1234, opened 2020-03-17
Improve hosted cellxgene

bmccandless:
- option to turn off the test index page, or supply a page for redirect.
   For EB, The default is to return 404.  For cli launch, the default is the test page.

 - option to select which matrix types are allowed for multi dataset servers.
   For EB, The default is CXG only.  For cli launch, the default is any matrix type.

 - Return early with an error response if diffexp is requested when not configured

 - Verified that reembedings and user annotations also return with an error response
   if used when not enabled.

TODO:  The new options cannot currently be set by the user.
I plan to add a configuration file where these and all other settings can be set.

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -36,10 +36,6 @@ def dataset_index(dataset=None):
 
     try:
         cache_manager = current_app.matrix_data_cache_manager
-        loader = MatrixDataLoader(location)
-        if not loader.matrix_data_type_allowed(config):
-            return make_response(f"Invalid matrix type: {loader.matrix_data_type}", HTTPStatus.BAD_REQUEST)
-
         with cache_manager.data_adaptor(location, config) as data_adaptor:
             dataset_title = config.get_title(data_adaptor)
             return render_template("index.html", datasetTitle=dataset_title, SCRIPTS=scripts)
@@ -101,9 +97,12 @@ def dataroot_test_index():
         datasets = []
         for fname in locator.ls():
             location = path_join(config.dataroot, fname)
-            matrix_data_loader = MatrixDataLoader(location)
-            if matrix_data_loader.matrix_data_type_allowed(config):
+            try:
+                MatrixDataLoader(location, app_config=config)
                 datasets.append(fname)
+            except DatasetAccessError:
+                # skip over invalid datasets
+                pass
 
         data += '<br/>Select one of these datasets...<br/>'
         data += '<ul>'
@@ -121,9 +120,9 @@ def dataroot_test_index():
 def dataroot_index():
     # Handle the base url for the cellxgene server when running in multi dataset mode
     config = current_app.app_config
-    if config.multi_dataset_index is None:
+    if not config.multi_dataset_index:
         abort(404)
-    elif config.multi_dataset_index == "test":
+    elif config.multi_dataset_index is True:
         return dataroot_test_index()
     else:
         return redirect(config.dataroot_index)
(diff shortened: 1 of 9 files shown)


server/app/app.py
```diff
@@ -36,6 +36,10 @@ def dataset_index(dataset=None):
 
     try:
         cache_manager = current_app.matrix_data_cache_manager
+        loader = MatrixDataLoader(location)
```
bkmartinjr: why not just have the MatrixDataLoader enforce the data format restrictions, and treat "disabled" types the same as "unsupported" types?   I'm concerned that this feature (enable/disable formats) belongs lower, in a single place, so any other future callers benefit from the config.

bkmartinjr: put another way, why not just have MatrixDataLoader return an "invalid" object whenever the location points at something unsupported/not-understood/disabled?   Or even throw an exception if the location is invalid/disabled/unsupported.

server/app/app.py
```diff
@@ -118,6 +118,17 @@ def dataroot_index():
     return make_response(data)
 
 
+def dataroot_index():
+    # Handle the base url for the cellxgene server when running in multi dataset mode
+    config = current_app.app_config
+    if config.multi_dataset_index is None:
```
bkmartinjr: None or False for safety?   or `if not config.multi_dataset_index: ...`

server/common/app_config.py
```diff
@@ -35,6 +35,15 @@ def __init__(self, **kw):
         self.enable_reembedding = False
         self.anndata_backed = False
 
+        # The index page when in multi-dataset mode:
+        #   None:  this returns a 404 code
+        #   "test":  loads a test index page, which links to the datasets that are available in the dataroot
+        #   <redirect>:  anything else will result in a redirect:  flask.redirect(config.dataroot_index)
+        self.multi_dataset_index = "test"
```
bkmartinjr: to make this more "clear", you might want to use a True/False/URL value -- ie,
- False or None - 404
- True - generate index
- string/URL - redirect
This is a nit.

server/data_common/matrix_loader.py
```diff
@@ -136,33 +136,56 @@ class MatrixDataType(Enum):
...
-    def matrix_data_type(self):
+    def __matrix_data_type(self):
         if self.location.path.endswith(".h5ad"):
             return MatrixDataType.H5AD
         elif ".cxg" in self.location.path:
             return MatrixDataType.CXG
         else:
             return MatrixDataType.UNKNOWN
 
+    def matrix_data_type_allowed(self, app_config):
```
bkmartinjr: I'd rather see these checks done in the __init__ so it is impossible to get an instance if you have an unsupported/invalid/disabled data type.   Suggestion: have __init__ throw an "invalid data type" exception.

server/app/app.py
bmccandless: Good suggestion.  The code is simpler now.

server/app/app.py
bmccandless: Done

server/common/app_config.py
bmccandless: Done

server/common/app_config.py
```diff
@@ -35,6 +35,15 @@ def __init__(self, **kw):
         self.enable_reembedding = False
         self.anndata_backed = False
 
+        # The index page when in multi-dataset mode:
+        #   False or None:  this returns a 404 code
+        #   True:  loads a test index page, which links to the datasets that are available in the dataroot
+        #   string/URL:  redirect to this URL:  flask.redirect(config.multi_dataset_index)
+        self.multi_dataset_index = True
```
bkmartinjr: I'd prefer this were off by default - is there any value in having debugging route on by default?

bmccandless: Okay, I can turn that off.  I was already off in the eb app.

PR #1254, opened 2020-03-19
Titlebar URL opens new page

### LINKED ISSUE #1250: dataset title "about link" should open new browser window
cellxgene has the ability to set an "about" link on the datasets - ie, a URL that would contain more info about the dataset.   This displays in the Info menu ("About this dataset") _and_ makes the dataset title (top-left corner) a link rather than straight text.   

The menu item opens a new browser window, leaving the current cellxgene state/window as is.    The "title link" does not - it browses the current window to the link, losing the cellxgene state and data.  if the user hits the link, then goes back, cellxgene reloads.

Both should open a new browser window - ie, the title link should behave like the Info menu link.

colinmegill:
Fixes #1250

### DIFF
--- a/client/src/components/leftSidebar/topLeftLogoAndTitle.js
+++ b/client/src/components/leftSidebar/topLeftLogoAndTitle.js
@@ -80,7 +80,7 @@ class LeftSideBar extends React.Component {
           }}
           title={datasetTitle}
         >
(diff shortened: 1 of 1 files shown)


colinmegill:
@mweiden I can't explain the codecov report here for a single line change :)

PR #1260, opened 2020-03-20
prepare - work around anndata bug

bkmartinjr:
Work around for the issue theislab/anndata#344

Changes to launch are just lint (black autoformat).  See change in prepare.py.

### DIFF
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -138,7 +138,7 @@ def dataset_args(func):
         "-t",
         default=DEFAULT_CONFIG.single_dataset__title,
         metavar="<text>",
-        help="Title to display. If omitted will use file name."
+        help="Title to display. If omitted will use file name.",
     )
     @click.option(
         "--about",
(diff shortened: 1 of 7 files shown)


mweiden:
@bkmartinjr posted a solution here. While I do think it is an improvement to the current anndata implementation, perhaps we should consider using different `join` character in the function since there are [gene names that follow the format `[a-zA-Z]-[0-9]+`](https://www.ncbi.nlm.nih.gov/gene/767561). I don't know if gene names follow a specific format, but perhaps it's possible to chose a character not in that format. Perhaps something like `.` or `#`?

bkmartinjr:
> I don't know if gene names follow a specific format, but perhaps it's possible to chose a character not in that format. Perhaps something like `.` or `#`?

How would this resolve the underlying algorithmic problem?   I elected to use the Scanpy/Anndata standard rather than inventing another.

mweiden:
@bkmartinjr 

> How would this resolve the underlying algorithmic problem?

it wouldn't solve the algorithmic problem. It would simply make the names in the indices we produce more easily interpretable.

> I elected to use the Scanpy/Anndata standard rather than inventing another.

Technically, if we can convince ourselves that a char like `.` or `#` doesn't occur in standard gene names, you could use the current Scanpy/Anndata release and just pass that char to the `join` parameter in `.var_names_make_unique` and `.obs_names_make_unique`

bkmartinjr:
> doesn't occur in standard gene names

There is no guarantee that the names will conform to this standard.  And we really can't assume that either.  They can be anything.

PR #1262, opened 2020-03-20
initial implementation of a health check route

### LINKED ISSUE #1247: health check route
Need a top-level server health-check route.

- [ ] round-trip read to dataroot to ensure access to datasets

CC: @parthshahva

bkmartinjr:
Implement a simple health-check route. Currently will check for accessibility of data root or data path, dependent on configuration.

Fixes #1247

### DIFF
--- a/server/common/health.py
+++ b/server/common/health.py
@@ -11,10 +11,8 @@ def _is_accessible(path):
 
     try:
         dl = DataLocator(path)
-        print("exists", dl.exists())
         return dl.exists()
     except RuntimeError as e:
-        print(e)
         return False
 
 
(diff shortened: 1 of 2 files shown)


server/common/health.py
```diff
@@ -0,0 +1,31 @@
...
+from server.common.data_locator import DataLocator
+
+
+def _is_accessible(path):
+    if path is None:
+        return True
+
+    try:
+        dl = DataLocator(path)
+        print("exists", dl.exists())
```
bmccandless: extra print statement should be removed.

bkmartinjr: doh!  will do

PR #1264, opened 2020-03-21
Introduce a config file to cellxgene

bmccandless:
The config file format is in yaml.  The default yaml file is located
in server/common/default_config.yaml.  A user my create a yaml file
that contains a subset of these fields.  It can be used during cellxgene
launch, or for hosted cellxgene.

The code has also been refactored.  Much of the logic to check arguments
has moved from launch to app config.

It is now possible to set the tiledb context parameters using the config
file.  Other feature will soon be handled in a similar way.


server/common/default_config.yaml
```diff
@@ -0,0 +1,55 @@
...
+annotations:
+  enable: true
+  type: local_file_csv
+  local_file_csv:
+    directory: null
+    file: null
+  ontology:
+    enable: false
+    obo_location: null
+  max_categories: 1000
```
bkmartinjr: not sure that `max_categories` belongs in this group.  The parent `annotations` is really controlling user-created annotations. The `max_categories` affects all annotations, even those in the base data (immutable), and is actually affecting the presentation of the annotations, not the ability to create/edit them.

Suggestions:
1. change top-level `annotations` to `user_anotations`
2. create a new top-level category `presentation`, and move `max_categories` into it

We have some new features (eg, color palette) that seem likely to need the `presentation` group.

server/common/default_config.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+embeddings:
+  names : []
+  enable_reembedding: false
+
+diffexp:
+  enable: true
+  lfc_cutoff: 0.01
+
+cxg_adaptor:
```
bkmartinjr: would it be useful to create a top-level group `adaptors`, and push the various adaptor config down a level (keep the top namespace clean)?   Ie, `cxg_adaptor` and `anndata_adpator` are children of `adaptor`.   That would also give you a spot to hang generic adaptor config when it arises.

server/data_common/data_adaptor.py
```diff
@@ -21,7 +21,9 @@ def __init__(self, config):
         if config is None:
```
bkmartinjr: Can we clean up the need for test backward compat (ie, fix the tests)?  Completely OK if it is done in a later cleanup PR, just don't want to forget about it (file an issue)

server/common/default_config.yaml
```diff
@@ -0,0 +1,55 @@
+
+server:
+  verbose: true
```
bkmartinjr: for most environments, the default verbose should be false

server/common/default_config.yaml
```diff
@@ -0,0 +1,55 @@
...
+  debug: false
+  host: "127.0.0.1"
+  port : null
```
bkmartinjr: shouldn't this by 5005 by default?

server/common/app_config.py
```diff
@@ -19,74 +38,312 @@ def todict(self):
...
     def get_title(self, data_adaptor):
-        return self.title if self.title else data_adaptor.get_title()
+        return self.single_dataset__title if self.single_dataset__title else data_adaptor.get_title()
 
     def get_about(self, data_adaptor):
-        return self.about if self.about else data_adaptor.get_about()
+        return self.single_dataset__about if self.single_dataset__about else data_adaptor.get_about()
 
-    def get_config(self, data_adaptor, annotation=None):
+    def get_client_config(self, data_adaptor, annotation=None):
```
bkmartinjr: I suggested adding a comment describing the purpose of this method, eg, 

"""
Return the configuration as required by the /config REST route
"""

server/common/app_config.py
```diff
@@ -1,6 +1,25 @@
...
+from urllib.parse import urlparse
+
+from server.common.errors import ConfigurationError, DatasetAccessError, OntologyLoadFailure
+from server.data_common.matrix_loader import MatrixDataLoader, MatrixDataCacheManager, MatrixDataType
+from server.common.utils import find_available_port, is_port_available
+import warnings
+from server.common.annotations import AnnotationsLocalFile
+from server.common.utils import custom_format_warning
+
+DEFAULT_SERVER_PORT = int(environ.get("CXG_SERVER_PORT", "5005"))
```
bkmartinjr: shouldn't these come from the default config?

server/common/app_config.py
```diff
@@ -19,74 +38,312 @@ def todict(self):
...
-                setattr(self, k, v)
-            else:
-                raise RuntimeError(f"unknown config parameter {k}.")
+        # load from default_config.yaml
+        default_config_file = pkg_resources.resource_filename(__name__, "default_config.yaml")
+        with open(default_config_file) as fyaml:
+            self.default_config = yaml.load(fyaml, Loader=yaml.FullLoader)
+
+        dc = self.default_config
+        self.server__verbose = dc["server"]["verbose"]
```
bkmartinjr: these references into the default config will generate a KeyError if the default config is missing something or otherwise malformed.   You might want to wrap all of these references in a try/except, and throw a ConfigurationError if it is caught, indicating a bad install/packaging of the cellxgene module (ie, default_config.yaml is malformed)

server/common/app_config.py
```diff
@@ -19,74 +38,312 @@ def todict(self):
...
+        self.anndata_adaptor__backed = dc["anndata_adaptor"]["backed"]
+
+        # The annotation object is created during complete_config and stored here.
+        self.annotations = None
+
+    def update_from_config_file(self, config_file):
+        with open(config_file) as fyaml:
+            config = yaml.load(fyaml, Loader=yaml.FullLoader)
+
+        # special cae for tiledb_ctx whose value is a dict, and cannot
```
bkmartinjr: typo:  s/cae/case/

server/common/app_config.py
```diff
@@ -19,74 +38,312 @@ def todict(self):
...
-
-        # A list of allowed matrix types.  If an empty list, then all matrix types are allowed
-        self.multi_dataset_allowed_matrix_type = []
-
-        # TODO these options may not apply to all datasets in the multi dataset.
-        # may need to invent a way to associate these config parameters with
-        # specific datasets.
-        self.obs_names = None
-        self.var_names = None
+    def __init__(self):
```
bkmartinjr: Suggestion to prevent future mistakes / be more paranoid:  add a check to ensure that complete_config() is called before the config is used.

In other words:
1. add an instance value called `is_completed`.   
    * set to False in '__init__'
    * Set it to True in `complete_config()`
    * Set to False in `update()` and `update_from_config_file()`
2. Add a check for True (throw exception if False) in all of the 'get_*' methods

server/common/default_config.yaml
bmccandless: Sounds good.   I'll make that change

server/common/default_config.yaml
bmccandless: Good idea, will do.

server/data_common/data_adaptor.py
bmccandless: I might as well do that.   I've already had to modify all the tests.

server/common/default_config.yaml
bmccandless: Yes, agreed it should be false.

server/common/default_config.yaml
bmccandless: I didn't change behavior.  The way it works is if port is null then it gets the starting port from the environment variable, and tries that.  If that one is taken, it tries the next one up until it finds a free port.  However if the port is specified (like 5005), then it only tries that port and fails if it is in use.

server/common/app_config.py
bmccandless: Will do.

server/eb/Makefile
```diff
@@ -20,6 +20,9 @@ build: clean
 	cp app.py artifact.dir/application.py; \
 	cp ../requirements.txt artifact.dir; \
 	cp -r .ebextensions artifact.dir; \
+	if [ -f config.yaml ] ; then \
```
bkmartinjr: something that would be useful down the road (ie, maybe file as an enhancement ticket if you agree) is ability to separately deploy the config from the code.

To do that, I would imagine introducing an environment variable such as CXG_CONFIG, that points at a data location containing the YAML.

Definitely for later - just wanted to drop the idea in here

server/eb/app.py
```diff
@@ -29,29 +29,30 @@
...
-        layout=[],
-        max_category_items=100,
-        diffexp_lfc_cutoff=0.01,
-        obs_names=None,
-        var_names=None,
-        anndata_backed=False,
-        disable_diffexp=True,
-        multi_dataset_index=None,
-        multi_dataset_allowed_matrix_type=["cxg"],
+    # verify certain features are set for the hosted server
```
bkmartinjr: suggest rewording:    'features are unsupported in the current hosted server'

server/eb/app.py
```diff
@@ -62,7 +63,10 @@
     logging.exception("Caught exception during initialization")
```
bkmartinjr: this should be a logging.critical()

bkmartinjr: as should line 75

server/common/default_config.yaml
bkmartinjr: oh, gotcha.  NM!

server/common/app_config.py
bkmartinjr: pls ignore

server/common/app_config.py
bmccandless: My thought was that we are in control of default_config.yaml, and if there is ever a KeyError, it would be a developer bug that we would fix.  But you are right, there could be an issue with an old or malformed default_config.   So, I think what I want to do is add the contents of "default_yaml.conif" to a python multiline string in app_config.py (or maybe a new file "default_yaml.py") and use that instead of opening and reading from a yaml file.  

1. this makes it so there is one less thing to worry about.  I can still add the try/except to give better error messages.

2. it solves my packaging problem (the reason why some of the tests failed to find default_config.yaml).  I won't have to have some special case for how to handle a non python file.  

3. I would like to add a new cli command: "cellxgene config" which returns the default config.   In this way a user can have a workflow where they ask cellxgene for the config, stash it in a config file, modify the file, then  launch cellxgene with their config file.

server/common/app_config.py
bmccandless: Will do

server/eb/Makefile
bmccandless: Yes, agreed.  I wasn't completely happy with this solution.

server/common/app_config.py
bkmartinjr: that is a great solution.

server/eb/app.py
bmccandless: I choose exception to get the exception info into the logs.  I think this will do the same thing for critical:   logging.critical( "Caught exception during initialization", exc_info=True)

bkmartinjr: we definitely want it in the logs...

server/cli/config.py
```diff
@@ -0,0 +1,7 @@
+import click
+from server.common.default_config import default_config
+
+
+@click.command(short_help="Print the default cellxgene server configuration")
```
bkmartinjr: "server configuration", or just "configuration"?   Given that it affects front & backend, suggest `Print the default cellxgene configuration`

server/cli/cli.py
```diff
@@ -29,3 +30,4 @@ def cli(upgrade_check):
 
 cli.add_command(launch)
 cli.add_command(prepare)
+cli.add_command(config)
```
bkmartinjr: Why add this?  Unless I am misunderstanding, all it can do is display the internal default config.  Several confusions about this from a UI design standpoint:
- the config is only relevant to the launch command, which is not being run (eg, it doesn't configure `prepare`)
- the names/default values are not the same you see when you run `cellxgene launch --help`

Is the idea that it would give the user a hint about what the config file should look like, if they want to use `cellxgene launch -c <file>`?

If so, I suggest:
- it should be an option on the `launch` command, as that is what it is showing the config for
- it should have a name like `cellxgene launch --dump-default-config` (print it and exit)

Having a separate top-level sub-command doesn't really make sense given the current command structure.

server/data_common/data_adaptor.py
```diff
@@ -20,8 +20,12 @@ def __init__(self, config):
         # the following is for backwards compatability with tests
         if config is None:
             config = AppConfig()
+            assert(False)
         elif type(config) == dict:
-            config = AppConfig(**config)
+            temp = AppConfig()
```
bkmartinjr: if the tests no longer rely on this, can we delete the backward compat code?

server/eb/app.py
```diff
@@ -24,49 +24,53 @@
...
 
 try:
     dataroot = os.getenv("CXG_DATAROOT")
-    if dataroot is None:
-        logging.error("CXG_DATAROOT environment variable must be set")
-        sys.exit(1)
+    app_config = AppConfig()
+
+    config_file = "config.yaml"
+    if os.path.exists(config_file):
```
bkmartinjr: future enhancement:  the config file should be pulled from a environment variable, with a hard-wired backup.  And it should allow for an S3 URL (eg, store it in the dataroot) - I would just open it with the DataLocator class.    I'll file this as a ticket so as to not extend this PR.

server/test/test_api.py
```diff
@@ -79,6 +79,7 @@ def test_get_annotations_obs_fbs(self):
         self.assertEqual(result.headers["Content-Type"], "application/octet-stream")
         df = decode_fbs.decode_matrix_FBS(result.content)
         self.assertEqual(df["n_rows"], 2638)
+        print("ANNOTATIONS ENABLED", self.ANNOTATIONS_ENABLED)
```
bkmartinjr: debugging code?

server/cli/cli.py
bmccandless: Okay, I'll make it part of launch.   Yes, the idea the user will use it like this:

> cellxgene launch --dump_default_config > myconfig.yaml
> vi myconfig.yaml  # customize for my environment 
> cellxgene launch -c myconfig.yaml

server/eb/app.py
bmccandless: I like the idea of having the option to store the config in s3.  Good idea.

server/data_common/data_adaptor.py
bmccandless: Ah yes, I went to the trouble of fixing all the tests, then forgot to remove this code.   Will do.

server/eb/app.py
bkmartinjr: filed as #1273

server/test/test_api.py
bmccandless: Yep.  will remove.

PR #1269, opened 2020-03-21
tighten up error reporting

### LINKED ISSUE #1253: in prod hosted mode, don't dump full stack on error
Do not leak stack trace to HTTP response when errors occur.  

Changes:
- [ ] expose debug mode via deployment config, default Off for production.
- [ ] when debug is Off/False, generate terse HTTP response error messages instead of full stack dumps (debug=False).
- [ ] log error in all cases

From initial sec review:

in a lot of places y'all just give the user a response with str(exception) or str(a_file_path) in it, and that can leak internal state in helpful ways for people trying to develop exploits.   all the error messages are super not-verbose in production (add a debug mode maybe?)

a debug mode where the errors are displaying how they currently are (for max usability), but then in production, consider carefully before you put anything besides "server error" or something equally nondescript in a response to the user.

### LINKED ISSUE #1258: XSS issue
On dataset route, there is a reflected XSS vulnerability, proof of concept http://cellxgene-dev.us-west-2.elasticbeanstalk.com/%3Csvg%20onload=alert(1)%3E.cxg/

bkmartinjr:
improve error reporting:
* log errors or debug output, as appropriate
* don't leak excessive information in HTTP responses
In addition, ran black over the server to enforce our coding conventions.

Fixes #1253 
Fixes #1258


server/common/rest.py
```diff
@@ -219,6 +202,6 @@ def layout_obs_put(request, data_adaptor):
             },
         )
     except NotImplementedError as e:
-        return make_response(str(e), HTTPStatus.NOT_IMPLEMENTED)
-    except (ValueError, DisabledFeatureError, FilterError) as e:
-        return make_response(str(e), HTTPStatus.BAD_REQUEST)
+        return abort_and_log(HTTPStatus.NOT_IMPLEMENTED, str(e))
+    except (NotImplementedError, ValueError, DisabledFeatureError, FilterError) as e:
```
bmccandless: nit:  NotImplementedError handled above already.

server/common/rest.py
```diff
@@ -14,6 +14,13 @@
 from server.data_common.fbs.matrix import decode_matrix_fbs
 
 
+def abort_and_log(code, logmsg, loglevel=logging.DEBUG):
+    """ Debug-level log the message, then abort with HTTP code """
+    current_app.logger.log(loglevel, logmsg)
```
bmccandless: One improvement here would be to get the exception into the log file.   We could have a variant of this function that captures the exception information.  Maybe for a later PR.

server/common/rest.py
bkmartinjr: good catch!

server/common/rest.py
bkmartinjr: very easy to do - I'll add it now.

PR #1277, opened 2020-03-22
Fix matrix cache race condition

### LINKED ISSUE #1255: load test error
Against the dev server, in a load test, I frequently see failures with this trace.  I am running the load test against 8 different data sets/URLs.  Maybe the loader cache handling?

**(update):** this seems to occur only (so far) when the number of datasets being fetched exceed the internal data adaptor cache size.   Suggests some sort of cache handling bug, which is consistent with the stack crawl.

```
[Fri Mar 20 00:27:55.050656 2020] [:error] [pid 3572] Exception on /GSE60361.cxg/api/v0.2/schema [GET]
[Fri Mar 20 00:27:55.050665 2020] [:error] [pid 3572] Traceback (most recent call last):
[Fri Mar 20 00:27:55.050668 2020] [:error] [pid 3572]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/app.py", line 1949, in full_dispatch_request
[Fri Mar 20 00:27:55.050677 2020] [:error] [pid 3572]     rv = self.dispatch_request()
[Fri Mar 20 00:27:55.050679 2020] [:error] [pid 3572]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/app.py", line 1935, in dispatch_request
[Fri Mar 20 00:27:55.050681 2020] [:error] [pid 3572]     return self.view_functions[rule.endpoint](**req.view_args)
[Fri Mar 20 00:27:55.050683 2020] [:error] [pid 3572]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask_restful/__init__.py", line 468, in wrapper
[Fri Mar 20 00:27:55.050685 2020] [:error] [pid 3572]     resp = resource(*args, **kwargs)
[Fri Mar 20 00:27:55.050687 2020] [:error] [pid 3572]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask/views.py", line 89, in view
[Fri Mar 20 00:27:55.050689 2020] [:error] [pid 3572]     return self.dispatch_request(*args, **kwargs)
[Fri Mar 20 00:27:55.050691 2020] [:error] [pid 3572]   File "/opt/python/run/venv/local/lib/python3.6/site-packages/flask_restful/__init__.py", line 583, in dispatch_request
[Fri Mar 20 00:27:55.050693 2020] [:error] [pid 3572]     resp = meth(*args, **kwargs)
[Fri Mar 20 00:27:55.050695 2020] [:error] [pid 3572]   File "/opt/python/current/app/server/app/app.py", line 76, in wrapped_function
[Fri Mar 20 00:27:55.050697 2020] [:error] [pid 3572]     return func(self, data_adaptor)
[Fri Mar 20 00:27:55.050699 2020] [:error] [pid 3572]   File "/opt/python/current/app/server/app/app.py", line 134, in get
[Fri Mar 20 00:27:55.050700 2020] [:error] [pid 3572]     return common_rest.schema_get(data_adaptor, current_app.annotations)
[Fri Mar 20 00:27:55.050702 2020] [:error] [pid 3572]   File "/opt/python/current/app/server/common/rest.py", line 31, in schema_get
[Fri Mar 20 00:27:55.050704 2020] [:error] [pid 3572]     schema = schema_get_helper(data_adaptor, annotations)
[Fri Mar 20 00:27:55.050706 2020] [:error] [pid 3572]   File "/opt/python/current/app/server/common/rest.py", line 19, in schema_get_helper
[Fri Mar 20 00:27:55.050708 2020] [:error] [pid 3572]     schema = data_adaptor.get_schema()
[Fri Mar 20 00:27:55.050710 2020] [:error] [pid 3572] AttributeError: 'NoneType' object has no attribute 'get_schema'
```

bmccandless:
The simplest fix is to acquire the cache_item lock while holding the matrix cache lock.
This potentially causes a slowdown when a new matrix is loaded for the first time.
However, even with large datasets with is hardly noticable.  If the server finds itself in
the situation where the cache is being thrashed then the solution is to increase the
cache size.

Fixes #1255

### DIFF
--- a/server/data_common/matrix_loader.py
+++ b/server/data_common/matrix_loader.py
@@ -65,7 +65,7 @@ class MatrixDataCacheManager(object):
     when the context ends.  This class currently implements a simple least recently used cache,
     which can delete a dataset from the cache to make room for a new oneo
 
-    This is the indended usage pattern:
+    This is the intended usage pattern:
 
            m = MatrixDataCacheManager()
            with m.data_adaptor(location, app_config) as data_adaptor:
@@ -77,7 +77,11 @@ class MatrixDataCacheManager(object):
     #  TODO:  This is very simple.  This can be improved by taking into account how much space is actually
     #         taken by each dataset, instead of arbitrarily picking a max datasets to cache.
     #         Also, this should be controlled by a configuration parameter.
-    MAX_CACHED = 3
+    MAX_CACHED = 5
+
+    @staticmethod
+    def set_max_datasets(max_cached):
+        MatrixDataCacheManager.MAX_CACHED = max_cached
 
     # FIXME:   If the number of active datasets exceeds the MAX_CACHED, then each request could
     # lead to a dataset being deleted and a new only being opened: the cache will get thrashed.
@@ -122,11 +126,12 @@ class MatrixDataCacheManager(object):
                 loader = MatrixDataLoader(location, app_config=app_config)
                 cache_item = MatrixDataCacheItem(loader)
                 self.datasets[location] = (cache_item, last_accessed)
-        try:
-            data_adaptor = cache_item.acquire(app_config)
-            yield data_adaptor
-        finally:
-            cache_item.release()
+
+            try:
+                data_adaptor = cache_item.acquire(app_config)
+                yield data_adaptor
+            finally:
+                cache_item.release()
 
 
 class MatrixDataType(Enum):
--- a/server/cli/launch.py
+++ b/server/cli/launch.py
@@ -231,7 +231,6 @@ def launch_args(func):
         "--config-file",
         "-c",
         "config_file",
-        is_flag=True,
         default=None,
         show_default=True,
         help="Location to yaml file with configuration settings",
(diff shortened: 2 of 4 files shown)


server/data_common/matrix_loader.py
```diff
@@ -122,11 +126,12 @@ def data_adaptor(self, location, app_config):
...
                 self.datasets[location] = (cache_item, last_accessed)
-        try:
-            data_adaptor = cache_item.acquire(app_config)
-            yield data_adaptor
-        finally:
-            cache_item.release()
+
+            try:
+                data_adaptor = cache_item.acquire(app_config)
+                yield data_adaptor
```
bkmartinjr: yielding will leave the lock held, which single-threads any caller that uses this as a generator.

do we actually use it this way?  It looks like the two invocations in app.py use it differently (one within a `with`, one just as a function).  I believe it will single thread callers through the first due to the fact that yield does not release the lock.

Issue #1209, opened 2020-03-10
Clean up value.js file


colinmegill:
As we've added annotations, `client/src/components/categorical/value.js` has increased in size and complexity. Time to refactor.

bkmartinjr:
Colin resolved this in a recent PR

PR #1278, opened 2020-03-23
add/fix label tooltips

### LINKED ISSUE #1274: truncated category names need a tooltip
Category names will be "tilde truncated" when they are too long to display.  

(ie, "a long name" becomes "a lo...name")

These truncated category names need a tooltip.   Labels have a tooltip, but categories do not.

### LINKED ISSUE #1275: long label tooltips can't be read (offscreen)
If a long, truncated label, in a category, is moused over, it displays a tooltip.  However, the tooltip is center justified, which means much of it can't be read as it is clipped by the left side of the browser window.

The entire tooltip should be visible.  

Example:

seve:
This PR adds a tooltip to concatenated category names and also formats label tooltips so they properly pop out to the right.

---
closes #1275 and closes #1274

### DIFF
--- a/client/src/components/categorical/category.js
+++ b/client/src/components/categorical/category.js
@@ -7,7 +7,7 @@ import {
   Button,
   Tooltip,
   Icon,
-  PopoverPosition
+  Position
 } from "@blueprintjs/core";
 import CategoryFlipperLayout from "./categoryFlipperLayout";
 import AnnoMenu from "./annoMenuCategory";
@@ -206,7 +206,7 @@ class Category extends React.Component {
             content={metadataField}
             disabled={truncatedString === null}
             hoverOpenDelay={globals.tooltipHoverOpenDelayQuick}
-            position={PopoverPosition.LEFT}
+            position={Position.LEFT}
             usePortal
             modifiers={{
               preventOverflow: { enabled: false },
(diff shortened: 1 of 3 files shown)


client/src/components/categorical/category.js
```diff
@@ -164,6 +170,11 @@ class Category extends React.Component {
       false
     );
 
+    const truncatedString = maybeTruncateString(
+      metadataField,
+      globals.categoryDisplayStringMaxLength
+    );
+
     return (
```
bkmartinjr: I am OK if we defer to a future PR, but this JSX is out of control -- super hard to understand.  It needs refactoring, either by pulling out bits into separate functions, or sub-components (if reusable).     (ditto for value.js)

client/src/components/categorical/category.js
```diff
@@ -164,6 +170,11 @@ class Category extends React.Component {
       false
     );
 
+    const truncatedString = maybeTruncateString(
```
bkmartinjr: why are value.js and category.js using different functions to truncate the string?

client/src/components/categorical/category.js
seve: Value.js is being addressed in #1209, but I agree, annotations have increased the complexity immensely.  I'll write up an issue for the category component and will take the learnings for #1209 and apply them here.

client/src/components/categorical/category.js
seve: I just learned about this function while working on this PR, I'll be utilizing this function when addressing #1209

PR #1280, opened 2020-03-23
Avoid race condition in the matrix cache handling.

### LINKED ISSUE #1255: load test error

bmccandless:
During the MatrixDataCacheItem acquire function there was a
time when the write lock was released and the read lock was taken.
During that time, the dataset could have been deleted, later
result in the MatrixDataCacheManageri data adaptor returning None.

The solution is to demote the writer lock to a reader lock instead
of unlocking and relocking.

Also, when a the cache needs to delete an entry, the delete
is done outside the MatrixDataCacheManager lock.   This operation
only requires the write lock for the MatrixDataCacheItem.

Fixes #1255

### DIFF
--- a/server/data_common/matrix_loader.py
+++ b/server/data_common/matrix_loader.py
@@ -19,32 +19,33 @@ class MatrixDataCacheItem(object):
         self.data_adaptor = None
         self.data_lock = RWLock()
 
-    def acquire(self, app_config):
-        """returns the data_adaptor if cached.  opens the data_adaptor if not.
-        In either case, the a reader lock is taken.  Must call release when
-        the data_adaptor is no longer needed"""
-
+    def acquire_existing(self):
+        """If the data_adaptor exists, take a read lock and return it, else return None"""
         self.data_lock.r_acquire()
         if self.data_adaptor:
             return self.data_adaptor
 
         self.data_lock.r_release()
-        try:
-            with self.data_lock.w_locked():
-                # the data may have been loaded while waiting on the lock
-                if not self.data_adaptor:
-                    self.loader.pre_load_validation()
-                    self.data_adaptor = self.loader.open(app_config)
+        return None
 
-        except Exception:
-            # necessary to acquire after an exception, since the release will occur when
-            # the context exits
-            self.data_lock.r_acquire()
-            raise
-
+    def acquire(self, app_config):
+        """returns the data_adaptor if cached.  opens the data_adaptor if not.
+        In either case, the a reader lock is taken.  Must call release when
+        the data_adaptor is no longer needed"""
         self.data_lock.r_acquire()
         if self.data_adaptor:
             return self.data_adaptor
+        self.data_lock.r_release()
+
+        self.data_lock.w_acquire()
+        # the data may have been loaded while waiting on the lock
+        if not self.data_adaptor:
+            self.loader.pre_load_validation()
+            self.data_adaptor = self.loader.open(app_config)
+
+        # demote the write lock to a read lock.
+        self.data_lock.w_demote()
+        return self.data_adaptor
 
     def release(self):
         """Release the reader lock"""
@@ -65,7 +66,7 @@ class MatrixDataCacheManager(object):
     when the context ends.  This class currently implements a simple least recently used cache,
     which can delete a dataset from the cache to make room for a new oneo
 
-    This is the indended usage pattern:
+    This is the intended usage pattern:
 
            m = MatrixDataCacheManager()
            with m.data_adaptor(location, app_config) as data_adaptor:
@@ -77,7 +78,11 @@ class MatrixDataCacheManager(object):
     #  TODO:  This is very simple.  This can be improved by taking into account how much space is actually
     #         taken by each dataset, instead of arbitrarily picking a max datasets to cache.
     #         Also, this should be controlled by a configuration parameter.
-    MAX_CACHED = 3
+    MAX_CACHED = 5
+
+    @staticmethod
+    def set_max_datasets(max_cached):
+        MatrixDataCacheManager.MAX_CACHED = max_cached
 
     # FIXME:   If the number of active datasets exceeds the MAX_CACHED, then each request could
     # lead to a dataset being deleted and a new only being opened: the cache will get thrashed.
@@ -97,6 +102,8 @@ class MatrixDataCacheManager(object):
     @contextmanager
     def data_adaptor(self, location, app_config):
         # create a loader for to this location if it does not already exist
+
+        delete_adaptor = None
         with self.lock:
             value = self.datasets.get(location)
             if value is not None:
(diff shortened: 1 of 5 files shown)


server/data_common/rwlock.py
```diff
@@ -88,6 +90,12 @@ def w_acquire(self):
     def w_release(self):
         self.w_lock.release()
 
+    def w_demote(self):
```
bkmartinjr: this does not release the write lock, and the caller in matrix_loader does not either.   That will work OK if it is called from within a `with lock.w_locked() as lock:...`, but won't work as currently called.

server/data_common/matrix_loader.py
```diff
@@ -19,32 +19,33 @@ def __init__(self, loader):
         self.data_adaptor = None
         self.data_lock = RWLock()
 
-    def acquire(self, app_config):
-        """returns the data_adaptor if cached.  opens the data_adaptor if not.
-        In either case, the a reader lock is taken.  Must call release when
-        the data_adaptor is no longer needed"""
-
+    def acquire_existing(self):
```
bkmartinjr: this appears to be dead code?

server/data_common/matrix_loader.py
```diff
@@ -115,14 +122,17 @@ def data_adaptor(self, location, app_config):
...
                     oldest_cache = oldest[1][0]
                     oldest_key = oldest[0]
-                    oldest_cache.delete()
                     del self.datasets[oldest_key]
+                    delete_adaptor = oldest_cache
 
                 last_accessed = time.time()
                 loader = MatrixDataLoader(location, app_config=app_config)
                 cache_item = MatrixDataCacheItem(loader)
                 self.datasets[location] = (cache_item, last_accessed)
```
bkmartinjr: The code exits the critical section (`with self.lock`) with the cache_item unlocked (ie, there is no read lock on it). Given this, I believe that it is possible for another thread to remove it from the cache (and call delete()), before the code acquires the read lock.

I think you need to acquire a read lock on the cache item within the critical section (but the yield must be outside the critical section).  It also seems fine to have the delete() outside the critical section.

server/data_common/rwlock.py
bmccandless: The point is not to release the w_lock.   w_lock is perhaps not a great name here, since the reader also hold the w_lock.   The effect I'm going for here is to get into a state that would occur if I had called this atomically.
   rw_lock.w_release()
   rw.local.r_acquire()

It's a little frustrating that python standard library does not have a rwlock.  They are tricky to get right.  I just realized that this implementation has a deadlock possibility.  So, I need to find a better way.   However demoting the writer to a reader is the right solution I believe

bkmartinjr: you are absolutely right!   apologies, will resolve this conversation.

server/data_common/matrix_loader.py
bmccandless: Kind of.  I  was thinking along the lines of your other comment about taking a read lock before leaving the critical section.  But then I didn't complete that part.  It's done now and fixes that issue.

server/data_common/matrix_loader.py
bmccandless: Agreed, fixed.

server/data_common/matrix_loader.py
```diff
@@ -19,32 +19,33 @@ def __init__(self, loader):
...
-                    self.data_adaptor = self.loader.open(app_config)
-
-        except Exception:
-            # necessary to acquire after an exception, since the release will occur when
-            # the context exits
-            self.data_lock.r_acquire()
-            raise
+        return None
 
+    def acquire(self, app_config):
```
bkmartinjr: you might consider naming this something more descriptive, like `acquire_and_load()`

server/data_common/matrix_loader.py
```diff
@@ -97,13 +102,19 @@ def __init__(self):
...
                 cache_item = value[0]
                 last_accessed = time.time()
                 self.datasets[location] = (cache_item, last_accessed)
-            else:
+                data_adaptor = cache_item.acquire_existing()
+
+            if data_adaptor is None:
                 while True:
                     # find the last access times for each loader
                     items = list(self.datasets.items())
```
bkmartinjr: for performance, I would move the sorted() to below the test for `< self.MAX_CACHED`.  No reason to sort if you aren't going to eject something

bmccandless: Done

server/data_common/matrix_loader.py
bmccandless: Done, called it acquire_and_open

PR #1283, opened 2020-03-23
fix health root exception

bkmartinjr:
the new config feature introduced a bug in the health root (new config names were not propagated to this code).

### DIFF
--- a/server/common/health.py
+++ b/server/common/health.py
@@ -24,7 +24,7 @@ def health_check(config):
     health = {"status": None, "version": "1", "releaseID": cellxgene_version}
 
     checks = [
-        lambda: config.single_dataset__datapath is not None or config.multi_dataset__dataroot is not None,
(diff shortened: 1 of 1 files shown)


server/common/health.py
```diff
@@ -23,7 +23,11 @@ def health_check(config):
     """
     health = {"status": None, "version": "1", "releaseID": cellxgene_version}
 
-    checks = [_is_accessible(config.datapath), _is_accessible(config.dataroot)]
+    checks = [
+        lambda: config.single_dataset__datapath is not None or config.multi_dataset__dataroot is not None,
```
bmccandless: I'm not sure I understand the lambda,   I think this produces a function , which always evaluates to True.
I think you just want just the expression: ```config.single_dataset__datapath is not None or config.multi_dataset__dataroot is not None```

bkmartinjr: indeed.  thanks for catching!

Issue #1287, opened 2020-03-24
invalid URLs can be reflected by the redirect page


bkmartinjr:
Redirect should not reflect invalid URLs

PR #1293, opened 2020-03-25
Various hardening to REST routes

### LINKED ISSUE #1242: /data/var/ limits
for anti-abuse, we should consider limiting the total size of data that can be requested from /data/var (and perhaps /data/obs). 

For example, no more than 10 columns in any given request.

Anything too  big will run into the 2GB flatbuffer memory limit, and is almost certainly an error.  The current front-end requests small number at a time.

### LINKED ISSUE #1249: 404 on favicon for all datasets in dataroot
load any dataset in the dataroot, and the front-end gets a 404 requesting /<dataset>/favicon.png

bkmartinjr:
Several changes:
* for the EB config, handle static route with a URL rewrite rather than a redirect.  Redirect left intact to support CLI and other deployments.   This should resolve #1249 and #1287, but this needs confirmation and testing.
* Add limits on the size of requests to /annotations, /data and /diffexp routes, to prevent excess resource allocations.   As part of this, added a nascent "quota" system to AppConfig. Currently not exposed via external config, but we can add that as it is useful.   Fixes #1242

### DIFF
--- a/server/common/app_config.py
+++ b/server/common/app_config.py
@@ -90,7 +90,7 @@ class AppConfig(object):
         except KeyError as e:
             raise ConfigurationError(f"Unexpected config: {str(e)}")
 
-        # quotas are used for various limits, eg, size of requests.  Not currently configurable.
+        # Used for various limits, eg, size of requests.  Not currently configurable.
         self.limits = Default_Limits
 
         # The annotation object is created during complete_config and stored here.
--- a/server/common/app_config.py
+++ b/server/common/app_config.py
@@ -21,8 +21,8 @@ DEFAULT_SERVER_PORT = int(environ.get("CXG_SERVER_PORT", "5005"))
 # anything bigger than this will generate a special message
 BIG_FILE_SIZE_THRESHOLD = 100 * 2 ** 20  # 100MB
 
-""" Default quotas """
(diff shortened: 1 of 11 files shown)


server/common/app_config.py
```diff
@@ -21,6 +21,18 @@
 # anything bigger than this will generate a special message
 BIG_FILE_SIZE_THRESHOLD = 100 * 2 ** 20  # 100MB
 
+""" Default quotas """
+Default_Quotas = {
```
bmccandless: I think the client needs to be aware of these limits so that we can provide proper feedback to the user.  Right now I think they just get back "BadRequest" if a limit is exceeded.   I suggest that we add these quotas to the config endpoint.

bmccandless: I see the response does have some descriptive information about the error.  But I still suggest that we need an tracker issue to tell the client what the limits are so the client code to warn the user that they've selected too many cells, for example.

bkmartinjr: True. Right now, I picked defaults that reflect what the front-end is actually doing, on the assumption we could co-evolve these.  If they ever become externally configurable, letting the client know is critical.

I'll go ahead and add to the config route for now.  Simple change...

PR #1300, opened 2020-03-26
Add TOS and Privacy url via config

colinmegill:
This PR adds, conditioned on their being present in config, two menu items with linkouts to external webpages for `tos` and `privacy`

image


client/src/components/menubar/index.js
```diff
@@ -32,7 +32,9 @@ import UndoRedoReset from "./undoRedo";
   aboutLink: state.config?.links?.["about-dataset"],
   disableDiffexp: state.config?.parameters?.["disable-diffexp"] ?? false,
   diffexpMayBeSlow: state.config?.parameters?.["diffexp-may-be-slow"] ?? false,
-  showCentroidLabels: state.centroidLabels.showLabels
+  showCentroidLabels: state.centroidLabels.showLabels,
+  tosURL: state.config?.about_legal_tos,
```
bkmartinjr: these are wrong - should be pulling from parameters.

Ie. `state.config?.parameters?.about_legal_tos`, etc.

PR #1302, opened 2020-03-26
Change route to the datasets

bmccandless:
Previously is was /<dataset>/...
Now there is an extra level:  /d/<dataset>/...


bkmartinjr:
I found the root cause of the favicon issue, and suggest we fix it in a later PR.  Am approving this one.

Issue #1310, opened 2020-03-28
backend fails when using latest s3fs package, v 0.4.1


bkmartinjr:
It works fine with s3fs==0.4.0.  Pinning to 0.4.0.  Remove pin when resolved.

Filed: dask/s3fs#303

Example failure:
```
# python
Python 3.8.2 (default, Feb 26 2020, 04:23:39) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import s3fs
>>> import fsspec
>>> fs = fsspec.filesystem('s3', config_kwargs={'region_name':'us-east-1'})
>>> fs.info('s3://czi.cxg-public/cxg/') # works
{'Key': 'czi.cxg-public/cxg', 'Size': 0, 'StorageClass': 'DIRECTORY', 'type': 'directory', 'size': 0, 'name': 'czi.cxg-public/cxg'}
>>> fs.info('s3://czi.cxg-public/cxg/') # fails
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/bruce/projects/cellxgene/venv38/lib/python3.8/site-packages/s3fs/core.py", line 525, in info
    if self.version_aware or (key and self._ls_from_cache(path) is None) or refresh:
  File "/mnt/c/Users/bruce/projects/cellxgene/venv38/lib/python3.8/site-packages/fsspec/spec.py", line 313, in _ls_from_cache
    raise FileNotFoundError(path)
FileNotFoundError: s3://czi.cxg-public/cxg/
```

to reproduce, on s3fs==0.4.1:
```
import s3fs
import fsspec
fs = fsspec.filesystem('s3', config_kwargs={'region_name':'us-east-1'})
fs.info('s3://czi.cxg-public/cxg/') # works
fs.info('s3://czi.cxg-public/cxg/') # fails
```

PR #1312, opened 2020-03-28
various fixes for s3fs use

bkmartinjr:
Several changes to make s3fs work better in the hosted configuration:
* pin package version to 0.4.0.  See #1310 
* add a configuration option to set the default S3 region for use in DataLocator
* use the region configuration in DataLocator

The latter two prevent the DataLocator from having to do a round-trip through the legacy s3 region lookup.

### DIFF
--- a/server/common/default_config.py
+++ b/server/common/default_config.py
@@ -60,6 +60,10 @@ diffexp:
   enable: true
   lfc_cutoff: 0.01
 
+data_locator:
+  s3:
+    region_name: us-east-1
+
 adaptor:
   cxg_adaptor:
     tiledb_ctx:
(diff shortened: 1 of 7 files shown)


server/common/default_config.py
```diff
@@ -60,6 +60,10 @@
   enable: true
   lfc_cutoff: 0.01
 
+data_locator:
```
bmccandless: One improvement, I think, would be to have one place to define the region name.
Right now I believe we need to set the region in the data_locator and the region in the tiledb context.

bkmartinjr: > Looks good to me, but there may be an improvement to consider to reuse the same region name
> for both the data locator and tiledb.

Yep, I thought about this but ran into some uncertainty about how best to do it.  How about I'll land this PR, but file an issue to do the cleanup in the near future?

PR #1324, opened 2020-04-01
work-around for column name encoding bug

bkmartinjr:
Add a work-around to cxgtool for a tiledb bug that does not correctly escape attribute/column names containing a `/`.  These characters will be replaced with an underscore, and then the names will be de-duped in the case that creates duplicate names.

### DIFF
--- a/server/converters/cxgtool.py
+++ b/server/converters/cxgtool.py
@@ -442,10 +442,18 @@ def sanitize_df(df):
     df.rename(columns=sanitize_keys(df.keys().tolist()), inplace=True)
 
 
+def sanitize_mapping(mapping):
+    clean_keys = sanitize_keys([k for k in mapping.keys()])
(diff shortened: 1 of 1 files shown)


server/converters/cxgtool.py
```diff
@@ -385,5 +393,64 @@ def save_metadata(container, metadata):
...
+    used_keys = set()
+    clean_unique_keys = {}
+    for k, v in clean_keys.items():
+        if v not in used_keys:
+            used_keys.add(v)
+            clean_unique_keys[k] = v
+            continue
+
+        # else, needs deduping.
+        counter = 1
```
bmccandless, line 423: I think you are missing a "counter += 1" inside the while loop.

bkmartinjr: nothing more embarrassing than an infinite loop :-(

Thanks for the catch!

PR #1330, opened 2020-04-01
refactor to remove non-standard terms in centroid files

seve:
Scanned the centroid files and replaced incorrectly terms with category/label as fits.

### DIFF
--- a/client/__tests__/util/centroid.test.js
+++ b/client/__tests__/util/centroid.test.js
@@ -69,7 +69,7 @@ describe("centroid", () => {
     });
   });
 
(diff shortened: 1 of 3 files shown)


seve:
@bkmartinjr I noticed that categorical selection still uses the term categoryValue. Should I write up an issue for that?

bkmartinjr:
> @bkmartinjr I noticed that categorical selection still uses the term categoryValue. Should I write up an issue for that or should I refactor and propagate?

Separate issue is fine w/ me.

PR #1333, opened 2020-04-01
add cache control headers

bkmartinjr:
This PR adds cache control headers to all routes.

They are enabled/disabled by a new configuration variable, `generate_cache_control_headers` which is off by default for backward compatibility in the CLI.

When it is on, the `Cache-Control` header is set as follows:
* the `/health` sets `no-cache`
* all PUT & POST routes set `no-cache`
* GET routes will set `public, max-age=3600`


bkmartinjr:
>  Either you can fix this now, or I can do it when I add the check.

Will fix.

PR #1335, opened 2020-04-02
Add eslint steps to lint job

seve:
This PR adds two conditional steps to the `lint` job within the `push_test` workflow.  The first step, `lint all with eslint`, is entered if the job is activated on master. Otherwise, if activated on a non-master branch, the step `Lint all with eslint` is activated only calling eslint on the diff files between master and the other branch.

### DIFF
--- a/.github/workflows/push_tests.yml
+++ b/.github/workflows/push_tests.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           pip install flake8
           make lint
-       - name: Lint all with eslint
+      - name: Lint all with eslint
         if: github.ref == 'refs/heads/master'
         run: |
            make lint-client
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -77,6 +77,14 @@ fmt-py:
 lint:
```
mweiden, line 77: For completeness, I'd rename this `lint-server` and then add this below:

```
.PHONY: lint
lint: lint-server lint-client
```

.github/workflows/push_tests.yml
```diff
@@ -10,100 +10,117 @@ jobs:
...
+        uses: actions/cache@v1
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+      - name: Lint with flake8
+        run: |
+          pip install flake8
+          make lint
```
mweiden, line 30: If you make the change below you'll have to change this to `lint-server` as well.

PR #1336, opened 2020-04-02
diffexp limit UI and configuration

### LINKED ISSUE #1328: front-end informs user of diffexp cell limits, does not fail ungracefully when they are hit

bkmartinjr:
Expose the differential expression cell count limit via UI, and make it configurable on the backend.

Fixes #1328


bkmartinjr:
I have reviewed the front-end elements, which are courtesy of @colinmegill (thank you!)

PR #1340, opened 2020-04-02
Improvements to the matrix cache

### LINKED ISSUE #1322: Improve caching behavior and error handling when underlying data store changes
- Matrix adaptor cache timeout.  configurable.
- handling IO errors so we "notice" when S3 datasets are gone.
- more predictable behavior when datasets are gone

Example of current issue: if a dataset exists, and is loaded into the matrix cache, and then later disappears, some routes generate odd failures.   Test case - data set is deleted after being loaded once:
* /config - continues to work
* /schema - *mostly* continues to work, but reports no layouts.  Ie, partial, corrupt schema information, which confuses the front-end
* some /annotations and /layout routes continue to work if they are in the tile cache

Have not yet tested what happens when something isn't in the tile cache, but presumably it generates an error.

Ideally we would have predictable behavior within a single route - eg, the schema would be either an error or fully correct.  Across routes & time, I don't think we can make guarantees - 4XX errors need to happen when they are discovered, and the front-end needs to improve its error handling UI.

bmccandless:
- Add a timelimit for the matrix in the cache.
Once the timelimit is reached, the matrix can be removed.

- If a DatasetAccessError occurs, then remove the dataset
from the matrix cache.

Fixes #1322

### DIFF
--- a/server/data_cxg/cxg_adaptor.py
+++ b/server/data_cxg/cxg_adaptor.py
@@ -251,8 +251,7 @@ class CxgAdaptor(DataAdaptor):
             pemb = self.get_path("emb")
             embeddings = [os.path.basename(p) for (p, t) in self.lsuri(pemb) if t == "array"]
         if len(embeddings) == 0:
-            # something went wrong. perhaps the dataset was removed.  re-validate
-            self.pre_load_validation(self.data_locator)
+            raise DatasetAccessError("cxg matrix missing embeddings")
         return embeddings
 
     @staticmethod
--- a/server/data_cxg/cxg_adaptor.py
+++ b/server/data_cxg/cxg_adaptor.py
@@ -250,6 +250,9 @@ class CxgAdaptor(DataAdaptor):
         with ServerTiming.time(f"layout.lsuri"):
             pemb = self.get_path("emb")
             embeddings = [os.path.basename(p) for (p, t) in self.lsuri(pemb) if t == "array"]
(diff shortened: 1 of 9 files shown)


server/data_cxg/cxg_adaptor.py
```diff
@@ -250,6 +250,9 @@ def get_embedding_names(self):
         with ServerTiming.time(f"layout.lsuri"):
             pemb = self.get_path("emb")
             embeddings = [os.path.basename(p) for (p, t) in self.lsuri(pemb) if t == "array"]
+        if len(embeddings) == 0:
+            # something went wrong. perhaps the dataset was removed.  re-validate
+            self.pre_load_validation(self.data_locator)
```
bkmartinjr, line 255: is there ever a case where pre_load_validation will not throw?  At which point the embedding are still potentially wrong, OR the dataset actually has no embeddings (which is rare, but in theory could occur)

If we know this is a hard error, we should just raise it here.  If it is spurious, and embeddings will exist, then they need to reload/retry.   If embeddings are optional, then the real issue is that the caller needs to handle this check (ie, /schema endpoint needs to do its own validation that the object exists).

Today, embeddings are required in CXG, as they are an essential part of the app.  So I recommend you just throw an error here....

bmccandless: Sounds good.  fixed.

PR #1342, opened 2020-04-02
categorical type handling fix

bkmartinjr:
There were two issues related to categorical type handling that were triggered when a category had a non-string type (eg, an integer category):
- the schema created by cxgtool was incorrect (encoded as the primitive type, rather than categorical).  This fix makes the anndata adaptor and cxgtool use the same semantics, so we should see the same app behavior when loading an H5AD or a CXG derived from the H5AD.
- the JS dataframe summarization code would incorrectly assume it should summarize based upon the underlying type, which confused our categorical controls.  This change lets the category helper code directly request a categorical summarization.

### DIFF
--- a/server/converters/cxgtool.py
+++ b/server/converters/cxgtool.py
@@ -295,7 +295,6 @@ def save_dataframe(container, name, df, index_col_name, ctx):
         for k, v in df.items():
             dtype, hints = cxg_type(v)
             value[k] = v.to_numpy(dtype=dtype)
-            print(k, dtype, hints)
             if hints:
                 schema_hints.update({k: hints})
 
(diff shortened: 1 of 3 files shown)


server/converters/cxgtool.py
```diff
@@ -277,6 +295,7 @@ def save_dataframe(container, name, df, index_col_name, ctx):
         for k, v in df.items():
             dtype, hints = cxg_type(v)
             value[k] = v.to_numpy(dtype=dtype)
+            print(k, dtype, hints)
```
bmccandless: debugging statement

bkmartinjr: fixed

PR #1350, opened 2020-04-05
Cleanup front-end build and python module contents

### LINKED ISSUE #1105: CLI python package contains extraneous files
There are a number of files included in the current package that are unused/unneeded.  Some are obsolete python code, some are front-end assets _also_ included in the index.html (eg, the CSS files, the main.js, etc).   We should clean up the package and audit the result to make sure we only have what is necessary.

This may also point out some clean up we could do in the build process.

### LINKED ISSUE #1303: favicon double inserted in index.html by prod build
prod build double-inserts the favicon, with two different routes...

this may help: https://github.com/jantimon/favicons-webpack-plugin

### LINKED ISSUE #1351: static assets need to be rev'ed to be CDN friendly

bkmartinjr:
This PR includes several fixes to the production build (should not affect dev build):
* stop double-insertion of favicon link resolving
* remove extraneous (unused) files and from deployment bundles (eg, the server unit test code, etc)
* remove the dead links to source maps in the production build, which caused console warnings in the latest Chrome devtools
* remove the obsolete/unused server worker for precaching
* consolidate all static assets into a single directory, and give them CDN-friendly (reved) names

As part of doing this, the build scripts were updated, and the JS dependencies updated.

Fixes #1303 
Fixes #1105 
Fixes #1351


client/configuration/webpack/webpack.config.prod.js
```diff
@@ -108,14 +94,28 @@ module.exports = {
...
+          windows: false,
+          yandex: false
+        }
+      }
     }),
-    new SWPrecacheWebpackPlugin({
-      cacheId: "cellxgene",
-      filename: "service-worker.js"
-    })
+    new HtmlWebpackInlineSourcePlugin(HtmlWebpackPlugin),
```
mweiden, line 117: Minor note here. The repo for the `HtmlWebpackInlineSourcePlugin` says that it is [no longer maintained](https://github.com/DustinJackson/html-webpack-inline-source-plugin/blob/master/README.md) and asks people to use [this Facebook plugin](https://github.com/facebook/create-react-app/blob/edc671eeea6b7d26ac3f1eb2050e50f75cf9ad5d/packages/react-dev-utils/InlineChunkHtmlPlugin.js#L10). Would it be worthwhile to upgrade this now?

common.mk
```diff
@@ -22,12 +22,9 @@ export JEST_ENV
...
 # $(2) is the destination
 define copy_client_assets
-	mkdir -p $(2)/common/web/static/img
-	mkdir -p $(2)/common/web/static/js
+	mkdir -p $(2)/common/web/static/assets
 	mkdir -p $(2)/common/web/templates/
 	cp $(1)/index.html $(2)/common/web/templates/
 	cp -r $(1)/static $(2)/common/web/
-	cp $(1)/favicon.png $(2)/common/web/static/img
-	cp $(1)/service-worker.js $(2)/common/web/static/js/
```
mweiden, line 31: nice cleanup 👍

client/configuration/webpack/webpack.config.prod.js
bkmartinjr: yeah, noticed that.  I'd rather defer as this is known to work fine with our code base.

PR #1352, opened 2020-04-06
load embeddings in parallel

### LINKED ISSUE #1365: parallel load embeddings
currently embeddings are loaded en masse.  We should move to a per-embedding HTTP request to improve overall parallelism in the backend, and reduce the risk that we run into the FBS max buffer size down the road.

bkmartinjr:
This PR moves our embedding data load from a single big HTTP request for all embeddings to a per-embedding HTTP request.  This should give us more parallelism in deployments that have back-end parallelism, and speed up initial data loading.

Fixes #1365

### DIFF
--- a/server/common/rest.py
+++ b/server/common/rest.py
@@ -10,6 +10,7 @@ from server.common.errors import (
     PrepareError,
     DisabledFeatureError,
     ExceedsLimitError,
+    DatasetAccessError,
 )
 
 import json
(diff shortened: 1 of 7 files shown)


server/common/rest.py
```diff
@@ -179,14 +180,21 @@ def diffexp_obs_post(request, data_adaptor):
 
 
 def layout_obs_get(request, data_adaptor):
+    fields = request.args.getlist("layout-name", None)
```
mweiden, line 183: Would wrapping this in a `set(...)` help protect against clients/abusers inflating the number of requests?

server/common/rest.py
```diff
@@ -179,14 +180,21 @@ def diffexp_obs_post(request, data_adaptor):
...
+    num_columns_requested = len(data_adaptor.get_embedding_names()) if len(fields) == 0 else len(fields)
+    if data_adaptor.config.exceeds_limit("column_request_max", num_columns_requested):
```
mweiden, line 185: Wouldn't the limit be the min of `column_request_max` and the number of embeddings given by the `data_adaptor`?

server/common/rest.py
bkmartinjr: how so?  URLs have a fixed length limit (number of characters), so in practice it can't ever be that big....

server/common/rest.py
bkmartinjr: not for this check - this is to simply do a resource check.  The existence check for the embedding comes later.

I'd like to keep these concerns separate for clarity.

PR #1363, opened 2020-04-06
Update brushable histogram x-tick number format

### LINKED ISSUE #1349: Incorrect x axis tick label for continuous metadata histogram
The tick label for the last (rightmost) tick can sometimes be incorrect. For example, the rightmost tick should be 150m for the 1st plot and 250m for the 2nd plot below. Also, showing "xxxm" seems rather non-intuitive compared to "0.xxx".

image

image

mweiden:
Fixes #1349

### Repeating tick mark issue
The issue was that significant digits that differentiated consecutive tick marks were cropped in number string formatting.

**Strategy to fix number formatting**: increase the number of significant digits shown in scientific notation number formatting to 2 (format `.2e`) instead of 1 (format `.1s`).

For more information see https://github.com/d3/d3-format

Note that does not _fully_ fix the issue described in #1349, but rather makes the formatting issue far less likely. It is _still_ possible for this to occur if the difference between two ticks in axes happens in the a significant digit cropped by the scientific notation format.

### `milli` notation (`123m`) is not intuitive
Users reported that using an `m` to denote a number (`trueValue = 0.001 * displayNumber`) was not intuitive. This PR changes the behavior of the formatting logic to use unabbreviated numbers when the domain of a variable is `-10000 < min < max < 10000` and use scientific notation otherwise. We let `d3` determine xticks for decimals since it generally choses values without many significant digits.

## Results

The following are images show improvements over the example problems in #1349:

Screen Shot 2020-04-06 at 5 44 24 PM

_litvinukova20_restricted.processed.h5ad_

Screen Shot 2020-04-06 at 5 42 22 PM

_lukassen20_airway_orig.processed.h5ad_

### DIFF
--- a/client/__tests__/e2e/data.js
+++ b/client/__tests__/e2e/data.js
@@ -41,7 +41,7 @@ export const datasets = {
         {
           metadata: "n_genes",
           "coordinates-as-percent": { x1: 0.25, y1: 0.5, x2: 0.55, y2: 0.5 },
-          count: "1552"
+          count: "1537"
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


mweiden:
Got positive feedback on this from users. Merging this. Will make further adjustments in the lead up to 0.16.0, if necessary.

PR #1364, opened 2020-04-06
Upgrade d3 to v5.15.1

mweiden:


mweiden:
The issue that was motivating this fix (https://github.com/chanzuckerberg/cellxgene/issues/1349) was solved a different way, so we can WONTFIX this PR.

PR #1371, opened 2020-04-08
CXG performance improvements

bkmartinjr:
Several performance optimizations in this PR, all to the CXG adaptor:
* add better caching to IO intensive or high latency paths, including schema creation, array open, and lsuri.
* do not load entire annotation (obs or var) when requester only wants specific columns
* improve lock concurrency in array handling

In addition, there are some minor test changes to make them insensitive to column ordering in dataframes.

### DIFF
--- a/server/common/immutable_kvcache.py
+++ b/server/common/immutable_kvcache.py
@@ -11,14 +11,44 @@ class ImmutableKVCache(MutableMapping):
     def __init__(self, factory):
         self.factory = factory
         self.lock = threading.Lock()
+        self.factory_calls = {}
         self.cache = {}
         super().__init__()
 
     def __getitem__(self, key):
-        if key not in self.cache:
-            with self.lock:
-                if key not in self.cache:
+        if key in self.cache:
+            return self.cache[key]
+
+        # we need to call factory.  First grab the main lock and the per-key CV.
+        factory_calls = None
+        with self.lock:
+            if key in self.cache:
+                return self.cache[key]
+            if key not in self.factory_calls:
+                creation_thr = True
+                self.factory_calls[key] = {'cv': threading.Condition(), 'is_done': False, 'error': None}
+            factory_calls = self.factory_calls[key]
+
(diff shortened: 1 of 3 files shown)


server/common/immutable_kvcache.py
```diff
@@ -11,14 +11,44 @@ class ImmutableKVCache(MutableMapping):
...
+                return self.cache[key]
+            if key not in self.factory_calls:
+                creation_thr = True
+                self.factory_calls[key] = {'cv': threading.Condition(), 'is_done': False, 'error': None}
+            factory_calls = self.factory_calls[key]
+
+        # with the CV, create the value (or wait for it to be created)
+        cv = factory_calls['cv']
+        with cv:
+            if creation_thr:
```
bmccandless, line 35: Is create_thr guaranteed to be set.  For example if key was in self.factor_calls (line 27) it might not be.   You may need a create_thr=False somewhere.

server/common/immutable_kvcache.py
```diff
@@ -11,14 +11,44 @@ class ImmutableKVCache(MutableMapping):
...
+            else:
+                """ wait for the value to be available """
+                while not factory_calls['is_done']:
+                    cv.wait()
+
+        with self.lock:
+            if key in self.factory_calls:
+                del self.factory_calls[key]
+
         return self.cache[key]
```
bmccandless, line 52: This could get a KeyError if __delitem__ was called between when this was set and when it was accessed here.
Maybe access to self.cache should have a lock.

server/common/immutable_kvcache.py
bkmartinjr: fixed!  good eye.

server/common/immutable_kvcache.py
bkmartinjr: as the cxg adaptor uses it, it won't ever have this issue (it is only called from CxgAdaptor.clean()).  But to make this general-purpose, you have a great point.

PR #1375, opened 2020-04-09
Render categories with one label without checkbox

colinmegill:
If a category has one label, render it without a checkbox. If it is writeable (user generated anno), render it normally.

image

### DIFF
--- a/client/src/components/categorical/categorical.js
+++ b/client/src/components/categorical/categorical.js
@@ -119,6 +119,42 @@ class Categories extends React.Component {
     }
   }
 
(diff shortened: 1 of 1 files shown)


client/src/components/categorical/categorical.js
```diff
@@ -176,10 +176,18 @@ class Categories extends React.Component {
           }
         />
 
+        {/* Categories with only one field */}
+        <div style={{marginLeft: 0}}>
+          {allCategoryNames.map(catName =>
```
bkmartinjr, line 181: can you break this out into a helper function.  This kind of inline code is hard to follow.

PR #1376, opened 2020-04-09
Clean up main Makefile

mweiden:
* Remove unnecessary indirect alias for `build-cli`
* Remove target for installing from build. This doesn't work and does not have a known use case.


mweiden:
@bkmartinjr good catch. Fixed.

PR #1377, opened 2020-04-10
Refactor developer convenience scripts

mweiden:
* Move the `backend-dev` target in `client/Makefile` to its own script (`scripts/backend_dev`), thereby better decoupling the `client/` from build targets elsewhere in the repo
* Remove the client `test` target, this doesn't have much practical use anymore
* Remove the `backend-dev-anno-ontology` target. Turning this on through the convenience scripts is still possible with `CXG_OPTIONS`, which you had to do anyway to run it through `backend-dev`
* Add the `scripts/frontend_dev` convenience method, which opens up Terminal, starts the cellxgene in one tab on port `5005`, and the node dev server on port `3000` for hot-reloading development
* Update all the docs

### DIFF
--- a/dev_docs/developer_scripts.md
+++ b/dev_docs/developer_scripts.md
@@ -75,7 +75,7 @@ the FE developer gets the current version of the backend with a single command
 and no knowledge of python necessary. It creates and activates a virtual
 environment and installs cellxgene from the current branch.
 
-**Requires** `Python3`, `virtual-env`, `pip`
+**Requires** `Python3.6+`, `virtual-env`, `pip`
 
 **Usage:** from the `$PROJECT_ROOT` directory run `./scripts/backend_dev`
 
--- a/dev_docs/developer_scripts.md
+++ b/dev_docs/developer_scripts.md
@@ -1,66 +1,61 @@
-# Developer convenience methods
+# Developer convenience scripts
 
-## Makefile
+This document describes scripts for accelerating cellxgene development.
 
-Documentation for the `Makefile` targets in the project root directory.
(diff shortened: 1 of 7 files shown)


dev_docs/developer_scripts.md
```diff
@@ -1,122 +1,114 @@
...
 
-**About** This script enables FE developers to run the REST API necessary to back the development server for the front end. It is intended to ensure that the FE developer gets the current version of the backend with a single command and no knowledge of python necessary. It creates and activates a virtual environment and installs cellxgene from the current branch.
+**About** This script enables FE developers to run the REST API necessary to
+back the development server for the front end. It is intended to ensure that
+the FE developer gets the current version of the backend with a single command
+and no knowledge of python necessary. It creates and activates a virtual
+environment and installs cellxgene from the current branch.
 
-**Requires** Python3.6 - `virtual-env`, `pip`
+**Requires** `Python3`, `virtual-env`, `pip`
```
bkmartinjr, line 78: we only work with 3.6+

mweiden: good point. Will make this adjustment.

seve:
Is there a way to set a `CXG_TERMINAL` to change what terminal this opens up?

mweiden:
@seve I don't know of a way to generalize that yet, but will give it some thought and follow up with a second PR if I can find a solution.

What terminal are you using? iTerm2?

seve:
> What terminal are you using? iTerm2?

I'm using Hyper(https://hyper.is/)

mweiden:
Hmmm, OK. Thinking.

PR #1384, opened 2020-04-11
Specialize diffexp for tiledb

bmccandless:
This patch adds a new diffexp algorithm which is tuned for tiledb.
This algorithm was written by Bruce and is adapted here to plug into the
current framework.  The anndata_adaptor still calls the original
algotithm (which was move from diffexp.py to diffexp_generic.py).
The cxg_adaptor now calls the new diffexp_tiledb version.  Some
code is shared between the two.

Further tuning and global throttles are still needed.

A script to run and time diffexp with various options is also
added: compute/run_diffexp.py.


server/compute/diffexp_generic.py
```diff
@@ -2,42 +2,7 @@
...
-        else:
-            mean = X.mean(axis=0)
-            dfm = X - mean
-            sumsq = np.sum(np.multiply(dfm, dfm), axis=0)
-            v = sumsq / (n - 1)
-
-    if fp_err_occurred:
-        mean[np.isfinite(mean) == False] = 0  # noqa: E712
-        v[np.isfinite(v) == False] = 0  # noqa: E712
-    return mean, v, n
```
bkmartinjr, line 37: need to merge with master and pick up the fixes for NaN handling

server/compute/run_diffexp.py
```diff
@@ -0,0 +1,88 @@
+import sys
```
bkmartinjr, line 1: seems like this belongs in the test directory, or somewhere else that will prevent it from being shipped as part of the package.

PR #1386, opened 2020-04-11
s3 region should have a single config param

bmccandless:
The s3 region can also now be automatically determined to further
reduce errors.

This patch also fixes a bug with order of handling the config params.
The tiledb config needs to be fixed before attempting to load
(need to handle_adaptor before handle_single_dataset)

### DIFF
--- a/server/common/default_config.py
+++ b/server/common/default_config.py
@@ -70,7 +70,7 @@ data_locator:
   s3:
     # s3 region name.
     #   if true, then the s3 location is automatically determined from the datapath or dataroot.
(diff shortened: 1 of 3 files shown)


server/common/default_config.py
```diff
@@ -68,14 +68,20 @@
 
 data_locator:
   s3:
-    region_name: us-east-1
+    # s3 region name.
+    #   if true, then the s3 location is automatically determined from the datapath or dataroot.
+    #   if false/null, then do not seta.
```
bkmartinjr, line 73: typo:  "seta"

PR #1388, opened 2020-04-13
Improve diffexp for tiledb

### LINKED ISSUE #1327: parallel diffexp implementation
add parallelism performance improvements to differential expression

bmccandless:
- The rows from the A and B sets are gathered and processed at the same time.  In this
  was the matrix is only accessed once instead of twice for each tile.
- There is now a single thread queue that gets shared between all callers of the diffexp.
  This will slow down work if diffexp gets too busy.
- There is a target_workunit amount of work given to each thread.  Previously the
  workunit was (rows selected * width of tile), which could be small.  Now multiple
  column tiles can be combined into one workunit.  If the target is too small then
  thread and other overheads may reduce performance.  If target_workunit is too large
  then the size of the gathered sub matrix may take up too much memory.
- add configuration parameters (max_workers, target_workunit)

Fixes #1327

### DIFF
--- a/server/common/app_config.py
+++ b/server/common/app_config.py
@@ -1,6 +1,6 @@
 from server import __version__ as cellxgene_version
 from flatten_dict import flatten
-from os import mkdir, environ
+import os
 from os.path import splitext, basename, isdir
 import sys
 from urllib.parse import urlparse
@@ -16,8 +16,9 @@ from server.common.utils import find_available_port, is_port_available
 import warnings
 from server.common.annotations import AnnotationsLocalFile
 from server.common.utils import custom_format_warning
+import server.compute.diffexp_tiledb as diffexp_tiledb
 
-DEFAULT_SERVER_PORT = int(environ.get("CXG_SERVER_PORT", "5005"))
+DEFAULT_SERVER_PORT = int(os.environ.get("CXG_SERVER_PORT", "5005"))
 # anything bigger than this will generate a special message
 BIG_FILE_SIZE_THRESHOLD = 100 * 2 ** 20  # 100MB
 
@@ -85,6 +86,8 @@ class AppConfig(object):
             self.diffexp__enable = dc["diffexp"]["enable"]
             self.diffexp__lfc_cutoff = dc["diffexp"]["lfc_cutoff"]
             self.diffexp__top_n = dc["diffexp"]["top_n"]
+            self.diffexp__alg_tiledb__max_workers = dc["diffexp"]["alg_tiledb"]["max_workers"]
+            self.diffexp__alg_tiledb__target_workunit = dc["diffexp"]["alg_tiledb"]["target_workunit"]
 
             self.data_locator__s3__region_name = dc["data_locator"]["s3"]["region_name"]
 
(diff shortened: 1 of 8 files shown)


server/common/app_config.py
```diff
@@ -442,6 +447,19 @@ def handle_diffexp(self, context):
                         f"running differential expression may take longer or fail."
                     )
 
+        max_workers = self.diffexp__alg_tiledb__max_workers
+        if type(max_workers) == str:
+            cpu_count = os.cpu_count()
+            try:
+                max_workers = eval(max_workers, {"cpu_count": cpu_count})
```
bkmartinjr, line 454: having an eval in the code makes me very squeamish from a reliability and infosec standpoint....   Is there a reason we need general purpose scripting here, rather than a much simpler set of params (eg, "max_threads" and "cpu_multiplier"?)

server/common/app_config.py
```diff
@@ -442,6 +447,19 @@ def handle_diffexp(self, context):
                         f"running differential expression may take longer or fail."
                     )
 
+        max_workers = self.diffexp__alg_tiledb__max_workers
```
bkmartinjr, line 450: this should be 'cxg', not 'tiledb'

server/compute/diffexp_tiledb.py
```diff
@@ -1,9 +1,8 @@
-import os
 import concurrent.futures
```
bkmartinjr, line 1: to be consistent with our naming, shouldn't the file be called `diffexp_cxg.py`?  Or perhaps `cxg_diffexp.py`?

server/common/default_config.py
```diff
@@ -66,6 +66,17 @@
   enable: true
   lfc_cutoff: 0.01
   top_n: 10
+  alg_tiledb:
```
bkmartinjr, line 69: see other comment about naming consistency:  s/tiledb/cxg/g

server/common/default_config.py
```diff
@@ -66,6 +66,17 @@
...
+    # The number of threads in the thread pool.
+    # This can be an integer, or a string.
+    # If it is a string, then it is evaluated as a python expression,
+    # where the variable "cpu_count" is provided.
+    # for example:  "max(64,4*cpu_count)"
+    max_workers: max(64, 4 * cpu_count)
```
bkmartinjr, line 75: see other comment about inadvisability of having Turing-complete configuration language.

bkmartinjr:
Additional nit:  please link your PRs to the issues that they resolve by using the `Fixes` workflow.

server/common/app_config.py
bmccandless: Yeah good point.  Not sure what I was thinking

server/compute/diffexp_tiledb.py
bmccandless: Done

server/common/app_config.py
bmccandless: Done

server/common/default_config.py
bmccandless: Done

PR #1390, opened 2020-04-13
performance and correctness fixes to brushable histogram

### LINKED ISSUE #1326: dataset with large (>20) continuous annotation columns renders slowly when color-by mode changes
Eg, Krasnow Lung smartseq2 data has a large number of continuous metrics.  Color-by mode changes are slow (~sec) and profiling shows excessive rendering/DOM manipulation.  Likely need improvements in component render optimization.

### LINKED ISSUE #1389: histogram brushing is not inclusive
the histogram range selection is not inclusive at the top-end, which means that some values the user intends to select are not actually selected.  Often this is inconsquential, except at the very top of the data range where values become impossible to select.

Trivial example where a very chunky range does not allow the highest value to be selected (the purple cells should be selected, but are not):

bkmartinjr:
This PR fixes two issues:
* rendering performance of continuous histograms was very slow due to poor state management in the component.  This caused excessive re-rendering of every histogram when _any_ histogram state changed.
* there was a edge case where the highest value in a continuous range was not selectable

Fixes #1326
Fixes #1389

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -24,10 +24,10 @@ import { makeContinuousDimensionName } from "../../util/nameCreators";
   );
   return {
     world: state.world,
-    scatterplotXXaccessor: state.controls.scatterplotXXaccessor,
-    scatterplotYYaccessor: state.controls.scatterplotYYaccessor,
+    isScatterplotXXaccessor: state.controls.scatterplotXXaccessor === field,
+    isScatterplotYYaccessor: state.controls.scatterplotYYaccessor === field,
     continuousSelectionRange: state.continuousSelection[myName],
-    colorAccessor: state.colors.colorAccessor
+    isColorAccessor: state.colors.colorAccessor === field,
   };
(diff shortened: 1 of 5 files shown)


client/src/components/brushableHistogram/index.js
```diff
@@ -24,10 +24,10 @@ import { makeContinuousDimensionName } from "../../util/nameCreators";
   );
   return {
     world: state.world,
-    scatterplotXXaccessor: state.controls.scatterplotXXaccessor,
-    scatterplotYYaccessor: state.controls.scatterplotYYaccessor,
+    isScatterplotXXaccessor: state.controls.scatterplotXXaccessor === field,
+    isScatterplotYYaccessor: state.controls.scatterplotYYaccessor === field,
     continuousSelectionRange: state.continuousSelection[myName],
-    colorAccessor: state.colors.colorAccessor
+    isColorAccessor: state.colors.colorAccessor === field,
```
colinmegill, line 30: Nice!

PR #1393, opened 2020-04-15
Handle loading state when rendering categories with one label

colinmegill:
This PR properly handles .cxg files when rendering categories with one label differently. It does not presently handle rendering those categories at the 'top' of the list, but renders them inline for now, as we don't know until all categories are loaded which have one label. After discussion, we may handle re-ordering in a subsequent PR.

This PR also refactors the folder structure for categorical metadata for readability. 

image


client/src/components/categorical/category/index.js
```diff
@@ -192,6 +194,20 @@ class Category extends React.Component {
       globals.categoryDisplayStringMaxLength
     );
 
+    if (
+      !schema?.annotations?.obsByName[metadataField]?.writable &&
```
bkmartinjr, line 198: This component also receives `isUserAnno`, which is essentially the same concept & value as `.writable`.   Suggest that you only use one of these if they are the same concept & always have the same value.

client/src/components/categorical/index.js
```diff
@@ -227,32 +176,33 @@ class Categories extends React.Component {
...
-        }
-        {
-          this.maybeRenderAnnotations(
-            allCategoryNames,
-            schema,
-            expandedCats,
-            createAnnoModeActive
-          ) /* Writable fields, ie., anno */
-        }
+        {/* READ ONLY CATEGORICAL FIELDS */}
```
bkmartinjr, line 179: Why pass `isUserAnno` if the <Category> already knows about the schema `.writable`?   (see comment in category/index.js)

client/src/components/categorical/category/index.js
```diff
@@ -7,27 +7,28 @@ import {
...
 import AnnoDialogAddLabel from "./annoDialogAddLabel";
 
-import * as globals from "../../globals";
-import maybeTruncateString from "../../util/maybeTruncateString";
+import * as globals from "../../../globals";
+import maybeTruncateString from "../../../util/maybeTruncateString";
 
-@connect(state => ({
+@connect((state) => ({
   colorAccessor: state.colors.colorAccessor,
```
bkmartinjr, line 21: this has the same bug I just fixed in histograms - it will re-render every time the color accessor changes.

Suggest changing this to a bool: `isColorAccessor`, using the same method I used in brushableHistogram/index.js

bkmartinjr:
ps. please fix lint

PR #1398, opened 2020-04-15
Provide a hook into the AWS Secret Manager for the flask secret key

### LINKED ISSUE #1239: User Sessions: Configuring Flask's secret key uniquely
@niacin wrote:

_We also need to make sure we are configuring Flask’s secret key uniquely using the CXG_SECRET_KEY environment variable in our production environment ... in terms of tasks: you probably should still set CXG_SECRET_KEY to something random in production, just so it isn’t a gotcha (much) later on_

bmccandless:
Also, the secret manager required a region name, so there was some
refactoring around how regions are handled.

Fixes #1239

### DIFF
--- a/server/common/data_locator.py
+++ b/server/common/data_locator.py
@@ -2,6 +2,9 @@ import os
 import tempfile
 import fsspec
 from datetime import datetime
+import boto3
+import botocore
+from urllib.parse import urlparse
 
 
 class DataLocator:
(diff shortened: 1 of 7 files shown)


server/common/data_locator.py
```diff
@@ -125,3 +128,25 @@ def __enter__(self):
     def __exit__(self, *args):
         if self.delete:
             os.unlink(self.tmp_path)
+
+
+def discover_region_name(uri):
```
bkmartinjr, line 133: Will this work on non-AWS S3-compatible object stores or is it hard-wired to AWS only?

Ideally, it would gracefully degrade to non-AWS S3 stores that may not have a region.

bkmartinjr, line 133: might want to rename this `discover_s3_region_name`

bmccandless: I suspect the code as written would throw and exception if it could not discover a region name.  I'll change this function to return something else if a region is not found, and then handle errors in the eb/app

Issue #1152, opened 2020-02-13
Use categorical colors from anndata object 


ivirshup:
Fairly related to #446, but I didn't want to pollute that conversation too much.

In scanpy, we store colors for categories in `adata.uns["{category}_colors"]`, so that plots are consistent after subsetting. Could `cellxgene` use those colors for the categories if they're present?

To get a mapping of categories to colors, you would do:

```python
dict(zip(adata.obs[cat].cat.categories, adata.uns[f"{cat}_colors"]))
```

prete:
Unstructured Annotations (`adata.uns`) are note part of the [current schema](https://github.com/chanzuckerberg/cellxgene/blob/master/server/app/scanpy_engine/scanpy_engine.py#L194). Adding them could simplify the whole _custom-palette_ discussion, but the option to use them or not should be provided as an extra flag when launching cellxgene, for example if you want that behaviour to be the default one, add an option to override, something like `--ignore-uns-colors`.

Once `uns` are added to the world/schema, you'll only need to change [createColorsByCategoricalMetadata](https://github.com/chanzuckerberg/cellxgene/blob/ac13b31e13e03c00c531cae2541b177eb4e9c0b0/client/src/util/stateManager/colorHelpers.js#L39) scale to  

```javascript
const { categoriesColors } = world.schema.annotations.unsColorByName[accessor];
const scale =
    categoriesColors
    ?
    d3.scaleOrdinal(categoriesColors).domain([0, categories.length]);
    :
    d3.scaleSequential(interpolateRainbow).domain([0, categories.length]);
```

bkmartinjr:
We want the app to remain agnostic to the backend data storage system, so this will also require adding plumbing to handle non-anndata cases.

bkmartinjr:
@ivirshup - can you point me at the docs for this?  I'm wary of using these interfaces inappropriately, or having them only be supported by some of the past releases.  In particular, what is the full set of syntax supported by the uns[*_colors]?  My quick skim suggests that it can accommodate almost anything matplotlib supports?  

We definitely do not want to import matplotlib into cellxgene. Would it still be useful if we only supported the simple #rrggbb, and _maybe_ a best-effort attempt to map names between the matplotlib world and web?   "red" may not mean exactly the same thing in both, and some names may not translate at all.

prete:
https://icb-scanpy.readthedocs-hosted.com/en/stable/api/scanpy.pl.umap.html

```
[...]

Issue #1307, opened 2020-03-27
RFC: User-defined colors


mweiden:
# RFC: User-defined colors

_This is a proposed use case for public comment.  We are considering adding this in a future release.  If you have comments, please leave them, or just give us a thumbs up/down reaction.  Thanks!_

Authors: @mweiden, @bkmartinjr 
Document status: Open for Review
**Last date for comments**: EOD 2020-04-06

## Need

Researchers publishing datasets with cellxgene need the ability to ensure colors match their paper publication colors. Some example deployments include:

*   [Human Kidney Cell Atlas](https://www.kidneycellatlas.org/)
*   [Human Lung Cell Atlas](https://hlca.ds.czbiohub.org/)
*   [Tabula Muris Senis](https://tabula-muris-senis.ds.czbiohub.org/)

Note that these data publisher user stories are distinct from use cases of users consuming data through the cellxgene deployment. Further, we are not prioritizing custom colors for continuous variables at this time.

For more, see the user stories below.

**In-scope user stories:**
*   As a cellxgene user hosting datasets accompanying my paper, I want to _specify custom colors for data with a category label per dataset_, so that I can make my cellxgene deployments and/or images match my publication.
*   As a cellxgene user hosting datasets accompanying my paper, I want to _select from a set of colormaps for continuous variables per dataset_, so that I can make my cellxgene deployments and/or images match my publication.

**Not-in-scope user stories:**
*   As a cellxgene user using a cellxgene deployment to explore data, I want to _override colors in the UI with my own custom colors_, so that I can better interpret the data and generate images for new publications.
*   As a cellxgene user, I want to _specify custom colors for continuous variables_, so that I can make my cellxgene deployments and/or images match my publication.

Sources:
*   https://github.com/chanzuckerberg/cellxgene/issues/446 - How to specify a custom palette for a category?
*   https://github.com/chanzuckerberg/cellxgene/issues/1152 - Use categorical colors from anndata object

Definitions:
*   **Category-label pair**: Currently in annotations, you create a category, then labels for that category. So colors will be assigned per combination of 1) category 2) label for that category. In CS-speak this is a 2-tuple of (category_name, label_name). Example: `(organ, spleen)`
*   **.cxg**: `.cxg` is cellxgene’s native file format.

## Approach

The implementation would follow this user flow:
[...]

PR #1402, opened 2020-04-16
Add user-defined category-label colors

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1152

Designed as described in https://github.com/chanzuckerberg/cellxgene/issues/1307


mweiden:
Re. eslint errors: I don't think it makes sense to resolve those as part of this PR, since they all have to do with old code not in the diff. Probably better to separate concerns an ship those changes in a separate PR.

bkmartinjr:
> Re. eslint errors: I don't think it makes sense to resolve those as part of this PR, since they all have to do with old code not in the diff. Probably better to separate concerns an ship those changes in a separate PR.

We should turn off this test until such time as it is adding signal, rather than just noise!

client/src/actions/index.js
```diff
@@ -23,52 +23,46 @@ function obsAnnotationFetchAndLoad(dispatch, schema) {
...
-          dispatch({
-            type: "universe: column load success",
-            dim: "obsAnnotations",
-            dataframe: df,
-          });
-        });
-      })
-    )
+    columns
+      .filter(col => col.name !== index)
```
bkmartinjr, line 27: why is this being added back to the code?  It is done on line 20, so this is redundant.

client/src/actions/index.js
```diff
@@ -14,7 +14,7 @@ import { requestReembed, reembedResetWorldToUniverse } from "./reembed";
 return promise to fetch the OBS annotations we need to load.  Omit anything
 we don't need.
 */
-function obsAnnotationFetchAndLoad(dispatch, schema) {
+async function obsAnnotationFetchAndLoad(dispatch, schema) {
```
bkmartinjr, line 17: good catch!

client/src/actions/index.js
```diff
@@ -79,26 +73,35 @@ return promise fetching layout we need
...
-  ).then((dfs) => {
-    const df = Dataframe.Dataframe.empty().withColsFromAll(dfs);
+  ).then(dfs =>
     dispatch({
       type: "universe: column load success",
       dim: "obsLayout",
-      dataframe: df,
-    });
-  });
+      dataframe: Dataframe.Dataframe.empty().withColsFromAll(dfs)
```
bkmartinjr, line 89: in some places (like this) you are removing blocks and intermediate variables (like `df`), and in others not (eg, line 55/56).  Should probably shoot for a single style.

client/src/actions/index.js
```diff
@@ -79,26 +73,35 @@ return promise fetching layout we need
...
 
   const plimit = new PromiseLimit(5);
   return Promise.all(
     embNames.map((e) =>
       plimit.add(() => {
-        const url = `${baseURL}?layout-name=${encodeURIComponent(e)}`;
-        return doBinaryRequest(url).then((buffer) =>
-          Universe.matrixFBSToDataframe(buffer)
-        );
+        return fetchBinary(`layout/obs?layout-name=${encodeURIComponent(e)}`)
```
bkmartinjr, line 81: why the block - why not just an expression?

client/src/actions/index.js
```diff
@@ -23,52 +23,46 @@ function obsAnnotationFetchAndLoad(dispatch, schema) {
...
-          });
-        });
-      })
-    )
+    columns
+      .filter(col => col.name !== index)
+      .map(col =>
+        plimit.add(() => {
+          const path = `annotations/obs?annotation-name=${encodeURIComponent(col.name)}`;
+          return fetchBinary(path)
```
bkmartinjr, line 31: I would standardize on a single style - either embed the path inside the call to `fetchBinary()` and remove the block, or not.  See line 81 and 98 for the opposite style

client/src/actions/index.js
```diff
@@ -447,6 +441,14 @@ const saveObsAnnotations = () => async (dispatch, getState) => {
   }
 };
 
+function fetchJson(pathAndQuery) {
+  return doJsonRequest(`${globals.API.prefix}${globals.API.version}${pathAndQuery}`);
+}
+
+function fetchBinary(pathAndQuery) {
```
bkmartinjr, line 448: why not also use this inside _doRequestExpressionData?

client/src/globals.js
```diff
@@ -17,7 +17,8 @@ export const configDefaults = {
   parameters: {
     "max-category-items": 1000,
     "disable-diffexp": false,
-    "diffexp-may-be-slow": false
+    "diffexp-may-be-slow": false,
+    color_config: false
```
bkmartinjr, line 21: I can't find any reference to this -- is it dead code?

client/src/actions/index.js
mweiden: Good catch. Likely got added upon rebase.

client/src/util/stateManager/colorHelpers.js
```diff
@@ -15,9 +15,12 @@ create new colors state object.   Paramters:
     "color by continuous metadata", "color by categorical metadata"
   -
```
bkmartinjr, line 16: comment is out of date

client/src/util/stateManager/colorHelpers.js
```diff
@@ -15,9 +15,12 @@ create new colors state object.   Paramters:
...
 */
-export function createColors(world, colorMode = null, colorAccessor = null) {
+export function createColors(world, colorMode = null, colorAccessor = null, userColors = null) {
   switch (colorMode) {
     case "color by categorical metadata": {
+      if (userColors) {
```
bkmartinjr, line 21: If I understand the code right, it will re-parse user colors on every categorical color-by action the user takes.   Why is that done repetitively when colors are loaded once, at init time, and are thereafter unchanging?   Suggest at least memoizing the parsing result.

client/src/globals.js
mweiden: Yes, good catch. Removing this.

server/common/colors.py
```diff
@@ -0,0 +1,26 @@
...
+        return "#%02x%02x%02x" % tup
+    elif isinstance(unknown, str) and HEX_COLOR_FORMAT.match(unknown):
+        return unknown.lower()
+    else:
+        raise Exception("Unknown color format!")
+
+
+def anndata_colors_to_cxg_colors(data):
+    return dict(
+        (k[:-7], dict(zip(data.obs[k[:-7]].cat.categories, [convert_color_to_hex_format(c) for c in data.uns[k]])))
```
bkmartinjr, line 23: This code is incredibly opaque.   Any chance you could unroll it for clarity, or failing that, comment it?

For example, what is the input data format?

server/common/colors.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+HEX_COLOR_FORMAT = re.compile("^#[a-fA-F0-9]{6,6}$")
+
+
+def convert_color_to_hex_format(unknown):
+    if type(unknown) in (list, tuple):
```
bkmartinjr, line 7: please add a doc string describing what this function accepts.

also, this file needs a reference to the data format spec - it seems to be anndata specific, so what does that entail?

server/common/colors.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+HEX_COLOR_FORMAT = re.compile("^#[a-fA-F0-9]{6,6}$")
```
bkmartinjr, line 3: need to add unit tests for this

server/converters/cxgtool.py
```diff
@@ -409,6 +412,7 @@ def save_metadata(container, metadata):
     with tiledb.DenseArray(a_name, mode="w") as A:
         A.meta["cxg_version"] = CXG_VERSION
         A.meta["cxg_properties"] = json.dumps(metadata)
+        A.meta["colors"] = json.dumps(anndata_colors_to_cxg_colors(adata))
```
bkmartinjr, line 415: please prefix with `cxg_` as this is not a well-defined format  and is therefore unique to us.

server/converters/cxgtool.py
```diff
@@ -6,26 +6,27 @@
...
-    |                       dataframe columm, shape (n_obs,)
+    |                       dataframe column, shape (n_obs,)
     |-- var                 TileDB array containing gene (column) attributes, with one attribute per
     |                       dataframe column, shape (n_var,)
-    |-- X                   Main count matrix as a 2D TileDB array, single unnanmed numeric attribute
+    |-- X                   Main count matrix as a 2D TileDB array, single unnamed numeric attribute
     |-- emb                 TileDB group, storing optional embeddings (group may be empty)
     |   |-- <name1>         TileDB Array, single anon attribute, ND numeric array, shape (n_obs, N)
     |-- cxg_group_metadata  Empty array used only to stash metadata about the overall object.
+    |   |-- colors          CXG colors object as described in https://github.com/chanzuckerberg/cellxgene/issues/1307
```
bkmartinjr, line 16: please inline the definition - the github issue is not sufficient documentation (ie, just cut & paste the 5 lines that describe the dict layout)

bkmartinjr: Also, given that these are only category colors, and there is no clear way to determine that from the data structure described  in the github issue, I recommend you rename this to something like `cxg_category_colors`

server/converters/cxgtool.py
bkmartinjr, line 415: there is no error handling here, and the `anndata_colors_to_cxg_colors` function just throws undifferentiated Exception objects when it gets bad data.

It seems very likely that this particular interface is going to be very fragile as humans will be poking values into it regularly.   It needs more error handling/reporting, if not an option to ignore the values/errors and convert without colors.

server/common/colors.py
```diff
@@ -0,0 +1,26 @@
...
+            tup = tuple(int(ele * 255) for ele in unknown)
+        elif all(0 <= ele <= 255 and isinstance(ele, int) for ele in unknown):
+            tup = tuple(unknown)
+        else:
+            raise Exception("Unknown color format!")
+        return "#%02x%02x%02x" % tup
+    elif isinstance(unknown, str) and HEX_COLOR_FORMAT.match(unknown):
+        return unknown.lower()
+    else:
+        raise Exception("Unknown color format!")
```
bkmartinjr, line 18: this error handling makes it incredibly hard to do anything useful upstream.  Can you define a specific exception so the callers can differentiate error type?

server/converters/cxgtool.py
bkmartinjr, line 415: lastly, this should only be stored in the CXG if the cxgtool user has _asked_ for it to be stored (just like title and about are optional).

Please add a CLI param to turn this feature on/off.

server/data_common/data_adaptor.py
```diff
@@ -83,6 +83,10 @@ def query_var_array(self, term_var):
     def query_obs_array(self, term_var):
         pass
 
+    @abstractmethod
+    def colors(self):
```
bkmartinjr, line 87: I believe the style in this class is `get_foo`, not `foo`, so should be `get_colors()`

server/common/rest.py
```diff
@@ -222,6 +222,10 @@ def data_var_get(request, data_adaptor):
         return abort_and_log(HTTPStatus.BAD_REQUEST, str(e), include_exc_info=True)
 
 
+def colors_get(data_adaptor):
+    return make_response(jsonify(data_adaptor.colors()), HTTPStatus.OK)
```
bkmartinjr, line 226: error handling?  What if `colors()` fails?  Right now that will generate a 500 response, which seems brittle.

Suggest that it either return an empty color dictionary, OR return a useful HTTP error (eg, 404 or something like that)

server/common/colors.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def anndata_colors_to_cxg_colors(data):
```
bkmartinjr, line 21: needs docstring

What should be returned by this function if the underlying data is bad or can't be parsed?  Error/exception/empty dict/or?

bkmartinjr:
> Also I would be happy to hop on a call to learn how and how to add or modify the `cxg_group_metadata` section where this color info is intended to go (according to #1307 at least), and how to make a .cxg .

This is only used in the (automatically generated) CXG format.  You want to learn how to manipulate the AnnData and add this to the `uns` attribute.

client/src/util/stateManager/colorHelpers.js
mweiden: Good point. I'll put in a fix for this.

server/common/rest.py
mweiden: Both methods currently return an empty dictionary if there is no color information in the file. I will add a `try` here, though I'm not entirely sure the extent of exceptions that could be thrown here!

bkmartinjr: I am more worried about badly formatted color info.  Eg., user hand created, etc.

bkmartinjr: I can provide an example if you want one - there are lots of things that will potentially cause this to blow up.

server/common/colors.py
mweiden: Will unroll this.

server/common/colors.py
mweiden: Added 👍

server/common/colors.py
mweiden: As the code is written:if there is garbage in, you

mweiden:
@bkmartinjr I've covered all your comments so re-requested review. That said, codecov just reposted and caught me lowering our coverage, so I might go back and add a few more tests (on adaptors and in the API test). 🤦

bkmartinjr:
> @bkmartinjr I've covered all your comments so re-requested review.

You have lint (Python, not JS) which needs to be fixed.

server/common/rest.py
mweiden: > I am more worried about badly formatted color info

I _think_ f1b2f0b addresses the core concern here.

WRT other color failure modes, the only other case I can think of is where the user puts in a `adata.uns['{category_name}_colors']` key for a category that does not exist in `adata.obs`. The current code will simply leave this out of the cxg color data structure, which I think is better behavior than throwing an exception.

server/app/app.py
```diff
@@ -207,6 +207,12 @@ def get(self, data_adaptor):
         return common_rest.data_var_get(request, data_adaptor)
 
 
+class ColorsAPI(Resource):
+    @rest_get_data_adaptor
```
bkmartinjr, line 211: Please set a caching policy on this.

server/common/colors.py
```diff
@@ -0,0 +1,228 @@
...
+    violet="#ee82ee",
+    wheat="#f5deb3",
+    white="#ffffff",
+    whitesmoke="#f5f5f5",
+    yellow="#ffff00",
+    yellowgreen="#9acd32",
+)
+
+
+class ColorFormatException(Exception):
```
bkmartinjr, line 158: this should be in errors.py with all the other app-specific exceptions.

server/common/colors.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    if type(unknown) in (list, tuple):
+        if all(0.0 <= ele <= 1.0 for ele in unknown):
+            tup = tuple(int(ele * 255) for ele in unknown)
+        elif all(0 <= ele <= 255 and isinstance(ele, int) for ele in unknown):
+            tup = tuple(unknown)
+        else:
+            raise ColorFormatException("Unknown color format!")
+        return "#%02x%02x%02x" % tup
+    elif isinstance(unknown, str) and unknown in CSS4_NAMED_COLORS:
```
bkmartinjr, line 185: CSS4 colors (AFAIK) are case insensitive.  Do you intend for this to be case sensitive?

bkmartinjr: the linked matplotlib docs confirm this needs to be a  case-insensitive test

server/common/colors.py
```diff
@@ -0,0 +1,228 @@
...
+    The function accepts for the following formats:
+    - A CSS4 color name, as supported by matplotlib https://matplotlib.org/3.1.0/gallery/color/named_colors.html
+    - RGB tuple/list with values ranging from 0.0 to 1.0, as in [0.5, 0.75, 1.0]
+    - RFB tuple/list with values ranging from 0 to 255, as in [128, 192, 255]
+    - Hex triplet string, as in "#08c0ff"
+
+    :param unknown: color info of unknown format
+    :return: a hex triplet representing that color
+    """
+    if type(unknown) in (list, tuple):
```
bkmartinjr, line 177: these seem to allow tuples of any length, ie, (0.0) or (0.0,0.0,0.0,0.0).

server/test/test_colors.py
```diff
@@ -0,0 +1,23 @@
...
+import anndata
+from server.common.colors import convert_color_to_hex_format, convert_anndata_category_colors_to_cxg_category_colors
+from server.test import PROJECT_ROOT
+from server.test.test_datasets.fixtures import pbmc3k_colors
+
+
+class ColorsTest(unittest.TestCase):
+    """ Test color helper functions """
+
+    def test_convert_color_to_hex_format(self):
```
bkmartinjr, line 12: this should test bad format handling, ie,
* unknown names
* malformed hex
etc

and that the correct exceptions are thrown

server/test/test_colors.py
```diff
@@ -0,0 +1,23 @@
...
+        self.assertEqual(convert_color_to_hex_format("wheat"), "#f5deb3")
+        self.assertEqual(convert_color_to_hex_format((245, 222, 179)), "#f5deb3")
+        self.assertEqual(convert_color_to_hex_format([245, 222, 179]), "#f5deb3")
+        self.assertEqual(convert_color_to_hex_format("#f5deb3"), "#f5deb3")
+        self.assertEqual(
+            convert_color_to_hex_format([0.9607843137254902, 0.8705882352941177, 0.7019607843137254]), "#f5deb3"
+        )
+
+    def test_anndata_colors_to_cxg_colors(self):
```
bkmartinjr, line 21: I would also test that this performs as expected with malformed data (either by throwing the right exception, or by returning an empty dict - I am unclear _which_ it should do based on the function docstring)

server/common/colors.py
```diff
@@ -0,0 +1,228 @@
...
+        return "#%02x%02x%02x" % tup
+    elif isinstance(unknown, str) and unknown in CSS4_NAMED_COLORS:
+        return CSS4_NAMED_COLORS[unknown]
+    elif isinstance(unknown, str) and HEX_COLOR_FORMAT.match(unknown):
+        return unknown.lower()
+    else:
+        raise ColorFormatException("Unknown color format!")
+
+
+def convert_anndata_category_colors_to_cxg_category_colors(data):
```
bkmartinjr, line 193: what is the expected behavior on malformed data?  Looks it will throw a variety of exceptions, but not all are caught and processed the same upstream (as written, `convert_color_to_hex_format` can generate a TypeError or a ColorFormatException, maybe more, and only the latter is caught/handled in the REST route).

Worth adding this to the docstring, and then adding test cases for common malformed data.

server/converters/cxgtool.py
```diff
@@ -409,6 +441,17 @@ def save_metadata(container, metadata):
     with tiledb.DenseArray(a_name, mode="w") as A:
         A.meta["cxg_version"] = CXG_VERSION
         A.meta["cxg_properties"] = json.dumps(metadata)
+        if extract_colors:
+            try:
+                A.meta["cxg_category_colors"] = json.dumps(
+                    convert_anndata_category_colors_to_cxg_category_colors(adata)
```
bkmartinjr, line 447: this (currently) can also generate a TypeError which needs to be handled somewhere

bkmartinjr:
one more thought - what docs updates are needed?

server/common/colors.py
mweiden: Added documentation and wrapped `convert_color_to_hex_format` logic so that it always returns a `ColorFormatException`

mweiden:
@bkmartinjr good idea. Documentation updated 👍

docs/_site/index.html
```diff
@@ -16,10 +16,10 @@
 <meta property="og:url" content="http://localhost:4000/cellxgene/" />
 <meta property="og:site_name" content="cellxgene" />
 <script type="application/ld+json">
-{"description":"An interactive explorer for single-cell transcriptomics data","publisher":{"@type":"Organization","logo":{"@type":"ImageObject","url":"http://localhost:4000/cellxgene/cellxgene-logo.png"}},"@type":"WebSite","headline":"Index","url":"http://localhost:4000/cellxgene/","name":"cellxgene","@context":"http://schema.org"}</script>
+{"publisher":{"@type":"Organization","logo":{"@type":"ImageObject","url":"http://localhost:4000/cellxgene/cellxgene-logo.png"}},"description":"An interactive explorer for single-cell transcriptomics data","@type":"WebSite","url":"http://localhost:4000/cellxgene/","headline":"Index","name":"cellxgene","@context":"https://schema.org"}</script>
```
bkmartinjr, line 19: I realize this is not due to your recent changes, but these meta tags are all wrong (see the URLs).  We should either remove or fix.

If trivial, can you fix.  If not, file a bug.

docs/posts/prepare.md
```diff
@@ -5,20 +5,46 @@ description: Preparing your data
...
 
-If your data is in `h5ad` file (from the [`anndata`](https://anndata.readthedocs.io/en/latest/index.html) library) and meets the following requirements, you can go straight to `cellxgene launch`:  
+If your data is in `h5ad` file (from the [`anndata`](https://anndata.readthedocs.io/en/latest/index.html) library) and meets the following requirements, you can go straight to `cellxgene launch`:
 
-- Expression values (raw or normalized) in `anndata.X`  
+- Expression values (raw or normalized) in `anndata.X`
 - At least one embedding (e.g., tSNE, UMAP) in `anndata.obsm`, specified with the prefix `X_` (e.g., by default scanpy stores UMAP coordinates in `anndata.obsm['X_umap']`)
-- A unique identifier for every cell is available in an `anndata.obs` field (you can specify this with the `--obs-names` option)  
-- A unique identifier for every gene is available in an `anndata.var` field (you can specify which field to use with the `--var-names` option)  
+- A unique identifier for every cell is available in an `anndata.obs` field (you can specify this with the `--obs-names` option)
```
bkmartinjr, line 12: this is misleading.  What is required is a unique identifier for each cell, which by default will be pulled from the `obs` DataFrame index.  If the index is not unique or does not contain the cell ID, an alternative column can be specified with `--obs-names`.

docs/posts/prepare.md
```diff
@@ -5,20 +5,46 @@ description: Preparing your data
...
+- A unique identifier for every gene is available in an `anndata.var` field (you can specify which field to use with the `--var-names` option)
```
bkmartinjr, line 13: same issue as above

docs/posts/prepare.md
```diff
@@ -5,20 +5,46 @@ description: Preparing your data
...
 We hear you! We'd also love to be able to ingest these files directly. This isn't currently possible, but in the meantime, you can use [sceasy](https://bioconda.github.io/recipes/r-sceasy/README.html) ([docs](https://cellgeni.readthedocs.io/en/latest/visualisations.html)) to convert to `h5ad`. Seurat also has some [handy conversion tools](https://satijalab.org/seurat/v3.0/conversion_vignette.html) that you can try out.
 
-#### Can I use data hosted on the web somewhere?  
+#### Can I use data hosted on the web somewhere?
 Yes! You can launch from a URL instead of a filepath. The same data format requirements apply. Please see [here](launch) for more details.
 
-# Using `cellxgene prepare`  
+# Data format options
+
+#### Category colors
```
bkmartinjr, line 23: You might want to add a note about disabling this behavior, as the app will _always_ use colors if present, which is going to cause a change in default behavior for existing datasets.

server/converters/cxgtool.py
```diff
@@ -409,6 +442,17 @@ def save_metadata(container, metadata):
...
+                )
+            except (KeyError, ColorFormatException):
+                log(
+                    0,
+                    "Warning: failed to extract colors from h5ad file! "
+                    "Fix the h5ad file or rerun with --ignore-colors. See help for details.",
```
bkmartinjr, line 454: this error message seems out of date, as there is no `--ignore-colors` option that I can see.

server/converters/cxgtool.py
```diff
@@ -409,6 +442,17 @@ def save_metadata(container, metadata):
     with tiledb.DenseArray(a_name, mode="w") as A:
         A.meta["cxg_version"] = CXG_VERSION
         A.meta["cxg_properties"] = json.dumps(metadata)
+        if extract_colors:
+            try:
+                A.meta["cxg_category_colors"] = json.dumps(
+                    convert_anndata_category_colors_to_cxg_category_colors(adata)
```
bkmartinjr, line 448: IMHO, it would be cleaner if you did all this color extraction in write_cxg(), and just provided save_metadata() with a color dict to save. That fits better with the existing API signature.

It also lets you extract all colors _before_ you start creating tiledb objects, which keeps the file system clean in the case of an error.   Ie, convert_anndata...() should be run at the beginning of write_cxg() before the initial call to tiledb.group_create().

bkmartinjr:
@mweiden - the one follow-up thought I had.    We normally try to keep `launch` and cxgtool aligned so that the default launch behavior matches the default cxgtool behavior (ie, if you launch on the h5ad, or convert & view the cxg you see the same thing).

In this case, they have opposite defaults for colors:
* launch on h5ad will show colors, and has no way to disable that.
* convert h5ad does not convert colors by default, so will not show them

This is new - all other features are aligned in behavior so that the defaults match.   Thoughts?

docs/_site/index.html
mweiden: Well spotted. I think I've fixed this with a change to the `_config.yml` file.

docs/posts/prepare.md
mweiden: Err. The only reason this line was changed is was that my IDE cleaned up some whitespace. That said, I'll do my best to clarify the docs. Not sure that I have all the context here.

server/converters/cxgtool.py
mweiden: True. Will separate the concerns here.

server/converters/cxgtool.py
mweiden: True. I've thought about your comment below and decided to add this back and add a `--disable-user-colors` flag to `cellxgene launch`. This makes the behavior between cxgtool.py and the CLI consistent.

mweiden:
@bkmartinjr good points re. consistent CLI behavior. I've decided to make the feature on by default with consistent disable flags: both `cxgtool.py` and `cellxgene launch` now have a `--disable-user-colors` flag.

PR #1408, opened 2020-04-17
s3 listings were cached, preventing new datasets from being available

### LINKED ISSUE #1407: The s3 listings were cached, preventing new datasets from being made available

bmccandless:
Fixes #1407

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -18,4 +18,4 @@ PyYAML>=5.3
 scipy>=1.3.0
 requests>=2.22.0
 tiledb>=0.5.3
-s3fs>=0.4.0,!=0.4.1
+s3fs>=0.4.2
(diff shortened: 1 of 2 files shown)


server/requirements.txt
```diff
@@ -18,4 +18,4 @@ PyYAML>=5.3
 scipy>=1.3.0
 requests>=2.22.0
 tiledb>=0.5.3
-s3fs>=0.4.0,!=0.4.1
+s3fs>=0.4.2
```
bkmartinjr, line 21: Can you say more why you are requiring 0.4.2?   We worked previously with 0.4.0.

bmccandless: listings_expiry_time was introduced in 0.4.1.  Since that version is broken, I chose the next version, (which fixes the issue we hit in 0.4.1).  If you prefer I set it to:  s3fs==0.4.2

bkmartinjr: was there any support for cache management pre-0.4.1?  If not, I agree with your choice of >=0.4.2.  Just wanted to make sure we had a good reason!

bmccandless: No support as far as I can tell.    Yeah, I should have commented on the reason for the change.

PR #1412, opened 2020-04-17
temporarily disable eslint in github workflow

bkmartinjr:
Commented out the eslint steps until we burn down our lint debt.

### DIFF
--- a/.github/workflows/push_tests.yml
+++ b/.github/workflows/push_tests.yml
@@ -35,16 +35,16 @@ jobs:
       - name: Lint with flake8
         run: |
           make lint-server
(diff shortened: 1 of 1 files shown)


seve:
@mweiden @bkmartinjr This PR removed both the master and PR tests, I thought the objective was to remove solely the always failing tests on master.

bkmartinjr:
That would have been a better change - I just didn't think of it.  Would you like me to submit another PR?  Easily done!

seve:
@bkmartinjr Matt just found a bug in the job step, I'll submit a fix for the step and reactivate it in the same PR 👍

Issue #1022, opened 2019-11-06
App minimum width 1240px 


colinmegill:
When shrinking the window, the application shouldn't shrink horizontally past 1240px, a scrollbar should appear. This is complicated by our use, to this point, of fixed positioning.

seve:
How would the scroll bar interact with zoom/pan interaction mode?

colinmegill:
cc @liaprins-czi

colinmegill:
@seve hoping we can find a solution upstream of that... should behave normally and stop responding to horizontal browser resize at some point, but otherwise behave normally. Am I missing something significant?

liaprins-czi:
I agree^^^. The main vis area can still exhibit normal zooming behavior. Once the window size gets to 1240px, then the UI width stops shrinking. Instead it would stay the same width (1240px) and user would need to side-scroll to see the right edge.

If that doesn't make sense then I can show you or look for example sites that behave this way, just let me know, @seve ! Thanks!

colinmegill:
This may mean we are moving to `position: absolute` from fixed for the right and left edge of a div which takes up the whole viewport, rather than using the viewport itself. I'm not sure it's possible to accomplish what we want and keep `position: fixed`

PR #1424, opened 2020-04-24
Refactor CSS and React logic for layout

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1022

## Need
* Menubar should wrap inside middle pane instead of overlapping left
  sidebar when window is scrunched
* cellxgene should have a minimum width of 1240px

## Approach
1. Replace absolute positioning and dimension calculation with css grid
2. Use flexbox for wrapping menubar buttons
3. Add a minimum width of 1240px

## Beneifts
* Menubar doesn't puncture left sidebar
* Middle pane (graph) can calculate its own size
* Removing components calculating their size/position relative to
  each other increases modularity, decreases use of global variables
* Improved some scrollbar behavior
* Removed responsive reducer, propagating window size to components
  triggers unnecessary events and encourages breaking modularity; doing
  this made some components state agnostic
* Improved behavior during resizing the window (see below)

Resizing before:
scrunch_before

Resizing after:
scrunch_after


bkmartinjr:
@mweiden - I am going to defer to @colinmegill on this PR.

mweiden:
@seve Good thought. I'll try to add that later today.

mweiden:
@seve documentation added 👍

PR #1426, opened 2020-04-24
Revised terms and privacy consent dialog, analytics hooks

### LINKED ISSUE #1339: extend privacy/tos hook to include cookie ack

bkmartinjr:
revised UI to accommodate affirmative yes/no response and make the prompt text general purpose.
1. renders consent drawer if either ToS or Privacy URL are set and user has not previously given an answer.
2. answer stored in local storage so it is not transmitted to servers

Also added hooks for inline scripts (rather than just linking to external scripts), enabling the injection of bootstrap scripts for services like google analytics.  This includes two elements:
* a global callback, if present on `window`, will be invoked upon acceptance of cookies
* a global state variable will be set in localstorage indicating the decision (if made)

The primary gap in this implementation is that we lack a UI to change the answer once it is given (short of using the devtools console to clear the state manually).

Fixes #1339


bkmartinjr:
> Looks good to me. However when I ran this version with the standard ToS and Privacy links,
> the CZI logo in those links were gigantic. Let me know if you want help reproducing.

Can you elaborate?   Not sure I am understanding what you mean by "logo in those links".   This change shouldn't touch the tos/privacy files in any way.

bmccandless:
> > Looks good to me. However when I ran this version with the standard ToS and Privacy links,
> > the CZI logo in those links were gigantic. Let me know if you want help reproducing.
> 
> Can you elaborate? Not sure I am understanding what you mean by "logo in those links". This change shouldn't touch the tos/privacy files in any way.

Agreed that the change shouldn't have an impact.   It reproduces on my laptop, The CZI logo takes up more than the entire screen.  However, when I re-deployed just now on the dev server it looks normal.   So, in that case maybe it's just something in my local environment.  

The DEV server is up with this change, if you want to check it out there.

PR #1427, opened 2020-04-24
[feature] Prototype a Heroku deployment of hosted cellxgene

edulop91:
Some code to deploy cellxgene to heroku. There's probably room for improvement but things seem to work as far as I can tell. 

Here's a link to a dataset in the heroku version: https://czi-cellxgene-prod.herokuapp.com/d/pbmc3k.cxg/

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "cellxgene",
+  "version": "0.1.0",
+  "scripts": {
+    "postinstall": "npm build --prefix client"
+  }
+}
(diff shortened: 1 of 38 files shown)


package.json
```diff
@@ -0,0 +1,7 @@
+{
```
edulop91, line 1: heroku has strong opinions regarding the location of certain files. We write this top level package.json to leverage client/package.json

package.json
```diff
@@ -0,0 +1,7 @@
...
+  "name": "cellxgene",
+  "version": "0.1.0",
```
edulop91, line 3:
```suggestion
```

Issue #1251, opened 2020-03-19
ci/cid pipeline


parthshahva:
From slack:

Parth Shah  8:46 AM
So I don't like to deploy via terraform. Terraform will manage the infra but won't manage versions of the environment being deployed. Its not really built to handle this type of situation well
8:47
Best solution here is to have a CI/CD pipeline (running in gitlab or travis) that auto deploys master > dev on merge after running a suite of tests (edited) 
8:47
And have a one button deploy to Prod so that step (for now) requires human review

MDunitz:
Notes from call w/ @parthshahva 
Using gitlab actions
[ ] dev has hook into head of cellxgene master branch, CICD runs when master is updated
[ ] requirements file pins production version of cellxgene, CICD runs when that is manually updated
[ ] functional tests auto run after a deploy
[ ] cellxgene configuration updates run automatically

MDunitz:
how to differentiate between hosted versions and local versions?

MDunitz:
Colin is writing tests that will run against deployments

MDunitz:
Currently relying on team to not push to master on tuesday/wednesday (unless its needed in the release)

MDunitz:
We can tag master on Tuesday and then use that version when promoting to prod

MDunitz:
rerun tests on all config changes

PR #1430, opened 2020-04-27
Add Deployment hook

MDunitz:
## Need
- To automate deployment of cellxgene through the single cell infra repo master is updated
- Completes part of issue 1251 https://github.com/chanzuckerberg/cellxgene/issues/1251

## Approach
- On push to master post to the single cell repo dispatch endpoint, and run the deployment from there

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,7 @@ name: Deploy via single cell infra repo
 
 on:
   push:
-    branches: [master, dunitz-deploy-hook]
+    branches: master
 
 jobs:
   deploy:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,13 @@
+name: Deploy via single cell infra repo
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,13 @@
...
+  push:
+    branches: master
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: repository dispatch
+        run: |
+          curl -XPOST -u mdunitz:${{secrets.SCINFRA_TOKEN}} -H "Accept: application/vnd.github.everest-preview+json"  -H "Content-Type: application/json" https://api.github.com/repos/chanzuckerberg/single-cell-infra/dispatches --data '{"event_type": "cellxgene-hook"}'
```
bmccandless, line 13: Is there a way to do this without tying it to you specific user name?  Maybe we invent a user for this purpose?

MDunitz: I think creating a single-cell user makes sense. I'll talk to infra about doing that/create a ticket

PR #1434, opened 2020-04-28
Update the EB build process to package in scripts

bmccandless:
This is implemented specifically to handle the google analytics
script for hosted cellxgene

### DIFF
--- a/server/eb/Makefile
+++ b/server/eb/Makefile
@@ -27,6 +27,9 @@ build: clean
 	  mkdir -p artifact.dir/server/common/web/static/deploy; \
 	  cp -r static/* artifact.dir/server/common/web/static/deploy;  \
 	fi; \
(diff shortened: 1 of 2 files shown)


server/eb/Makefile
```diff
@@ -27,6 +27,9 @@ build: clean
 	  mkdir -p artifact.dir/server/common/web/static/deploy; \
 	  cp -r static/* artifact.dir/server/common/web/static/deploy;  \
 	fi; \
+	if [ -d scripts ] ; then \
```
bkmartinjr, line 30: very tiny point: should this directory be called `inline_scripts`?

bmccandless: I was trying to make it a place you could put scripts or inline_scripts. 
But I don't have a strong preference.

PR #1438, opened 2020-04-29
run prettier on client

seve:
This PR ran prettier(2.0.5) on the entirety of our client code


mweiden:
@seve what line width is this configured to use? We've used 120 for python/flake8.

seve:
@mweiden since we don't have a `.prettierrc` this is with the default prettier config: https://prettier.io/docs/en/options.html

mweiden:
@seve OK, so long as it's a conscious choice to do differently.

Issue #1440, opened 2020-04-30
cellxgene colors endpoint fails for .cxg v0


mweiden:
The `CxgAdaptor.get_colors` method currently assumes that the `.cxg` file has `cxg_group_metadata`. As a result, the `/api/v0.2/colors` endpoint always fails for `.cxg` v0.0 files.

PR #1441, opened 2020-04-30
Return empty colors for .cxg v0.0 files

mweiden:
Fixes https://github.com/chanzuckerberg/cellxgene/issues/1440

The CxgAdaptor.get_colors method currently assumes that the .cxg file has cxg_group_metadata. As a result, the /api/v0.2/colors endpoint always fails for .cxg v0.0 files.


mweiden:
@bkmartinjr aspirationally, yes

PR #1447, opened 2020-05-03
Add server plugin system

mweiden:
Plugins are optional modules loaded at runtime. Specification:
* Plugins are loaded from the server.plugins module (directory
  `server/plugins`)
* The import_plugins method is run as part of the initialization of the
  server module in `__init__.py`
* Plugins are copied into the ElasticBeanstalk artifact from the `server/eb` build configuration.

### DIFF
--- a/server/eb/Makefile
+++ b/server/eb/Makefile
@@ -41,6 +41,9 @@ build: clean
 	      cp -r customize/ebextensions/* artifact.dir/.ebextensions;  \
 	  fi; \
 	fi; \
+	if [ -d plugins ] ; then \
+	  cp -r plugins artifact.dir/server/;  \
+	fi; \
 	(cd artifact.dir; \
 	  cp -r server/common/web/static static; \
 	  zip -r ../artifact.zip . --exclude server/test/\* server/eb/\* ; ); \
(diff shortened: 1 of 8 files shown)


server/eb/Makefile
```diff
@@ -30,6 +30,9 @@ build: clean
 	if [ -d scripts ] ; then \
 	  cp -r scripts/* artifact.dir/server/common/web/templates;  \
 	fi; \
+	if [ -d plugins ] ; then \
```
bmccandless, line 33: There is now a customization directory that all these things are placed in.
You will need to make sure that this gets updated to  [ -d customize/plugins ]

server/eb/README.md
```diff
@@ -108,29 +108,35 @@ To include these scripts in the deployment, use the following steps:
    $ grep inline_scripts config.yaml
      inline_scripts : [ myscript.js ]
    ```
+
+1. Optionally, you can add plugins to the server python code. To include a plugin in the deployment use the following steps:
```
bmccandless, line 112: I have a PR out right now that adds custom requirements.   If your plugin pulls in new python modules, you can
place them in the custom requirements.txt file.   When both of these get to master, the docs should add a line for that.

PR #1449, opened 2020-05-04
uncomment popover conditional

seve:
One-liner to reinstate the diffexp warning popover


colinmegill:
@liaprins-czi thoughts on going down to one option?

bkmartinjr:
this is now obsolete, as it was based on old code.  Closing.

PR #1452, opened 2020-05-05
hosted, update order to look for config file.

bmccandless:
The app now uses a local config.yaml file bundled with the artifact
(if present), if it exists, then looks in the CXG_CONFIG_FILE
environment variable.  This is the reverse of previous behavior.
The purpose of this change is to move away from using the
config file on s3, since that could lead to problem where an older
version of the app uses a newer version of the config.

Also in this PR:
1. Changed documentation around dataroot, to describe the posibility of using lustre.
2. Added a few improvements around the secret manager region name.  If we use lustre for dataroot and a local config file, then we will no longer be able to
auto determine the region for the secret manager.  I plan to start using the
environment variable option for hosted cellxgene.

### DIFF
--- a/server/eb/README.md
+++ b/server/eb/README.md
@@ -32,7 +32,7 @@ There are many more options to these commands that may be important or necessary
    The following choices are known to work.
    
    * S3 Bucket. 
-   * Lustre filesystem
+   * POSIX filesystem (such as Lustre)
    * Lustre filesystem backed by S3
    
    S3 is convenient and the relatively inexpensive option.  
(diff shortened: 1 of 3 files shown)


server/eb/README.md
```diff
@@ -27,9 +27,22 @@ https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-install.html
 These steps are meant to serve as an example.  
 There are many more options to these commands that may be important or necessary for your environment.
 
-1. Create an S3 bucket
+1. Make your matrix files available to the EB servers.
 
-   Upload your matrix files to this bucket
+   The following choices are known to work.
```
bkmartinjr, line 32: we also know Posix file systems work.

PR #1453, opened 2020-05-05
Pin tiledb version 0.5.9

mweiden:
There is a regression in 0.6.0


mweiden:
@bkmartinjr bug here https://github.com/chanzuckerberg/cellxgene/issues/1454 I'll tag this PR there.

PR #1458, opened 2020-05-05
Reduce truncated category maxwidth

### LINKED ISSUE #1161: long category titles in LSB incorrectly layout
When a category name is long (>40 or 50 characters) it will cause the LSB to layout incorrectly.   The droplet icon is not visible, and the "expand category" chevron may wrap.

These categories are fairly common in datasets from the HCA.

Ideally we would truncate the name using the same method used elsewhere (ie, clip middle of string, replacing it with an ellipsis, and use a tooltip to show the entire name).

Screen shot:

colinmegill:
Fixes #1161

image

### DIFF
--- a/client/src/globals.js
+++ b/client/src/globals.js
@@ -73,7 +73,7 @@ export const leftSidebarSectionHeading = {
 export const leftSidebarSectionPadding = 10;
 export const categoryLabelDisplayStringLongLength = 27;
 export const categoryLabelDisplayStringShortLength = 11;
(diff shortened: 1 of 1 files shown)


bkmartinjr:
out of curiosity, is there a way to do this based on the rendered width (because the font is variable width)?     Seems like this solution will leave extra whitespace in some situations.

colinmegill:
Yes, this leaves extra whitespace in some situations. We'd need to do a bit more engineering to figure out the rendered width first, then choose the characters.

PR #1459, opened 2020-05-05
Create new label button overflow

### LINKED ISSUE #1131: Annotations label-creation modal enhancements
See design below:

annotations modal

colinmegill:
Fixes #1131

By removing `and`, and `currently`, we get the characters we need without losing meaning. If we don't like the comma, removing `currently` is enough.

image


bkmartinjr:
would ampersand be small enough? (ie, '&' instead of 'and')

colinmegill:
I like it!

PR #1460, opened 2020-05-05
Hooks for sentry integration

bkmartinjr:
Several hooks to enable more sophisticated integrations, motivated by Sentry.
* prod webpack config now wrapable by external build.
* script configuration now accepts more than a src filename.  It will also inject integrity and crossorigin attributes.
* top-level index template file not minified or mutated during webpack'ing.  This lets jinja template fields flow through without modification


client/configuration/webpack/cspHashPlugin.js
```diff
@@ -32,17 +32,15 @@ class CspHashPlugin {
...
-            $(elmt).removeAttr("no-csp-hash");
-            foundOne = true;
-          });
-          $("style[no-csp-hash]").each((i, elmt) => {
-            $(elmt).removeAttr("no-csp-hash");
-            foundOne = true;
-          });
-          if (foundOne) data.html = $.html();
+          // remove no-csp-hash attributes.  Cheerio does not parse Jinga templates
+          // correctly, so we brute force this wiht a regular expression.
```
bkmartinjr, line 36: spelling error in comment.

server/eb/app.py
```diff
@@ -60,6 +60,7 @@ def _before_adding_routes(app, app_config):
         script_hashes, style_hashes = WSGIServer.get_csp_hashes(app, app_config)
         csp = {
             "default-src": ["'self'"],
+            "connect-src": ["'self'"],
```
bkmartinjr, line 63: why is this needed?  My understanding is that if is unspecified, it falls back to the `default-src` definition, which is identical.   @nacin - do you know?

mweiden: @bkmartinjr when the client code pings our sentry hostname, the console reported it looking in `connect-src` first. I added this so as to be as specific as possible about the scope of that hostname. Happy to change it, if there is a better solution.

PR #1463, opened 2020-05-06
remove all linting errors on client/src

seve:
This PR goes through src/ and removes all eslint errors as well as any hacky eslint disables.

~I've tried to keep this as unopinionated as possible, but please note that in favor of maintaining the js convention of camelcase, I have sanitized the config object obtained from the server API and converted it from snake_case to camelCase. Let me know if you agree or disagree with this in the comments :)~(Camelcasing sanitation was removed in favor of subscription snake case)

I've also begun adding comments to justify eslint-disables, I'd recommend continuing this practice, especially with [descriptions coming in eslint 7.0](https://eslint.org/blog/2020/02/whats-coming-in-eslint-7.0.0#descriptions-in-directive-comments)

Note this does not include testing code


bkmartinjr:
@seve - can you expand on your decision to camelcase the json data structure returned in the REST API?   I'm not keen to have to mentally map between the two worlds, as it seems error prone and somewhat arbitrary (after all, this is a style issue, which should reduce cognitive load, not increase it).

I'd much rather say that the REST API "is what it is", and doesn't get remapped in the client just for stylistic reasons.   Right now, it is obeying Python conventions, not JS conventions.  We could switch, end-to-end to use JS, and use camelcase in the back-end, but that seems arbitrary.

To my mind, working on in both worlds, the worst case is the remapping.

client/src/components/termsPrompt/index.js
```diff
@@ -26,13 +25,13 @@ function storageSet(key, value) {
   try {
     window.localStorage.setItem(key, value);
   } catch {
-    return;
+    // continue
```
bkmartinjr, line 28: this seems error prone, if we were to ever add code after the try/catch.   Continue and return have different semantics.

client/src/components/graph/overlays/centroidLabels.js
```diff
@@ -58,13 +56,16 @@ class CentroidLabels extends PureComponent {
...
         <g
+          // label is unique so disabling eslint rule
           // eslint-disable-next-line react/no-array-index-key
           key={label}
           className="centroid-label"
           transform={`translate(${coords[0]}, ${coords[1]})`}
           data-testclass="centroid-label"
           data-testid={`${label}-centroid-label`}
         >
+          {/* The mouse actions for centroid labels do not have a screen reader alternative */}
```
bkmartinjr, line 67: why put comments in expressions - doesn't this cause runtime overhead?

seve: Since the comment is a child it must be wrapped in brackets or it will be converted to a react text node

client/src/components/termsPrompt/index.js
seve: Looks like there were two steps to get to this result.  Running auto fix on the file led to the removal of `return` as it was unnecessary but the introduction of an empty block statement.   `no-empty` has the condition:  

> This rule ignores block statements which contain a comment (for example, in an empty catch or finally block of a try statement to indicate that execution should continue regardless of errors).

So the comment was me getting rid of the new eslint error, didn't consider the semantic meaning behind continue when making the comment.  Would you suggest a new comment such as: 

> // continue regardless of error

bkmartinjr: thinking about this more, I don't have a great suggestion.  And the code is simple enough to ignore this.  So lets leave as is, and I"ll resolve this comment.

seve:
> @seve - can you expand on your decision to camelcase the json data structure returned in the REST API? I'm not keen to have to mentally map between the two worlds, as it seems error prone and somewhat arbitrary (after all, this is a style issue, which should reduce cognitive load, not increase it).

My decision is pretty weakly backed and I've actually start leaning towards keeping the key-names as is an accessing via subscripting. I'll revert the commits :)

seve:
@bkmartinjr removed the camelcase cleansing in favor of subscripts

client/src/components/scatterplot/scatterplot.js
```diff
@@ -316,6 +307,15 @@ class Scatterplot extends React.PureComponent {
     };
   }
 
+  handleResize = () => {
```
seve, line 310: @mweiden I'm not sure if this function is getting the expected functionality

bkmartinjr:
> @bkmartinjr removed the camelcase cleansing in favor of subscripts

PR description still says you are camelCasing everything - may want to update it.

client/src/components/categorical/category/index.js
```diff
@@ -198,6 +202,15 @@ class Category extends React.Component {
       globals.categoryDisplayStringMaxLength
     );
 
+    const handleCategoryClick = () => {
```
bkmartinjr, line 205: why take the performance hit of doing this on each render?  You could make it an instance value and have same effect, with less overhead.

seve: The function was originally an arrow function in render/props and I refactored without thinking of performance enhancements.  I'll add it as a class property 👍

client/src/util/stateManager/controlsHelpers.js
```diff
@@ -13,8 +13,8 @@ import {
 
 export function maxCategoryItems(config) {
   return (
-    config.parameters?.["max-category-items"] ??
-    globals.configDefaults.parameters["max-category-items"]
+    config.parameters?.maxCategoryItems ??
```
bkmartinjr, line 16: I thought the camelcase change was reverted?

client/src/components/graph/graph.js
```diff
@@ -261,7 +264,10 @@ class Graph extends React.Component {
       const { drawPoints, pointBuffer, colorBuffer, flagBuffer } = this.state;
 
       if (hasResized) {
-        projectionTF = createProjectionTF(this.reglCanvas.width, this.reglCanvas.height);
+        projectionTF = createProjectionTF(
```
bkmartinjr, line 267: What line length are we splitting at?  Originally, I thought we used 120 characters as we all have big screens, but this looks like it is splitting on a much shorter length.

client/src/components/categorical/category/index.js
bkmartinjr: cool - it will need to be an assigned instance property, so that it gets the bind for free.

client/src/components/graph/graph.js
seve: [Prettier defaults to a length of 80](https://prettier.io/docs/en/options.html#print-width)

client/src/components/categorical/category/index.js
seve: My terminology knowledge is failing me. Could you elaborate on "assigned instance property"? I typically get my bind by defining the prop as `prop={()=>functionName()}`

client/src/util/stateManager/controlsHelpers.js
seve: shoot, I must have forgotten to revert a commit. I'm going to grep the codebase to comb for anything left behind

client/src/components/graph/graph.js
bkmartinjr: ah yes, i am confusing it with our Python standards.   You might poll the team on what line length they prefer, but for now, leaving it with the default is sensible.

client/src/components/categorical/category/index.js
bkmartinjr: it should be unique to each instance, rather than being in the prototype.

In other words, create it thusly:
```
class Foo {
   // ...
   handleCategoryClick = (e) => { /* ... */ };
}
```

Rather than the more typical:
```
class Foo {
  // ...
  handleCategoryClick(e) { /* ... */ }
}
``

The former will have `this` bound to the instance, the latter will not.

client/src/components/scatterplot/scatterplot.js
seve: Filed #1482

PR #1467, opened 2020-05-11
diffexp, precompute mean/var for selections based on labels

### LINKED ISSUE #1418: diffexp timeout/504 on DEV cluster
Test case:
1. Load Kidney Cell Atlas dataset
2. Select louvain label "6" and make that cell set 1
3. Select louvain label "7" and make that cell set 2
4. run diffexp and it times out (> 30 seconds) resulting in a 504 from load balancer

The total number of cells is ~24K, which is well under our configuration limit of 50K

bmccandless:
This can help fix #1418

### DIFF
--- a/server/converters/diffexp_precompute.py
+++ b/server/converters/diffexp_precompute.py
@@ -1,0 +1,144 @@
+"""This program precomputes data to speed up diffexp."""
+
+import sys
+import os
+import tiledb
+import argparse
+import numpy as np
+import shutil
+from server.common.app_config import AppConfig
+from server.data_common.matrix_loader import MatrixDataLoader
+from server.compute.diffexp_generic import compute_indices_hash, compute_indices_hash_version
+from server.compute.diffexp_cxg import compute_mean_var
+from server.data_cxg.cxg_adaptor import CxgAdaptor
+from server.common.errors import DatasetAccessError, ComputeError
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("dataset", help="input CXG dataset")
+    parser.add_argument("--max_groups", "--max", type=int, default=100)
+    parser.add_argument("--min_groups", "--min", type=int, default=2)
+    parser.add_argument("--max_rows", "--maxr", type=int, default=1000000)
+    parser.add_argument("--overwrite", action="store_true")
+    args = parser.parse_args()
+    dataset = args.dataset
+
+    try:
+        app_config = AppConfig()
+        app_config.single_dataset__datapath = dataset
+        app_config.server__verbose = True
+        app_config.diffexp__alg_cxg__target_workunit = 8000000
+        app_config.complete_config()
+
+        loader = MatrixDataLoader(dataset)
+        adaptor = loader.open(app_config)
+    except DatasetAccessError as e:
+        print(f"Unable to open {dataset}: {str(e)}")
+        sys.exit(1)
+
+    if not isinstance(adaptor, CxgAdaptor):
+        print(f"Expected {dataset} to be a valid CxgAdaptor")
+        sys.exit(1)
+
+    if not adaptor.data_locator.islocal():
+        print(f"Expected {dataset} to be local")
+        sys.exit(1)
+
+    diffexp_group = os.path.join(dataset, "diffexp")
+    if os.path.exists(diffexp_group):
+        print(f"{diffexp_group} exists")
+        if args.overwrite:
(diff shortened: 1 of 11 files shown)


bkmartinjr:
need to fix lint that causes tests to fail.

server/converters/diffexp_precompute.py
```diff
@@ -0,0 +1,144 @@
...
+    parser.add_argument("dataset", help="input CXG dataset")
+    parser.add_argument("--max_groups", "--max", type=int, default=100)
+    parser.add_argument("--min_groups", "--min", type=int, default=2)
+    parser.add_argument("--max_rows", "--maxr", type=int, default=1000000)
+    parser.add_argument("--overwrite", action="store_true")
+    args = parser.parse_args()
+    dataset = args.dataset
+
+    try:
+        app_config = AppConfig()
```
bkmartinjr, line 28: converters (ie, cxgtool) is currently stand-alone. this forces it to be run as part of the cellxgene package.  is that intentional?

server/converters/diffexp_precompute.py
```diff
@@ -0,0 +1,144 @@
...
+
+        print(f"Queueing '{attr}': num groups {ngroups}")
+        num_precompute += ngroups
+        attr_precompute.append(attr)
+
+    hashkey = np.empty((num_precompute,), dtype=object)
+    metakey = np.empty((num_precompute, 2), dtype=object)
+    mean_var = np.zeros((2, num_precompute, shape[1]), dtype=np.float64)
+
+    print()
```
bkmartinjr, line 83: print?

server/converters/diffexp_precompute.py
```diff
@@ -0,0 +1,144 @@
...
+from server.data_cxg.cxg_adaptor import CxgAdaptor
+from server.common.errors import DatasetAccessError, ComputeError
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("dataset", help="input CXG dataset")
+    parser.add_argument("--max_groups", "--max", type=int, default=100)
+    parser.add_argument("--min_groups", "--min", type=int, default=2)
+    parser.add_argument("--max_rows", "--maxr", type=int, default=1000000)
```
bkmartinjr, line 22: why have a max?  Seems like the bigger the row set, the more we need the precompute?

server/converters/diffexp_precompute.py
```diff
@@ -0,0 +1,144 @@
...
+                print(f" '{attr}'='{str(val)}': num indices = {grouplen} > {max_rows}: skipping")
+                continue
+            else:
+                print(f" '{attr}'='{str(val)}': num indices = {grouplen}")
+            try:
+                tmean, tvariance = compute_mean_var(adaptor, groupindices)
+            except ComputeError as e:
+                print(f"   error: {str(e)}")
+                continue
+            mean_var[0][pindex] = tmean
```
bkmartinjr, line 122: what is the benefit of strong mean and var in a single (2, N) array, rather than creating a two attribute array?  The latter has the benefit of more easily being expanded when you decide you want to compute additional summary metrics for these row groups - you just add another tiledb attribute to the array.

it also makes the code a lot easier to understand.

server/converters/diffexp_precompute.py
```diff
@@ -0,0 +1,144 @@
...
+            mean_var[1][pindex] = tvariance
+            hashkey[pindex] = compute_indices_hash(groupindices, hash_version)
+            metakey[pindex][0] = attr
```
bkmartinjr, line 125: why store `attr` and `str(val)` in metadata, rather than in additional columns of the same array?

bkmartinjr, line 125: why not just store these as additional attribute (aka columns) in the same array?

server/data_cxg/cxg_adaptor.py
```diff
@@ -133,6 +138,15 @@ def isvalid(url):
             return False
         return True
 
+    def diffexp_cache(self, hashval):
+        """Find the hashval in the cache.  For now, the cache is only populated from the precomputed values"""
+        if not self.diffexp_has_cache:
+            return None
+        if self.diffexp_hashkey is None:
+            hashkey = self.open_array("diffexp/hashkey")
+            self.diffexp_hashkey = {h: i for i, h in enumerate(hashkey[:])}
```
bkmartinjr, line 147: why pre-load (memory cache) this index, but not the mean & var?  They are not going to take much more space than the dict itself....

PR #1469, opened 2020-05-12
Fix smoke tests to work with remote deployments of cellxgene

mweiden:
## Needs

* The smoke test needs to accept any terms of service (ToS) shown in the browser before running the test suite.
* The version of Chromium browser used by puppeteer does not seem to support all CSP directives: `Unrecognized Content-Security-Policy directive 'require-trusted-types-for'.`

## Approach

* Add a test at the beginning of the suite that accepts the ToS.
* Ignore the `require-trusted-types-for` directive console error, for now.

### DIFF
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -29,6 +29,17 @@ describe("did launch", () => {
     );
     expect(element).toBe(data.title);
   });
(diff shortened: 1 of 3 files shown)


mweiden:
@bkmartinjr I have a separate PR out to fix the lint errors (python).

PR #1485, opened 2020-05-20
Upgrade to tiledb 2.0

bmccandless:


server/requirements.txt
```diff
@@ -17,6 +17,6 @@ pandas>=0.24.2
 PyYAML>=5.3
 scipy>=1.3.0
 requests>=2.22.0
-tiledb==0.5.9
+tiledb==0.6.1
```
bkmartinjr, line 20: do we have an explicit dependency on 0.6.1?  Generally we do not force update package versions unless we have an explicit reason (this lets people stay on old versions if they have a reason to do so).  If you are avoiding a release due to a package bug, you can exclude a version.

Also, we normally use >= rather than == for the pip module.

So, I'd suggest either:
* if we need a feature/function in 0.6.1, then `tiledb>=0.6.1`
* if we don't, then perhaps `tiledb>=0.5.9` or `tiledb>=0.5.9,!=0.6.0` depending on the need

bmccandless: good point.   I think for now the second option is best.

PR #1492, opened 2020-05-20
Add help/enforcement with ensuring lint pass

seve:
This PR adds:
* Husky + lint-staged - Prevents lint fails from being committed
* ESLint-prettier - new ESLint rules aligned with our prettier styling
* ESLint tests on PR - Test will fail if there are any ESLint errors

Also removes ESLint error introduced in #1473

### DIFF
--- a/.github/workflows/push_tests.yml
+++ b/.github/workflows/push_tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: master
   pull_request:
+    branches: *
 
 env:
   JEST_ENV: prod
@@ -37,14 +38,9 @@ jobs:
           make lint-server
       - name: Lint src with eslint
         working-directory: ./client
-        if: github.event_name	!= 'pull_request'
         run: |
           make lint
-      # - name: Lint diff with eslint
-      #   working-directory: ./client
-      #   if: github.event_name	== 'pull_request'
-      #   run: |
-      #     make lint-diff
+
 
   unit-test:
     runs-on: ubuntu-latest
--- a/client/configuration/lint-staged/lint-staged.config.js
+++ b/client/configuration/lint-staged/lint-staged.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  "/src/**/*.js": "eslint --fix ./src/",
+};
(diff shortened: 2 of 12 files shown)


client/configuration/lint-staged/lint-staged.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  "./src/**/*.js": "eslint --fix ./src/",
```
tihuan, line 2: Thanks for adding this, @seve 🎉 👏 !!

Do we wanna run `prettier` on staged files too?

And in Meta we added `git add` to add the fixed files into the commit

e.g., 

```ts
'*.js': ['prettier --write', 'git add'],
```

Do we want to do it here too?

Thank you!

https://github.com/chanzuckerberg/khepri/blob/develop/lint-staged.config.js

seve: We currently don't have prettier added as a dev dep, however since we are using the prettier ESLint plugin, `--fix` will auto-format for us similarly to how prettier --write' would. The question is: do we have to git add after running `eslint --fix`.
> The rule is autofixable -- if you run eslint with the --fix flag, your code will be formatted according to prettier style. [(link)](https://github.com/prettier/eslint-plugin-prettier)

Thank's for the feedback!  @tihuan

seve: [It seems like it automatically adds it to the commit.](https://github.com/okonet/lint-staged#automatically-fix-code-style-with---fix-and-add-to-commit)

tihuan: Ah nice!! Thanks for the link 😄  Glad to know `git add` is no longer needed. Will update on the Meta too!

.github/workflows/push_tests.yml
```diff
@@ -37,14 +38,9 @@ jobs:
...
       - name: Lint src with eslint
         working-directory: ./client
-        if: github.event_name	!= 'pull_request'
         run: |
           make lint
-      # - name: Lint diff with eslint
-      #   working-directory: ./client
-      #   if: github.event_name	== 'pull_request'
-      #   run: |
-      #     make lint-diff
```
mweiden, line 47: If the lint tests are now run on all source, we can remove the `lint-diff` make commands, true?

seve: True, I'll add that in. Thanks!

PR #1496, opened 2020-05-26
Support for sparse tiledb arrays for the X matrix

bmccandless:
1. cxgtool can now output sparse matrices
2. cxg_adaptor and diffexp_cxg updated to handle sparse matrices
3. added a test in test_diffexp to test sparse diffexp and get_X_array

### DIFF
--- a/server/compute/diffexp_cxg.py
+++ b/server/compute/diffexp_cxg.py
@@ -3,6 +3,8 @@ import numpy as np
 from server.compute.diffexp_generic import diffexp_ttest_from_mean_var, mean_var_n
 from server.data_cxg.cxg_util import pack_selector_from_indices
 from server.common.errors import ComputeError
+from numba import jit
+
 
 """
 See the comments in diffexp_generic for a description of this algorithm
@@ -33,26 +35,33 @@ def get_thread_executor():
 
 
 def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01):
+
+    matrix = adaptor.open_array("X")
     row_selector_A = np.where(maskA)[0]
     row_selector_B = np.where(maskB)[0]
     nA = len(row_selector_A)
     nB = len(row_selector_B)
-    matrix = adaptor.open_array("X")
 
     dtype = matrix.dtype
     cols = matrix.shape[1]
     tile_extent = [dim.tile for dim in matrix.schema.domain]
 
-    # The rows from both row_selector_A and row_selector_B are gathered at the
-    # same time, then the mean and variance are computed by subsetting on that
-    # combined submatrix.  Combining the gather reduces number of requests/bandwidth
-    # to the data source.
-    row_selector_AB = np.union1d(row_selector_A, row_selector_B)
-    row_selector_A_in_AB = np.in1d(row_selector_AB, row_selector_A, assume_unique=True)
-    row_selector_B_in_AB = np.in1d(row_selector_AB, row_selector_B, assume_unique=True)
-    row_selector_AB = pack_selector_from_indices(row_selector_AB)
+    is_sparse = matrix.schema.sparse
 
-    # because all IO is done per-tile, and we are always dense and col-major,
+    if is_sparse:
+        row_selector_A = pack_selector_from_indices(row_selector_A)
+        row_selector_B = pack_selector_from_indices(row_selector_B)
+    else:
+        # The rows from both row_selector_A and row_selector_B are gathered at the
+        # same time, then the mean and variance are computed by subsetting on that
+        # combined submatrix.  Combining the gather reduces number of requests/bandwidth
+        # to the data source.
+        row_selector_AB = np.union1d(row_selector_A, row_selector_B)
+        row_selector_A_in_AB = np.in1d(row_selector_AB, row_selector_A, assume_unique=True)
+        row_selector_B_in_AB = np.in1d(row_selector_AB, row_selector_B, assume_unique=True)
+        row_selector_AB = pack_selector_from_indices(row_selector_AB)
+
+    # because all IO is done per-tile, and we are always col-major,
     # use the tile column size as the unit of partition.  Possibly access
     # more than one column tile at a time based on the target_workunit.
     # Revisit partitioning if we change the X layout, or start using a non-local execution environment
@@ -62,6 +71,7 @@ def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01):
     # the target_workunit.  A potential improvement would be to partition by both columns and rows.
     # However partitioning the rows is slightly more complex due to the arbitrary distribution
     # of row selections that are passed into this algorithm.
+
     cells_per_coltile = (nA + nB) * tile_extent[1]
     cols_per_partition = max(1, int(target_workunit / cells_per_coltile)) * tile_extent[1]
     col_partitions = [(c, min(c + cols_per_partition, cols)) for c in range(0, cols, cols_per_partition)]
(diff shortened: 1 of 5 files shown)


server/converters/cxgtool.py
```diff
@@ -398,44 +408,87 @@ def save_embeddings(container, adata, ctx):
...
-def create_X(X_name, shape):
+def create_X(X_name, shape, is_sparse):
     """
-    Dense, always.  Future task: explore if sparse encoding is worth the trouble
-    below a sparsity threshold.
-
     The X matrix is access in both row and column oriented patterns, depending on the
     particular operation.  Because of the data type, default compression works best.
-    The tile size (50, 100) and global layout (row/col) was choosen empirically, by benchmarking
+    The tile size (50, 100) and global layout (row/col) was chosen empirically, by benchmarking
```
bkmartinjr, line 415: please update the comment to include information about the sparse tile sizes (I assume they were also empirically discovered based on perf?)

server/converters/cxgtool.py
```diff
@@ -398,44 +408,87 @@ def save_embeddings(container, adata, ctx):
...
 
 
-def save_X(container, adata, ctx):
+def count_nonzero_X(adata):
+    shape = adata.X.shape
+    stride = min(int(np.power(10, np.around(np.log10(1e9 / shape[1])))), 10_000)
+    nnz = 0
+    for row in range(0, shape[0], stride):
+        lim = min(row + stride, shape[0])
+        a = adata.X[row:lim, :]
```
bkmartinjr, line 445: can you confirm this was tested with the `--backed` flag?  I think it should be OK, but just want to confirm that this test occurred.  We don't use `--backed` much, but for datasets that are larger than main memory, it is important to retain the functionality.

server/converters/cxgtool.py
```diff
@@ -398,44 +408,87 @@ def save_embeddings(container, adata, ctx):
...
-    create_X(X_name, shape)
+
+    if sparse_threshold == 100:
+        is_sparse = True
+    elif sparse_threshold == 0:
+        is_sparse = False
+    else:
+        nnz = count_nonzero_X(adata)
+        percent = 100.0 * nnz / (shape[0] * shape[1])
+        log(1, "shape:", str(shape), "non-zeros:", nnz, "percent: %6.2f" % percent)
```
bkmartinjr, line 464: would be useful to log the is_sparse value, just to let the user have better logs

server/compute/diffexp_cxg.py
```diff
@@ -3,6 +3,8 @@
 from server.compute.diffexp_generic import diffexp_ttest_from_mean_var, mean_var_n
 from server.data_cxg.cxg_util import pack_selector_from_indices
 from server.common.errors import ComputeError
+from numba import jit
```
bkmartinjr, line 6: This is a big change in dependencies, as it will flow downstream into all users who pip install.  Has there been any research or validation that the potential for downstream breakage is acceptable?   I like the idea of using it, just noting that it is a very large change hidden behind a couple of lines of python :)

server/data_cxg/cxg_adaptor.py
```diff
@@ -200,15 +204,44 @@ def get_colors(self):
         meta = self.open_array("cxg_group_metadata").meta
         return json.loads(meta["cxg_category_colors"]) if "cxg_category_colors" in meta else dict()
 
+    def __remap_indices(self, coord_range, coord_mask, coord_data):
```
bkmartinjr, line 207: to my eye, this is challenging to understand/validate, and uses a lot of memory to avoid a loop where you just iteratively create coordindices.  

For the get_X_array(), did you consider using something simpler like the scipy sparse implementations, which are efficient and well tested?  

For example, if you create a [coo_matrix](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.html#scipy.sparse.coo_matrix) and then call [toarray()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.spmatrix.toarray.html#scipy.sparse.spmatrix.toarray), you could dispense with this entire function and simplify get_X_array().

server/converters/cxgtool.py
```diff
@@ -114,6 +114,13 @@ def main():
         help="URL providing more information about the dataset (hint: must be a fully specified absolute URL).",
     )
     parser.add_argument("--out", "--output", "-o", help="output CXG file name")
+    parser.add_argument(
```
bkmartinjr, line 117: one idea triggered by this work -- it may be useful to add a stand-alone validation tool that compares a CXG to an H5AD and complains if they don't match.   It could be used as part of creating a new CXG to confirm correctness, and as a stand-alone option used to validate the integrity of pre-existing CXGs (eg, does the version on S3 match the H5AD) and would make our unit tests more robust.

server/converters/cxgtool.py
bmccandless: They were chosen based on a relatively small amount of benchmarking.   I intend to fine too this as a follow-on update.  I'll update the comment though

server/compute/diffexp_cxg.py
bmccandless: From http://numba.pydata.org/
From there there is is a link to more info on portability.
Let me know if you think we should do something else here.

"""
Ship high performance Python applications without the headache of binary compilation and packaging. Your source code remains pure Python while Numba handles the compilation at runtime. We test Numba continuously in more than 200 different platform configurations.

Numba supports Intel and AMD x86, POWER8/9, and ARM CPUs, NVIDIA and AMD GPUs, Python 2.7 and 3.4-3.7, as well as Windows/macOS/Linux. Precompiled Numba binaries for most systems are available as conda packages and pip-installable wheels.
"""

server/data_cxg/cxg_adaptor.py
bmccandless: I did think about using coo_matrix, but it doesn't quite do what we want.
We want to compress out rows and columns that were not selected.  The toarray function returns a dense array with a shape like (max row selected, max col selected).  coo_matrix is also not scriptable, so you can't slice into it either.  I could use coo_matrix, then convert to a 2d numpy array, then slice out of that.  The code would be simpler, but it would create a large dense matrix as an intermediate step.  Please let me know if you see a path to use coo matrix that doesn't blow up memory.

This approach does not use a lot of memory.   Assuming that we want to return a dense submatrix (what the frontend currently expects), the amount of temporary memory used here is less than the memory of the output array.  memory is O(nrows) or O(ncols).  The amount of compute is just O(nnz).   I did write this to avoid loop, and so maybe it's not so obvious from reading the code.   But the basic operation isn't too complicated.

Say there is a matrix like this:

```
0  1  2  0  3  0
4  0  5  0  6  0  
0  0  7  8  0  9
```

And we wanted to select all the rows,  and cols = 0, 2, 5
The output matrix should be:

```
0  2  0
4  5  0
0  7  9
```

From tiledb we get:
obs = [0, 1, 1, 2, 2]
var =  [2, 0, 2, 2, 5]
data = [2,4,5,7,9]

So the challenge is how to transform the three arrays (obs, var, data) to the 2d output matrix.

The first step is to create a mapping from X var coords to the output columns:
e.g. map 0->0, 2->1, 5->2.

```
# coord_mask = [1,0,1,0,0,1]
indices = np.where(coord_mask)[0]    # indices = [0,2,5]
ncoord = indices.shape[0]                   # ncoord = 3
maprange = np.arange(ncoord)          # maprange = [0,1,2]
mapindex = np.zeros(indices[-1] + 1, dtype=int)
mapindex[indices] = maprange          # mapindex = [0,0,1,0,0,2]
```

next map the var indices to the output columns:  [2,0,2,2,5] -> [1,0,1,1,2]
```
coordindices = mapindex[coord_data]. # coordindices = [1,0,1,1,2]
```

Now the obs and var coordinates are in right space, and can be used to broadcast to the 
dense output array:

```
densedata = np.zeros((nrows, ncols), dtype=self.get_X_array_dtype())
densedata[obsindices, varindices] = data[""]
```

server/converters/cxgtool.py
bmccandless: Yeah, that's a good idea.   Mind if I add that as a separate PR?

server/converters/cxgtool.py
bmccandless: This uses the same data access pattern as save_X, so it should work in both places.
I just verified it does work (slowly) with a pretty big matrix.

server/compute/diffexp_cxg.py
bkmartinjr: That seems robust. And I think we can take comfort from the fact that ScanPy has used it for quite some time.

server/converters/cxgtool.py
bkmartinjr: Definitely a separate PR.

server/converters/cxgtool.py
bkmartinjr: great - just wanted to confirm.   Slow is expected given how it actually works under the covers.

server/data_cxg/cxg_adaptor.py
bkmartinjr: this is helpful.  I did eventually puzzle it out when doing the review, but given that you have done this super nice writeup, maybe add it to a comment in the code for the next person :-)

bmccandless: Sounds good.  comment updated.

Issue #871, opened 2019-08-05
Better handle long dataset titles


colinmegill:
See #858, which cuts off the title if it's too long (less than ideal) to prevent overflow onto new line, or out of sidebar area. 

Currently, the full title is available on hover. 

The quickest thing to do is probably to leave things as they are now, but make it more obvious the title is cut off.

liaprins-czi:
@colinmegill Is it possible to do ellipses in the middle of the file name (preserving the first and last portions of the name + file extension)? E.g. how Finder does it (see screen shot below). This would be a good way I feel, since some ppl add version numbers/indicators at the end of the file name, and some at the beginning. This method preserves most of the parts of the file name that are likely to provide helpful context, I feel.

![Screen Shot 2019-08-06 at 9 19 51 AM](https://user-images.githubusercontent.com/53277128/62557354-ad2ca700-b82b-11e9-8c3b-ae4102f72f1f.png)

Here are a few links I found from a quick search, which may or may not provide some clues as to feasibility:

- https://codepen.io/markchitty/pen/RNZbRE
- https://stackoverflow.com/questions/8403086/long-path-with-ellipsis-in-the-middle

colinmegill:
Fantastic idea, yes, we can do that!

liaprins-czi:
@colinmegill Should I assign this one back to you now?

colinmegill:
Seve!

seve:
@liaprins-czi @colinmegill @bkmartinjr , I'm on it!

PR #1500, opened 2020-05-28
Create Truncation Component

### LINKED ISSUE #1489: Consolidate text truncation into a unified component
Text truncation has become common and widespread enough that keeping our approach consistent has become difficult.  The app would benefit from having a `SmartTruncate` component that handles truncation in a smart and consistent fashion.

seve:
Across the app, we practice truncating large strings by taking a slice of some size out of the middle of the string and replacing it with ellipses.  The number of truncations needed throughout the app has begun to grow and the work associated with keeping the methodology between instances has been growing with it.

This PR introduces a Truncation component in the new dir: `client/src/components/util/`.  Given a single child holding text and a width style, the component calculates whether the text needs truncation and truncates to the largest fitting size.

 h/t @tihuan for providing a CSS solution and steering me along the way 🎉 

---

* `client/src/components/util/truncate.js`
  * Created the truncate component
* `client/src/components/categorical/category/index.js` + `client/src/components/categorical/value/index.js` + `client/src/components/leftSidebar/topLeftLogoAndTitle.js`
  * Refactor in `Truncate` component
* `client/src/components/graph/overlays/centroidLabels.js` + `client/src/components/categorical/value/occupancy.js`
  * Remove `fontFamily` style declarations
* `client/configuration/eslint/eslint.js`
  * Removal of redundant/conflicting manual ESLint rules
* `client/configuration/lint-staged/lint-staged.config.js`
  * Correctly lint only staged files instead of all of src

---
Closes #1489


tihuan:
Thanks for the PR, @seve 😄 🎉 !

I haven't looked into middle cropping much, but what's your thought on this CSS solution?

https://codepen.io/markchitty/pen/RNZbRE

I'm thinking if we can outsource the calculation to CSS, we will have less thread blocking calculations in JS and improve our app's performance!

seve:
@tihuan I think your CSS solution could do the job!

I opted for a Javascript method mainly for three reasons:
* Prevent DOM rerendering
* Work with fall back fonts
* Be cognizant of when truncation is occurring and render tooltips accordingly

Since this doesn't require manipulating the DOM once the text is rendered it passes the first one.  If I find the "magic font face number" for both our main font and our backup font it will work with the fallback.  

I'm not too sure about the conditional rendering of tooltips, perhaps we can naively render them on all could-be-truncated strings.  I imagine that might get messy, but it shouldn't be too bad.

seve:
If we want to remove even more thread blocking operations, we could also just remove truncation's guarantee of a polished cellxgene experience if our fonts are not loaded.  I could break out the fontface check to still error on a failed load so we have the functionality.

tihuan:
@seve Great idea!

Yeah if we can find a magic number for `$fontFaceScaleFactor` that works for both `Roboto Condensed` and `sans-serif`, then we probably don't need to do the fontface check too. Especially since it's highly unlikely `Roboto Condensed` doesn't load?

colinmegill:
@tihuan linking back to issue, magic number approach in JS (solution this PR replaces), see: https://github.com/chanzuckerberg/cellxgene/issues/1489#issuecomment-631697832

tihuan:
Thanks for the link, @colinmegill 😄 ! Ahh yeah, `WMWMWM` is the worst lol

I added `iii` and `WMWM` examples in the Codepen to try out the cases in the comment:

https://codepen.io/tihuan/pen/ExVqVNO

Screen Shot 2020-05-29 at 11 39 04 AM

I see that the `WMWM` case, the last three chars got cut off, but not sure how frequently we'll run into such issue and if users will be fine with that

What do you think?

colinmegill:
@seve, what do you think?

seve:
@colinmegill @tihuan I'm going to implement the CSS method and play around with a few datasets.  If I start seeing discrepancies due to character width I might consider going back to JS.  IMO, Tooltips on all + iffy truncation is enough UX friction to justify a slightly more than negligible perf hit.

Colin: Do you have any datasets off the top of your head that would be good test cases? Ideally, I'd like one with long category names.

tihuan:
Sounds awesome! Thanks for experimenting with the CSS approach, @seve ! 🙏 

Also, is it a hard requirement to do middle cropping instead of end cropping? It feels like end cropping is much easier to do?

https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

CC: @colinmegill

seve:
@tihuan Here is the original ask: https://github.com/chanzuckerberg/cellxgene/issues/871#issuecomment-518743340

Funny Lia suggested that same CSS solution way back when

tihuan:
@seve Thanks for the link! Yeah that's really funny 😆 

Ahh I see, so researchers add version number and/or indicators either at the beginning or the end of the name, so we can't cut off the end. 

I just looked at how the CSS is done and tweaked it a little more to avoid chunking off the end:

Screen Shot 2020-06-01 at 12 46 49 PM

```css
// Variables to control the truncation behaviour
$startFixedChars     : 5;    // Number of chars before ellipsis - have priority over end chars
$endFixedChars       : 5;    // Number of chars after ellipsis  - lower priority than start chars
$fontFaceScaleFactor : 0.46; // Magic number dependent on font face - set by trial and error

// Dervied from the 3 variables above
$startWidth : 1em * $fontFaceScaleFactor * ($startFixedChars * 2);
$endWidth   : 1em * $fontFaceScaleFactor * ($endFixedChars * 2);
```

Essentially changing the the fixed chars of start and end to be both `5` and then change `$startWidth` and `$endWidth` to be `$startFixedChars * 2` and `$endFixedChars * 2`, so we will always leave enough width for the ending part to fit the widest 5 char width.

The downside of this is there will be scenarios where we are truncating strings with more lower cases too soon:

Screen Shot 2020-06-01 at 12 50 40 PM

compared to previously:

Screen Shot 2020-06-01 at 12 52 40 PM


There's also a lengthy discussion on Firefox's implementation of the feature. I'm still reading it but so far this looks promising!

https://codepen.io/anon/pen/LdWLem

seve:
@tihuan thanks for the fix!  I'm working on getting a CSS-in-JS implementation up right now.  

In the edge case that the length of the string is less than 10 characters(5 start + 5 end), we get padding in between the two parts with a size equal to the difference. In order to prevent that I could dynamically set `startFixedChars` and `endFixedChars` to fit the string length better, but I feel like I'm slipping back into a JS solution.  Any thoughts?

tihuan:
Ahh thanks for testing the `5 + 5` case! Yeah the gap is really huge 😆 Definitely not good

Screen Shot 2020-06-01 at 2 50 26 PM

What's your thought on this Firefox discussion board's solution?

I added the same test cases from the Stackoverflow one, and it seems to work better:

Screen Shot 2020-06-01 at 2 49 46 PM

https://codepen.io/tihuan/pen/qBbWoog

And the only hard coded value is `min-width: 30px;`!

```css
.split {
  width: 300px;
  background: pink;
  display: flex;
  overflow: hidden;
  justify-content: flex-start;
}

.first {
  overflow: hidden;
  text-overflow: ellipsis;
  white-space: nowrap;
  flex-shrink: 1;
  min-width: 30px;
}

.second > span {
  position: absolute;
  right: 0;
  color: initial;
}
.second {
  color: transparent;
  position: relative;
  overflow: hidden;
  white-space:nowrap;
}
```

PTAL 🎉 !

seve:
@tihuan I think that's the one! I'm going to run it against some other datasets, but its looking good so far!
image

tihuan:
So glad to hear that 😭 !! Thanks for your patience on this one, @seve !! 

If that works for our cases, I'll share the code and credit the author back to the original Stackoverflow solution, so other people can see it too 😆

tihuan:
Woot woot go @seve !! 🎉 🎉 🎉 

Thanks so much for testing this approach thoroughly in the app! (I finally set up cellxgene with Matt's help this morning lol)

I'll share the solution back on Stackoverflow now too!

client/src/components/util/truncate.js
```diff
@@ -0,0 +1,78 @@
+import React, { cloneElement } from "react";
+import { Tooltip, Position } from "@blueprintjs/core";
+
+import { tooltipHoverOpenDelayQuick } from "../../globals";
+
+export default (props) => {
+  const { children } = props;
+  // Truncate only support a single child with a text child
+
+  if (
```
tihuan, line 10: nit: for readability, we can early throw error if the expected condition is not met, so we can avoid nesting our main code block in a giant if condition! 

Like so:

```ts
export default () => {
  const { children } = props;

  // Take care of exception up front
  if (React.Children.count(children) !== 1 && React.Children.count(children.props?.children) !== 1) {
    throw Error("Only pass a single child with text to Truncate");
  }

  // main functionality starts here with no nesting!
  const originalString = children.props.children;
  
  ...
}
```

client/src/components/util/truncate.js
```diff
@@ -0,0 +1,78 @@
...
+      overflow: "hidden",
+      justifyContent: "flex-start",
+    };
+
+    const firstStyle = {
+      overflow: "hidden",
+      textOverflow: "ellipsis",
+      whiteSpace: "nowrap",
+      flexShrink: 1,
+      minWidth: 5,
```
tihuan, line 31: `minWidth: '5px',`?

client/configuration/eslint/eslint.js
```diff
@@ -21,13 +21,7 @@ module.exports = {
     "react/jsx-filename-extension": "off",
     "comma-dangle": "off",
     "no-underscore-dangle": "off",
-    quotes: ["error", "double"],
```
bkmartinjr, line 24: general comment for future PRs - independent changes like this should be in their own PR.

client/src/components/util/truncate.js
```diff
@@ -0,0 +1,78 @@
...
+    const secondString = originalString.substr(originalString.length / 2);
+
+    const splitStyle = {
+      ...children.props.style,
+      display: "flex",
+      overflow: "hidden",
+      justifyContent: "flex-start",
+    };
+
+    const firstStyle = {
```
tihuan, line 26: Since the style objects are constant, we can lift them out of the function declaration, so JS doesn't need to re-instantiate a new object every time this function is called.

E.g.,

```ts
const SPLIT_STYLE = {
  display: "flex",
  overflow: "hidden",
  justifyContent: "flex-start",
}

const FIRST_STYLE = {
  overflow: "hidden",
      textOverflow: "ellipsis",
      whiteSpace: "nowrap",
      flexShrink: 1,
      minWidth: '5px', 
}

const SECOND_STYLE_INNER = {
  position: "absolute",
  right: 0,
  color: "initial",
}

const SECOND_STYLE = {
  color: "transparent",
  position: "relative",
  overflow: "hidden",
  whiteSpace: "nowrap",
}

...

export default (props) => {
    ...

   const splitStyle = { ...children.props.style, ...SPLIT_STYLE }

    const truncatedJSX = (
      <span style={splitStyle}>
        <span style={FIRST_STYLE}>{firstString}</span>
        <span style={SECOND_STYLE}>
          {secondString}
          <span style={SECOND_STYLE_INNER}>{secondString}</span>
        </span>
      </span>
    );
})


```

client/configuration/eslint/eslint.js
bkmartinjr, line 24: general comment for future PRs - independent changes like this should be in their own PR.

tihuan, line 24: thanks for the cleanup!

client/configuration/lint-staged/lint-staged.config.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  "./src/**/*.js": "eslint --fix ./src/",
+  "./src/**/*.js": "eslint --fix",
```
tihuan, line 2: noice fix!

client/src/components/categorical/value/index.js
```diff
@@ -332,6 +324,11 @@ class CategoryValue extends React.Component {
 
     const valueToggleLabel = `value-toggle-checkbox-${displayString}`;
 
+    const labelWidth =
+      colorAccessor && !isColorBy
+        ? globals.leftSidebarWidth - 145 - 15 - 100
```
tihuan, line 329: can we define these magic numbers please :D? Thank you!

client/src/components/leftSidebar/topLeftLogoAndTitle.js
```diff
@@ -3,6 +3,7 @@ import React from "react";
 import { connect } from "react-redux";
 import * as globals from "../../globals";
 import Logo from "../framework/logo";
+import Truncate from "../util/truncate";
```
tihuan, line 6: Uber nit: Should we change the file name to `Truncate.js` to signal that it's a React Component instead of a plain old JS function?

client/src/components/categorical/category/index.js
```diff
@@ -22,6 +22,8 @@ import maybeTruncateString from "../../../util/maybeTruncateString";
   };
 })
 class Category extends React.Component {
+  _labelWidth = globals.leftSidebarWidth - 80;
```
bkmartinjr, line 25: Why is this an instance variable, and not just a variable in the outer (file) closure?

client/src/components/leftSidebar/topLeftLogoAndTitle.js
```diff
@@ -13,16 +14,8 @@ import Logo from "../framework/logo";
...
-    const displayTitle =
-      datasetTitle.length > globals.datasetTitleMaxCharacterCount
-        ? `${datasetTitle.substring(
-            0,
-            Math.floor(globals.datasetTitleMaxCharacterCount / 2)
-          )}…${datasetTitle.slice(
-            -Math.floor(globals.datasetTitleMaxCharacterCount / 2)
-          )}`
-        : datasetTitle;
+    const width = 190;
```
tihuan, line 17: nit: 

`WIDTH = '190px'`
`FONT_SIZE = '14px'`

Even though browser automatically appends `px` for number without explicit units, it's clearer for readers to know the exact unit being used for clarity

Also since they are constants, let's use the conventional UPPER_CASE to name the variables!

client/src/components/leftSidebar/topLeftLogoAndTitle.js
colinmegill: Good q! We have not adopted this pattern elsewhere in the codebase, so it would mean renaming lots of files. Prefer to discuss separately and leave this lowercase for now (ie., if we decide to change, change all at once).

tihuan: SGTM!

client/src/components/leftSidebar/topLeftLogoAndTitle.js
colinmegill: We have a globals file for times that we need to do this, and there are long descriptive names in there so we explicitly export / import things like `export const cellxgeneTitleLeftPadding = 14;` These might fit nicely there.

client/src/components/util/truncate.js
```diff
@@ -0,0 +1,78 @@
...
+    };
+    const secondStyle = {
+      color: "transparent",
+      position: "relative",
+      overflow: "hidden",
+      whiteSpace: "nowrap",
+    };
+
+    const truncatedJSX = (
+      <span style={splitStyle}>
```
colinmegill, line 47: @seve prefer no variables here, prefer inlining for readability (that's the convention elsewhere in the codebase).

client/src/components/leftSidebar/topLeftLogoAndTitle.js
seve: @colinmegill Should globals be holding variables that aren't "global".  It may be hurt to globalize variables and hide their values before they are accessed from outside the file.  I think file-scoped constants may have a place in our standards.

client/__tests__/e2e/__snapshots__/e2e.test.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`did launch page launched 1`] = `"<span style=\\"width: 185px; display: flex; overflow: hidden; justify-content: flex-start;\\"><span style=\\"overflow: hidden; text-overflow: ellipsis; white-space: nowrap; flex-shrink: 1; min-width: 5px;\\">pbm</span><span style=\\"color: transparent; position: relative; overflow: hidden; white-space: nowrap;\\">c3k<span style=\\"position: absolute; right: 0px; color: initial;\\">c3k</span></span></span>"`;
+
+exports[`metadata loads categories and values from dataset appear 1`] = `undefined`;
```
tihuan, line 5: is this one expected to be `undefined` 😄 ? Thank you!

seve: Oof no, I'll have to take a look into that. I didn't even think of looking to see what was being stored in the snapshot. Thanks for catching this!

tihuan: My pleasure 🚀 ! Yeah curious to see what caused the `undefined` 😆

PR #1502, opened 2020-05-29
sparse column shift encoding.

bmccandless:
Many of our matrices are log normalized, which tends to eliminate
the number of non zero values (if there were any).  This prevents
the matrix from being stored as a sparse matrix.  The solution here
is to use a simple transformation to make it sparse again.  The most
common value from each column is subtracted from that column.  These
values that were subtracted are saved in an array called X_col_shift.

The cellxgene code needs to understand how to undo the transformation when
operating over the X matrix.

### DIFF
--- a/server/converters/cxgtool.py
+++ b/server/converters/cxgtool.py
@@ -10,6 +10,8 @@ The organization of the TileDB structure is:
     ├─ var                          TileDB array containing gene (column) attributes, with one attribute per
     │                               dataframe column, shape (n_obs,)
     ├─ X                            Main count matrix as a 2D TileDB array, single unnamed numeric attribute
+    ├─ X_col_shift                  TilebDB Array used in column shift encoding, shape (n_var,).
+    │                               Single unnamed numeric attribute.  May only exist if X is sparse.
     ├─ emb                          TileDB group, storing optional embeddings (group may be empty)
     │  └─ <name1>                   TileDB Array, single anon attribute, ND numeric array, shape (n_obs, N)
     └─ cxg_group_metadata           Empty array used only to stash metadata about the overall object.
@@ -70,6 +72,7 @@ import argparse
 import numpy as np
 from os.path import splitext, basename
 import json
+from scipy.stats import mode
 
 from server.common.colors import convert_anndata_category_colors_to_cxg_category_colors
 from server.common.errors import ColorFormatException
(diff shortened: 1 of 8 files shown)


server/converters/cxgtool.py
```diff
@@ -10,6 +10,8 @@
     ├─ var                          TileDB array containing gene (column) attributes, with one attribute per
     │                               dataframe column, shape (n_obs,)
     ├─ X                            Main count matrix as a 2D TileDB array, single unnamed numeric attribute
+    ├─ X_col_shift                  TilebDB Array used in column shift encoding, shape (n_var,).
```
bkmartinjr, line 13: I would further clarify in the comment (assuming I understand the design): the X_col_shift type must be identical to the X type (ie, float32, int32, etc)

server/converters/cxgtool.py
```diff
@@ -455,11 +458,39 @@ def check_sparse_X(adata, sparse_threshold):
     return (nnz, shape[0] * shape[1])
 
 
+def check_for_sparse_column_shift_encoding(adata, sparse_threshold):
```
bkmartinjr, line 461: function name is  misleading - it doesn't do the check, it merely computes the info needed to perform the check.   Might make it more self-explanatory.

bkmartinjr: comment is also somewhat misleading ("Determine if ...")

server/test/test_diffexp.py
```diff
@@ -88,16 +88,34 @@ def test_cxg_generic(self):
     def test_cxg_sparse(self):
         with tempfile.TemporaryDirectory() as dirname:
             sparsename = os.path.join(dirname, "sparse.cxg")
+            sparse_col_shift_name = os.path.join(dirname, "sparse_col_shift.cxg")
             densename = os.path.join(dirname, "dense.cxg")
             source_h5ad = anndata.read_h5ad(f"{PROJECT_ROOT}/example-dataset/pbmc3k.h5ad")
+
             # create a cxg sparse array
+            # TODO:  pbmc3k.h5ad is 100% non-zero.  Converting it to sparse is useful, but it would
```
bkmartinjr, line 96: I suggest creating a couple of synthetic H5AD files to test the various encoding schemes, and include it with this PR so we have better test coverage.

server/converters/cxgtool.py
bmccandless: I'll try to make it more clear.  I currently returns two pieces of information,  the nnz/nelems that it processed, and the col_shift array that contains the encoding.  If it determines that col shift encoding is not a benefit, the col_shift is None.

server/converters/cxgtool.py
bmccandless: Added more comments here.

server/test/test_diffexp.py
bmccandless: Sounds good, added.

bmccandless:
> Several suggestions:
> 
> 1. we are starting to add enough complexity and optionality to CXG that we need to up our testing game.   Could we add more test datasets that stress the edge conditions?  For example, it would be simple to create a synthetic H5AD that would be useful to validate sparse encoding and the column shift.
> 2. This column shift relies on absolute equality for the values it makes sparse (ie, the value must be precisely zero, or precisely the column mode).   In practice, this data is somewhat noisy, and encoded as floats, so there are cases where values may cluster around a mode, with a small epsilon of difference.  We already down-sample to 32 bit floats, so you might also consider moving the sparse value test from "eq mode" to "within epsilon of mode".
> 3. I noted that the new tiledb 2 has a "vacuum", which must be used alongside "consolidate".  We do not appear to be doing that.


1.  Added new tests
2.  I'm hesitant to change this now.   The equal test is working fine, with about 16 of the datasets in dev converting to col shift with very small nnz percents.   Changing it to be fuzzy would make it more expensive and complicated.
3. Added vacuum.

Issue #1513, opened 2020-06-04
multi-dataroot/bucket 


bkmartinjr:
Enable multi-dataroot/bucket deployment of WSGI cellxgene, with each root being served via a different base URL, and then reconfigure cellxgene.cziscience.com deployment to support one namespace for Covid-19 data and another for Corpora data

PR #1531, opened 2020-06-04
add config option to handle multiple dataroots

bmccandless:
#1513

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -58,7 +58,7 @@ def cache_control_always(**cache_kwargs):
 
 @webbp.route("/", methods=["GET"])
 @cache_control(public=True, max_age=ONE_WEEK)
-def dataset_index(dataset=None):
+def dataset_index(url_dataroot=None, dataset=None):
     config = current_app.app_config
     if dataset is None:
         if config.single_dataset__datapath:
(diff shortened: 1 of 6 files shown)


server/common/app_config.py
```diff
@@ -375,6 +389,16 @@ def handle_multi_dataset(self, context):
         if self.multi_dataset__dataroot is None:
             return
 
+        if type(self.multi_dataset__dataroot) == str:
+            self.multi_dataset__dataroot = dict(d=self.multi_dataset__dataroot)
+
+        for key in self.multi_dataset__dataroot.keys():
+            # sanity check for well formed keys
+            if type(key) != str:
```
bkmartinjr, line 397: Suggest you add an error check that the key is URL-safe, ie, the key only contains characters that are not reserved for special use in URL.  If you want to be super-duper safe, mandate only characters in the `unreserved` set for all URIs, which are 

   unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"

From [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)

You could probably include a bit more punctuation, but I'm not sure it is all that useful.  For example, JS encodeURI() treats the following as unreserved:  A-Z a-z 0-9 ; , / ? : @ & = + $ - _ . ! ~ * ' ( ) #

server/app/app.py
```diff
@@ -91,18 +94,21 @@ def health():
     return health_check(config)
```
bkmartinjr, line 94: I think you need to add multi-dataroot support to `health_check` as well.

Issue #1510, opened 2020-06-04
smoke test remote deployments


bkmartinjr:
Make smoke test suite more robust against remote deployments, allowing us to reduce the human testing required for a release.

PR #1548, opened 2020-06-04
1510-smoke-test

tihuan:
Added a few things for our smoke tests. Please see the comments for more details!

1. Add retry twice before failing a test

2. Add [`expect-puppeteer`](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer) to avoid using Puppeteer's error prone low level API where possible

3. Puppeteer utils and cellxgene action utils are now top level exports for ease of use

4. Split `testBrowser.js` function into two config files:
    1.  `client/jest-puppeteer.config.js` is for configuring Puppeteer's launch config options
    1. `client/__tests__/e2e/puppeteer.setup.js` is for configuring `jest`, `browser`

5. Tests are now atomic with its own setup step, so tests can be run individually

6. Add a few `npm` commands and a flag `HEADFUL=true` to improve test writing experience

7. Introduce the use of `ndb` to be the recommended way to debug tests [source](https://github.com/GoogleChromeLabs/ndb)

NOTE: Currently updating developer Readme to reflect the new changes

Thanks for your time to review!

closes https://github.com/chanzuckerberg/cellxgene/issues/1510

🚨Please run `npm install` due to the new dependencies!🚨


client/src/components/autosave/filenameDialog.js
```diff
@@ -113,7 +113,7 @@ class FilenameDialog extends React.Component {
             this.handleCreateFilename();
```
tihuan, line 113: added attributes for testing

client/package.json
```diff
@@ -5,12 +5,16 @@
...
   "repository": "https://github.com/chanzuckerberg/cellxgene",
   "scripts": {
-    "clean": "rimraf build",
     "build": "npm run clean && webpack --config",
-    "prod": "npm run build -- configuration/webpack/webpack.config.prod.js",
+    "clean": "rimraf build",
     "dev": "npm run build -- configuration/webpack/webpack.config.dev.js",
-    "fmt": "eslint --fix src",
-    "lint": "eslint src"
+    "e2e": "jest --config __tests__/e2e/e2eJestConfig.json e2e/e2e.test.js",
```
tihuan, line 11: resorted object keys alphabetically and added extra commands `e2e`, `e2e-annotations`, and `e2e-prod` for ease of development

client/package.json
```diff
@@ -88,13 +92,16 @@
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-react": "^7.20.0",
     "eslint-plugin-react-hooks": "^4.0.4",
+    "expect-puppeteer": "^4.4.0",
```
tihuan, line 95: added these 3 libraries for better testing developer experience, and allow us to use `expect-puppeteer`'s API instead of puppeteer's error prone low level API:

https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer

client/src/components/graph/drawPointsRegl.js
```diff
@@ -1,6 +1,6 @@
 import { glPointFlags, glPointSize } from "../../util/glHelpers";
```
tihuan, line 1: linter at work! no exported anonymous function

client/src/components/scatterplot/drawPointsRegl.js
```diff
@@ -1,6 +1,6 @@
 import { glPointFlags, glPointSize } from "../../util/glHelpers";
```
tihuan, line 1: linter at work! no exported anonymous function

client/src/globals.js
```diff
@@ -1,5 +1,8 @@
 import { Colors } from "@blueprintjs/core";
 import { dispatchNetworkErrorMessageToUser } from "./util/actionHelpers";
+import { CXG_SERVER_PORT as DEFAULT_CXG_SERVER_PORT } from "../__tests__/e2e/config";
+
+const CXG_SERVER_PORT = process.env.CXG_SERVER_PORT || DEFAULT_CXG_SERVER_PORT;
```
tihuan, line 5: default `CXG_SERVER_PORT` to `5005` as specified in `__tests__/e2e/config`

common.mk
```diff
@@ -21,6 +21,7 @@ export CXG_CLIENT_PORT := $(call get_or_else_dev_env_default,CXG_CLIENT_PORT)
 export JEST_ENV := $(call get_or_else_dev_env_default,JEST_ENV)
 export DATASET := $(call get_or_else_dev_env_default,DATASET)
 export CXG_OPTIONS := $(call get_or_else_dev_env_default,CXG_OPTIONS)
+export PROD := $(call get_or_else_dev_env_default,PROD)
```
tihuan, line 24: export constant `PROD`

client/__tests__/e2e/puppeteer.setup.js
```diff
@@ -0,0 +1,55 @@
...
+// `RETRY_ATTEMPTS` in `screenshot_env.js`
+jest.retryTimes(2);
+
+beforeEach(async () => {
+  await jestPuppeteer.resetBrowser();
+
+  const userAgent = await browser.userAgent();
+  await page.setUserAgent(`${userAgent}bot`);
+
+  await page._client.send("Animation.setPlaybackRate", { playbackRate: 12 });
```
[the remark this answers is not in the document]
mweiden: Huh. Neat!

client/__tests__/e2e/puppeteerUtils.js
```diff
@@ -1,73 +1,129 @@
-const puppeteerUtils = (page) => ({
```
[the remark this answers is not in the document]
mweiden: Looking much cleaner!

client/Makefile
```diff
@@ -34,13 +34,17 @@ start-frontend:
 
 .PHONY: e2e
 e2e:
-	node node_modules/jest/bin/jest.js \
+	JEST_ENV=${PROD} \
+	CXG_CLIENT_PORT=${CXG_SERVER_PORT} \
```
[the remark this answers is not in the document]
mweiden: tl;dr: I think you can delete the addition of these environment variables in `e2e` and `e2e-annotations`. Details follow:

* The intent behind `common.mk` was that it introduced defaults from one source of truth, `environment.default`. Having overriding env vars in the code and Makefiles kinda breaks that convention. This might be OK, I'm less zealous about there being more than one source of truth on this that I used to be.
* You could make `JEST_ENV=prod` in `environment.default` and delete the variable `PROD` there and `JEST_ENV` here. Cleaner and has the same effect 🙂
* Switching `CXG_CLIENT_PORT` and `CXG_SERVER_PORT` makes my head spin a bit. Our naming convention isn't great, but I think we should stick to `CLIENT` referring to the frontend and `SERVER` referring to the backend API.

client/package.json
mweiden: This duplicates the code that runs the test suites. The other copies are in `client/Makefile`. It makes sense to me to pick one to be authoritative: either make or npm scripts. This is more DRY and reduces the chance that some future developer will run the test one way, it passes, and then it goes out to CI, is run with a different entrypoint, and fails.

Some history: I refactored these a while ago and (showing my BE-engineering bias a bit here here) I put them all in make. This had the advantage of being able to parameterize runs of node with environment variables from `environment.default`, locking everything together.

common.mk
mweiden: see comment above about potentially eliminating this var

client/__tests__/e2e/config.js
```diff
@@ -1,11 +1,13 @@
-export const jestEnv = process.env.JEST_ENV;
-export const appPort = process.env.CXG_SERVER_PORT;
+// # Similar constants are defined in file: `environment.default`
+export const DEV = "dev";
```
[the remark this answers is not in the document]
mweiden: See my comment about choosing either `make` or `npm` scripts. I don't think we should do both.

If we keep the make targets, we don't have to copy the default environment variables here.

client/__tests__/e2e/config.js
```diff
@@ -1,11 +1,13 @@
...
+export const DEBUG = "debug";
+export const PROD = "prod";
+export const CXG_SERVER_PORT = 5005;
+export const CXG_CLIENT_PORT = 3000;
+export const jestEnv = process.env.JEST_ENV || PROD;
+export const appPort = process.env.CXG_CLIENT_PORT || CXG_CLIENT_PORT;
```
mweiden, line 8: See my comment about switching `CXG_CLIENT_PORT` and `CXG_SERVER_PORT` in `client/Makefile`. I think we should avoid this switcheroo.

client/Makefile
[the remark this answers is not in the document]
tihuan: Thanks for the great feedback and background on this, @mweiden ! I see what you mean now 😄 

This idea just came to me, but what's your thought on converting `environment.default` to `environment.default.json`, so both Python and JS can consume it, and we will have that as the single source of truth?

Re: `CXG_CLIENT_PORT` and `CXG_SERVER_PORT`: 

Yeah the only reason I had to do this is that GHA's e2e tests hit pages at `CXG_SERVER_PORT`(5005), but FE local development hit `CXG_CLIENT_PORT` (3000) (cuz we spin up `make start-frontend`).

So I decided to update `config.js` with:

```ts
export const CXG_CLIENT_PORT = 3000;
export const appPort = process.env.CXG_CLIENT_PORT || CXG_CLIENT_PORT;
```

So e2e tests will hit 3000 by default, and CI/CD to hit 5005 through passing `CXG_CLIENT_PORT=${CXG_SERVER_PORT}` in the make file:

```
JEST_ENV=${PROD} \
	CXG_CLIENT_PORT=${CXG_SERVER_PORT} \
		node node_modules/jest/bin/jest.js \
		--verbose false \
		--config __tests__/e2e/e2eJestConfig.json \
		e2e/e2e.test.js
```

client/package.json
tihuan: Thanks for the great idea, @mweiden ! I just updated `client/Makefile` to use the npm scripts:

E.g.,

```
.PHONY: e2e
e2e:
	JEST_ENV=${PROD} \
	CXG_CLIENT_PORT=${CXG_SERVER_PORT} \
		npm run e2e -- \
		--verbose false \

.PHONY: e2e-annotations
e2e-annotations:
	JEST_ENV=${PROD} \
	CXG_CLIENT_PORT=${CXG_SERVER_PORT} \
	npm run e2e-annotations -- \
		--verbose false \
```
 -- 

I only duplicated some vars from `environment.default` to `e2e/config.js` because I wanted to set:

```ts
export const appPort = process.env.CXG_CLIENT_PORT || CXG_CLIENT_PORT;

export const appUrlBase =
  process.env.CXG_URL_BASE || `http://localhost:${appPort}`;
```

This allows FE engineers to optionally run npm scripts locally with add-on tools like [`ndb`](https://github.com/GoogleChromeLabs/ndb) for debugging, and extra flags, like `-u` to update snapshots and `HEADFUL=true` to run tests in headful mode regardless of `JEST_ENV` value (prod, dev, debug)

But I was thinking maybe turning `environment.default` into `environment.default.json` would mean both Python and JS could share the same source of truth 😄 ?

Thank you!

tihuan:
Ooh didn't know about `e2e_tests.md`, will update that one on top of `developer_guidelines.md` 🎉 Thanks @bkmartinjr !

client/Makefile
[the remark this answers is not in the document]
tihuan: Ah just realized that it's not about sharing environment defaults between JS and Python, but between JS and Makefile 😆 

In that case, maybe this would do?

```
define GetValueFromJson
$(shell node -p '\
    const getVal = (key = "", obj = {}) => {
          const currKey = key.split(".")[0];
          const val = obj[currKey];
          if(typeof val !== "object") return val;
          const nextKey = key.split(".").slice(1).join(".");
          return getVal(nextKey, val);
    }; \
    getVal(`$(1)`.replace(" ", ""), require("./package.json")); \
')
endef

PORT := $(call GetValueFromJson, config.port)
```

[source](https://stackoverflow.com/questions/36664010/read-json-and-assign-to-a-list-of-make-variables)

tihuan:
@mweiden I added some more replies to your comments 😄 PTAL thanks a lot!

client/Makefile
```diff
@@ -34,17 +34,15 @@ start-frontend:
 
 .PHONY: e2e
 e2e:
-	node node_modules/jest/bin/jest.js \
+	APP_PORT=${CXG_SERVER_PORT} \
```
tihuan, line 37: @mweiden Ahh I see what you're saying now! Before I had:

```
CXG_SERVER_PORT=${CXG_SERVER_PORT} \
```

Which is confusing, because I really wanted to set which port Puppeteer should hit, instead of overwriting the default `CXG_SERVER_PORT` value

So I changed the name to `APP_PORT` now, and using it in `client/__tests__/e2e/config.js` like so:

```ts
export const appPort = process.env.APP_PORT || ENV_DEFAULT.CXG_CLIENT_PORT;
```

Which means we can change the makefile script here to be:

```
APP_PORT=${CXG_SERVER_PORT} \
```

Does it look better now? Thank you!

client/Makefile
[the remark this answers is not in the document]
tihuan: Just updated the PR to use `environment.default.json`!

client/Makefile
```diff
@@ -36,19 +36,15 @@ start-frontend:
 e2e:
 	JEST_ENV=${PROD} \
 	CXG_CLIENT_PORT=${CXG_SERVER_PORT} \
-		node node_modules/jest/bin/jest.js \
+		npm run e2e -- \
```
tihuan, line 38: also uses `npm run e2e` in makefile script now 👌

.github/workflows/push_tests.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: |
           pip install flake8
           cd client
-          npm i "eslint" "eslint-config-airbnb" "eslint-config-prettier" "eslint-loader" "eslint-plugin-filenames" "eslint-plugin-import" "eslint-plugin-jest" "eslint-plugin-jsx-a11y" "eslint-plugin-react" "eslint-plugin-react-hooks" "eslint-plugin-prettier"
```
tihuan, line 35: using `npm install`, so we'll install `eslint` and friends following the versions and deps defined in `package-lock.json`. This prevents surprises such as linter working locally, but not in GHA.

If `npm install` taking a long time is the problem, we can always cache it!

client/package.json
tihuan: Using `environment.default.json` now 🎉 !

tihuan:
@mweiden I was thinking about extracting the scripting part from this PR, but then realized that the debugging tool (ndb) relies on the new npm scripts 😆 So I've updated the PR to use `environment.default.json` now. PTAL again! Thank you 🙏

client/Makefile
mweiden: nit: Given that you are now bringing the environment into `config.js` via `environment.default.json`, isn't it possible to remove the `APP_PORT` variable entirely and have two variables/concepts (`CXG_SERVER_PORT` and `CXG_CLIENT_PORT`) instead of three?

in `config.js`, this could be:
```
export const clientPort = process.env.CXG_CLIENT_PORT || ENV_DEFAULT.CXG_CLIENT_PORT;
export const appUrlBase = process.env.CXG_URL_BASE || `http://localhost:${clientPort}`;
```

This change would also make your make target simpler.

client/Makefile
mweiden: Nice!

client/package.json
mweiden: Looks better!

client/Makefile
```diff
@@ -21,51 +21,33 @@ WEBPACK_CONFIG ?= configuration/webpack/webpack.config.prod.js
...
+# start an instance of cellxgene and run the end-to-end annotations tests
 .PHONY: smoke-test-annotations
 smoke-test-annotations:
 	$(eval TMP_DIR := $(shell  mktemp -d /tmp/cellxgene_XXXXXX))
 	cp $(ANNOTATIONS) $(TMP_DIR)/ && \
 	start_server_and_test \
 		'CXG_OPTIONS="--annotations-file $(TMP_DIR)/$(ANNOTATIONS_FILENAME)" $(MAKE) start-server' \
 		$(CXG_SERVER_PORT) \
-		'$(MAKE) e2e-annotations'
+		'CXG_URL_BASE="https://localhost:$(CXG_SERVER_PORT)" npm run e2e-annotations -- --verbose false'
```
tihuan, line 45: ooh nice collapsing the logic here 🔥 👏 !!

dev_docs/developer_guidelines.md
```diff
@@ -41,7 +41,7 @@ The `JEST_ENV` environment variable enables the following E2E test options:
 Run end to end tests interactively during development
```
tihuan, line 41: YAY! Thanks for updating the docs too! I'll add more for the new tools introduced in the story now as well 😄 Really appreciate your help, @mweiden !! 🎉

client/Makefile
mweiden: @tihuan just trying to further deduplicate things here. Tests are failing though. Trying to figure out why.

tihuan:
Ooh let me check errors!

tihuan:
Hi all! Just updated docs to introduce the new tools we can use, such as `ndb` and extra explanation on what different config files are used for [(diff)](https://github.com/chanzuckerberg/cellxgene/pull/1548/commits/012a88a5b34ac687d08b7437432c75f82990212c):

PTAL at the following links for the additions:

1. https://github.com/chanzuckerberg/cellxgene/blob/1510-smoke-test/dev_docs/developer_guidelines.md
2. https://github.com/chanzuckerberg/cellxgene/blob/1510-smoke-test/dev_docs/e2e_tests.md

If it all looks good I'll merge on tomorrow 😄 🥳 🎉 !

Thanks for helping to make this PR so much better!

CC: @mweiden

PR #1549, opened 2020-06-05
refactor categorical controls state

bkmartinjr:
This PR refactors the reducer store `categoricalSelection` state, which had become a mashup of:
* categorical selection state
* summary information about the categorical annotations

The `categoricalSelection` reducer is now solely selection state, and the data summary is computed on demand by the <Category> component.   

This refactoring is a step toward the upcoming annotated matrix refactor.

### DIFF
--- a/client/src/components/categorical/category/annoDialogEditCategoryName.js
+++ b/client/src/components/categorical/category/annoDialogEditCategoryName.js
@@ -8,9 +8,9 @@ import { labelPrompt } from "../labelUtil";
 import { AnnotationsHelpers } from "../../../util/stateManager";
 
 @connect((state) => ({
-  categoricalSelection: state.categoricalSelection,
   annotations: state.annotations,
   universe: state.universe,
+  schema: state.world?.schema,
   ontology: state.ontology,
 }))
 class AnnoDialogEditCategoryName extends React.PureComponent {
@@ -36,10 +36,15 @@ class AnnoDialogEditCategoryName extends React.PureComponent {
   };
 
   handleEditCategory = (e) => {
-    const { dispatch, metadataField, categoricalSelection } = this.props;
+    const { dispatch, metadataField } = this.props;
     const { newCategoryText } = this.state;
 
-    const allCategoryNames = _.keys(categoricalSelection);
+    /*
+    test for uniqueness against *all* annotation names, not just the subset
+    we render as categorical.
+    */
+    const { schema } = this.props;
+    const allCategoryNames = schema.annotations.obs.columns.map((c) => c.name);
 
     if (
       (allCategoryNames.indexOf(newCategoryText) > -1 &&
@@ -60,7 +65,7 @@ class AnnoDialogEditCategoryName extends React.PureComponent {
   };
 
   editedCategoryNameError = (name) => {
-    const { metadataField, categoricalSelection } = this.props;
+    const { metadataField } = this.props;
 
     /* check for syntax errors in category name */
     const error = AnnotationsHelpers.annotationNameIsErroneous(name);
@@ -69,7 +74,14 @@ class AnnoDialogEditCategoryName extends React.PureComponent {
     }
 
     /* check for duplicative categories */
-    const allCategoryNames = _.keys(categoricalSelection);
+
+    /*
+    test for uniqueness against *all* annotation names, not just the subset
+    we render as categorical.
+    */
+    const { schema } = this.props;
+    const allCategoryNames = schema.annotations.obs.columns.map((c) => c.name);
+
     const categoryNameAlreadyExists = allCategoryNames.indexOf(name) > -1;
     const sameName = name === metadataField;
     if (categoryNameAlreadyExists && !sameName) {
--- a/client/src/components/categorical/category/index.js
+++ b/client/src/components/categorical/category/index.js
@@ -10,6 +10,7 @@ import AnnoDialogAddLabel from "./annoDialogAddLabel";
 
 import * as globals from "../../../globals";
 import maybeTruncateString from "../../../util/maybeTruncateString";
+import { createCategorySummary } from "../../../util/stateManager/controlsHelpers";
 
 @connect((state, ownProps) => {
   const { metadataField } = ownProps;
@@ -18,6 +19,7 @@ import maybeTruncateString from "../../../util/maybeTruncateString";
     categoricalSelection: state.categoricalSelection,
     annotations: state.annotations,
     universe: state.universe,
+    world: state.world,
     schema: state.world?.schema,
   };
 })
@@ -26,36 +28,56 @@ class Category extends React.Component {
     super(props);
     this.state = {
       isChecked: true,
+      categorySummary: this.createCategorySummary(),
     };
   }
 
+  createCategorySummary() {
+    const { world, metadataField } = this.props;
+    if (!world || !metadataField || !world.obsAnnotations.hasCol(metadataField))
+      return null;
+    return createCategorySummary(world, metadataField);
+  }
+
   componentDidUpdate(prevProps) {
-    const { categoricalSelection, metadataField } = this.props;
+    const { categoricalSelection, metadataField, world } = this.props;
+    let { categorySummary } = this.state;
+
+    if (
+      world !== prevProps.world ||
+      metadataField !== prevProps.metadataField ||
+      !categorySummary
+    ) {
+      const newCategorySummary = this.createCategorySummary();
+      if (categorySummary !== newCategorySummary) {
+        categorySummary = newCategorySummary;
+        this.setState({ categorySummary });
+      }
+    }
+
     const cat = categoricalSelection?.[metadataField];
     if (
       categoricalSelection !== prevProps.categoricalSelection &&
       !!cat &&
       !!this.checkbox
     ) {
-      const categoryCount = {
-        // total number of categories in this dimension
-        totalCatCount: cat.numCategoryValues,
-        // number of selected options in this category
-        selectedCatCount: _.reduce(
-          cat.categoryValueSelected,
-          (res, cond) => (cond ? res + 1 : res),
-          0
-        ),
-      };
-      if (categoryCount.selectedCatCount === categoryCount.totalCatCount) {
+      // total number of categories in this dimension
+      const totalCatCount = categorySummary.numCategoryValues;
+      // number of selected options in this category
+      const selectedCatCount = categorySummary.categoryValues.reduce(
+        (res, label) => (cat.get(label) ?? true ? res + 1 : res),
+        0
+      );
+
+      if (selectedCatCount === totalCatCount) {
         /* everything is on, so not indeterminate */
         this.checkbox.indeterminate = false;
         this.setState({ isChecked: true }); // eslint-disable-line react/no-did-update-set-state
-      } else if (categoryCount.selectedCatCount === 0) {
+      } else if (selectedCatCount === 0) {
         /* nothing is on, so no */
         this.checkbox.indeterminate = false;
         this.setState({ isChecked: false }); // eslint-disable-line react/no-did-update-set-state
-      } else if (categoryCount.selectedCatCount < categoryCount.totalCatCount) {
+      } else if (selectedCatCount < totalCatCount) {
         /* to be explicit... */
         this.checkbox.indeterminate = true;
         this.setState({ isChecked: false }); // eslint-disable-line react/no-did-update-set-state
@@ -83,25 +105,28 @@ class Category extends React.Component {
 
   toggleNone() {
     const { dispatch, metadataField } = this.props;
+    const { categorySummary } = this.state;
     dispatch({
       type: "categorical metadata filter none of these",
       metadataField,
+      labels: categorySummary.categoryValues,
     });
     this.setState({ isChecked: false });
   }
 
   toggleAll() {
     const { dispatch, metadataField } = this.props;
+    const { categorySummary } = this.state;
     dispatch({
       type: "categorical metadata filter all of these",
       metadataField,
+      labels: categorySummary.categoryValues,
     });
     this.setState({ isChecked: true });
   }
 
   handleToggleAllClick() {
     const { isChecked } = this.state;
-    // || this.checkbox.indeterminate === false
     if (isChecked) {
       this.toggleNone();
     } else {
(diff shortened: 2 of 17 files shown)


client/src/components/categorical/category/annoDialogEditCategoryName.js
```diff
@@ -69,7 +73,14 @@ class AnnoDialogEditCategoryName extends React.PureComponent {
...
 
     /* check for duplicative categories */
-    const allCategoryNames = _.keys(categoricalSelection);
+
+    /*
+    test for uniqueness against *all* annotation names, not just the subset
+    we render as categorical.
+    */
+    const { schema } = this.props;
+    const allCategoryNames = schema.annotations.obs.columns.map((c) => c.name);
```
colinmegill, line 82: 👍

client/src/components/categorical/category/index.js
```diff
@@ -30,36 +31,49 @@ class Category extends React.Component {
...
+
+    if (
+      world !== prevProps.world ||
+      metadataField !== prevProps.metadataField ||
+      !categorySummary
+    ) {
+      const newCategorySummary = this.createCategorySummary();
+      if (categorySummary !== newCategorySummary) {
+        categorySummary = newCategorySummary;
+        this.setState({ categorySummary }); // eslint-disable-line react/no-did-update-set-state
```
tihuan, line 50: I could be missing something, but what's your thought on listing `categorySummary` creation into `mapStateToProps()` inside `@connect`? And if `_createCategorySummary()` call is expensive, we could memoize it based on `world` and `metadataField`?

E.g.,

```ts
@connect((state, ownProps) => {
  const { metadataField } = ownProps;
  const { world } = state

  return {
    world,
    isColorAccessor: state.colors.colorAccessor === metadataField,
    categoricalSelection: state.categoricalSelection,
    categorySummary: categorySummary(world, metadataField), // <-- NEW
    annotations: state.annotations,
    universe: state.universe,
    schema: state.world?.schema,
  };
})
```

client/src/components/categorical/value/index.js
```diff
@@ -21,20 +21,17 @@ import { AnnotationsHelpers } from "../../../util/stateManager";
...
 
 /* this is defined outside of the class so we can use it in connect() */
-function _currentLabel(ownProps, categoricalSelection) {
-  const { metadataField, categoryIndex } = ownProps;
-  return String(
-    categoricalSelection[metadataField].categoryValues[categoryIndex]
-  ).valueOf();
+function _currentLabelAsString(ownProps) {
+  const { categorySummary, categoryIndex } = ownProps;
+  return String(categorySummary.categoryValues[categoryIndex]).valueOf();
```
tihuan, line 26: Seems like we can remove `.valueOf()`?

`String(categorySummary.categoryValues[categoryIndex]) === String(categorySummary.categoryValues[categoryIndex]).valueOf()`

client/src/components/categorical/value/index.js
```diff
@@ -171,16 +192,24 @@ class CategoryValue extends React.Component {
...
+    const newLabel = newCategorySummary.categoryValues[newCategoryIndex];
+    const labelChanged = label !== newLabel;
     const valueSelectionChange =
-      categoricalSelection[metadataField].categoryValueSelected[
-        categoryIndex
-      ] !==
-      newCategoricalSelection[metadataField].categoryValueSelected[
-        categoryIndex
-      ];
+      (categoricalSelection[metadataField].get(label) ?? true) !==
```
tihuan, line 211: for my education, what edge case does `??` help take care of?

I'm imagining `categoricalSelection[metadataField]` could be either `undefined` or a value, so just wondering if `categoricalSelection[metadataField].get(label) !== newCategoricalSelection[metadataField].get(newLabel)` would do?

client/src/components/graph/overlays/centroidLabels.js
```diff
@@ -57,7 +54,7 @@ class CentroidLabels extends PureComponent {
         fontWeight = "800";
       }
 
-      const selected = categoryValueSelected[categoryValueIndices.get(label)];
+      const selected = category.get(label) ?? true;
```
tihuan, line 57: this seems to imply if `category.get(label)` is `null` or `undefined`, we want to set `selected` to `true`. But in that case, what is selected?

client/src/reducers/categoricalSelection.js
```diff
@@ -30,30 +32,25 @@ const CategoricalSelection = (
...
     case "categorical metadata filter select": {
       /*
       Set the specific category in this field to false
       */
-      const newCategoryValueSelected = Array.from(
-        state[action.metadataField].categoryValueSelected
-      );
-      newCategoryValueSelected[action.categoryIndex] = true;
+      const { metadataField, label } = action;
+      const newSelected = new Map(state[metadataField]);
```
tihuan, line 49: I personally love `Map`, but `Map` is not serializable, so if ever want to serialize the Redux store as JSON, extra processing is needed. If that's not a concern, then we're good!

>ES2015 Map is not a serializable data structure so you need extra works if you need to persist your Store data. Redux DevTools doesn’t display data of ES2015 Map by default, so you have to specify serialize option for this.

[source](https://medium.com/@koba04/redux-store-patterns-using-es2015-map-f62804755675#4955)

client/src/components/categorical/value/index.js
colinmegill: Good catch @tihuan, I would agree. @bkmartinjr was there a corner case here?

client/__tests__/e2e/config.js
```diff
@@ -1,10 +1,11 @@
...
 export const appUrlBase =
-  process.env.CXG_URL_BASE || `http://localhost:${appPort}`;
+	process.env.CXG_URL_BASE || `http://localhost:${appPort}`;
 export const DEV = jestEnv === "dev";
 export const DEBUG = jestEnv === "debug";
 export const DATASET = "pbmc3k";
 
-if (DEBUG) jest.setTimeout(100000);
-if (DEV) jest.setTimeout(10000);
+if (DEBUG) jest.setTimeout(240000);
```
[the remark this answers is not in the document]
bkmartinjr: I'm happy to do it but given that the native units are ms, I'm not sure how this improves readability.

client/__tests__/e2e/e2eAnnotations.test.js
```diff
@@ -207,7 +208,7 @@ describe.each([
       node.getAttribute("aria-label")
     );
 
-    expect(result).toBe(categoryName);
+    return expect(result).toBe(categoryName);
```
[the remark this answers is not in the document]
bkmartinjr: correct.  Not something I added, but I'll fix it.

client/src/components/categorical/category/annoDialogEditCategoryName.js
```diff
@@ -36,10 +35,15 @@ class AnnoDialogEditCategoryName extends React.PureComponent {
...
+    const { dispatch, metadataField } = this.props;
     const { newCategoryText } = this.state;
 
-    const allCategoryNames = _.keys(categoricalSelection);
+    /*
+    test for uniqueness against *all* annotation names, not just the subset
+    we render as categorical.
+    */
+    const { schema } = this.props;
+    const allCategoryNames = schema.annotations.obs.columns.map((c) => c.name);
```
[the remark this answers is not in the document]
bkmartinjr: I generally don't find adding functions for simpler one-liners is a win, as it is more code, not less.  But will add.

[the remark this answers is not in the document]
colinmegill: Wrapping in this case does add clarity IMO

client/src/reducers/categoricalSelection.js
bkmartinjr: we can't use an object, as the keys are native types (and must be native types).  I used Map explicitly, not as a replacement for Object.

client/src/components/graph/overlays/centroidLabels.js
bkmartinjr: correct, the default state for missing values is selected (aka true).

all that is happening here is that our default selection state is true, and because we asynchronously load labels, we don't know all of them at reducer init time.  So if the reducer store does not yet "know" about the labels inside the data, we presume they are selected.

client/src/components/categorical/value/index.js
bkmartinjr: the corner case is when the reducer does not (yet) know about the selection state (because data is being loaded or otherwise transformed).  In this case, our selection state defaults to true, and if you change the code as proposed, it would eval incorrectly (it changes the default to false).

client/src/components/categorical/value/index.js
bkmartinjr: We want the primitive string type, not the String object.   If we remove the valueOf, we get the String() object.

client/src/components/categorical/category/index.js
bkmartinjr: sure, that would work today, but it won't work with the (soon to land) asynchronous data loading that motivated this PR.  So I would have to revert that change almost immediately :-)

TL;DR - I'd rather leave it in state, which is where it must be soon.

client/src/components/categorical/value/index.js
bkmartinjr: turns out that when the constructor is called as a function, it does return a primitive.  Only when used with the new operator does it return a String object.

So we can remove the valueOf call.

client/src/components/categorical/category/index.js
tihuan: hmm since Redux's `connect()` is triggered whenever the Redux store has an update, creating `categorySummary` could still happen in `mapStateToProps()` instead of in the view component?

What I'm thinking is, if Redux state doesn't have the needed data, `categorySummary(world, metadataField)` will return something boring, like `null` or `undefined`. But once the store is updated, `categorySummary(world, metadataField)` will automatically get called again via `connect()`. So it should work in async model too?

client/src/components/categorical/value/index.js
tihuan: Synced with Bruce via Slack!

Gonna try to summarize my understanding here:

Inside the rendering logic, we default the selection state of a label to `true`, if a user hasn't touched the label yet. AKA: `if (categoricalSelection[metadataField].get(label)) === undefined) return true`

So inside `shouldComponentUpdate()` here, `const valueSelectionChange` needs to check if the selection state of a label has changed from `true` to `false`, or `false` to `true`.

But since `categoricalSelection[metadataField].get(label))` could be `undefined`, we need to coalesce `undefined` to `true`, so when in async model, we don't trigger an unnecessary re-render when `categoricalSelection` state gets populated asynchronously:

E.g.,

Phase 1: `categoricalSelection` is NOT populated with persisted data yet:

```ts
shouldComponentUpdate() {
  const oldSelection = categoricalSelection[metadataField]?.get(label) // undefined
  const newSelection = newCategoricalSelection[metadataField]?.get(newLabel) // undefined

  const valueSelectionChange = oldSelection !== newSelection // false, since undefined === undefined
}
```

Phase 2: `categoricalSelection` is now asynchronously loaded with the persisted data:

```ts
shouldComponentUpdate() {
  const oldSelection = categoricalSelection[metadataField]?.get(label) // undefined
  // Turned out that `newLabel` is actually selected according to the persisted data, instead of unknown
  const newSelection = newCategoricalSelection[metadataField]?.get(newLabel) // true

  // This causes unnecessary re-render, because `undefined` should be coalesced to `true`
  const valueSelectionChange = oldSelection !== newSelection // true, since undefined !== true.
}
```

So I'm assuming when we have the async model, there's a case that `categoricalSelection` could be asynchronously populated with `true` as a default value for a category's label?

Because if a category in `categoricalSelection` is **always** populated with `new Map()`, then by default, `undefined` gets coalesced to `true`, so all labels are selected. And therefore a label's selection needs to transition from `undefined` to `false` in `categoricalSelection`, **before** it can transition to `true`?

For example,

Step 1: 
User sees all labels are selected
`categoricalSelection[aCategory].get(aLabel)`: `undefined`

Step 2:
User deselects label: `aLabel`
`categoricalSelection[aCategory].get(aLabel)`: `false`

Step 3:
User reselects label: `aLabel`
`categoricalSelection[aCategory].get(aLabel)`: `true`

And there doesn't exist a state transition of a label's selection from `undefined` to `true`, because `undefined` is `true` already, so the Phase 2 comparison scenario will never happen, and therefore:

```ts
const valueSelectionChange =
  categoricalSelection[metadataField].get(label) !== newCategoricalSelection[metadataField].get(newLabel)
```

would work?

I think I'm really close to understanding the reason behind using `?? true` here now lol 

Thanks for your patience, everyone 😂

client/src/components/graph/overlays/centroidLabels.js
tihuan: Got it! Thanks for the contexts, Bruce 😄 Makes sense now!

client/src/reducers/categoricalSelection.js
tihuan: Ahh I see, so `state[metadataField]` returns `Map`, and we create a copy of it via `new Map(state[metadataField])` 👌 !

client/src/components/categorical/category/index.js
```diff
@@ -184,11 +202,7 @@ class Category extends React.Component {
 
     const isUserAnno =
       schema?.annotations?.obsByName[metadataField]?.writable ?? false;
-    const isTruncated = _.get(
-      categoricalSelection,
-      [metadataField, "isTruncated"],
-      false
-    );
+    const isTruncated = categorySummary?.isTruncated ?? false;
```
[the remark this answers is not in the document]
bkmartinjr: actually, we don't even need the Boolean().   Just `categorySummary?.isTruncated` is equivalent.  I'll fix.

client/src/components/categorical/category/index.js
bkmartinjr: true, but it would be wasted work, as it only needs to be run when the data changes.

Currently that "data" is in the reducer state, but very soon it won't be.

client/src/reducers/categoricalSelection.js
bkmartinjr: yep!

client/src/components/categorical/value/index.js
bkmartinjr: resolved offline.  fix pushed.

PR #1554, opened 2020-06-09
If min / max are the same, hide histo + legend, show value

### LINKED ISSUE #1497: Gene expression not shown in gene's histogram after subset (not all genes, though)
Found when testing with https://cellxgene.dev.single-cell.czi.technology/d/WongAdultRetina.cxg/

I had added the gene KCNQ1, then subset to 12,117 cells by lassoing (will be difficult for others to to reproduce due to this). Then the following phenomena resulted:
-  KCNQ1's min number on its x-axis became a very long-digited negative decimal, so long it does not fit in the space provided under the x-axis. Its max number became also a negative decimal.
- I could not select anything on the histogram; there were no bars shown: none of the cells remaining after subset have any KCNQ1 expression I think. However, KCNQ1 should still include “0” as part of its x-axis and then show the tall bar there (if all cells have 0-level of expression).
- The scatterplot shows nothing when that gene is chosen for one of the axes.
- Furthermore, when I color by that gene, it colors everything purple, but the color scale legend is black.

Because lasso-subsetting created this bug, it will be difficult to reproduce. So I hit "undo" bunch until the point at which I subset, then made a video of me hitting "redo" a bunch to show the sequence of events. Giphy made the resolution to small to see the numbers, **so I made a video instead: https://drive.google.com/open?id=1kFXpmuru2XnWVa-O8jAwMgUqL2XtcdjP**

Here are screenshots:

_The long x-axis tick number_
Screen Shot 2020-05-27 at 9 27 49 AM

_The Console message + scatterplot without any dots_
Screen Shot 2020-05-27 at 9 31 19 AM

_The inspector to see what the long digited decimal was trying to show_
Screen Shot 2020-05-27 at 9 32 19 AM

_Coloring by KCNQ1 strangely makes all dots purple, and the legend black!_
Screen Shot 2020-05-27 at 9 33 54 AM

Mysterious!

colinmegill:
Fixes #1497

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -556,7 +556,7 @@ class HistogramBrush extends React.PureComponent {
           </Tooltip>
         </div>
         <svg
-          style={{ display: isSingleValue ? "none" : "auto" }}
+          style={{ display: isSingleValue ? "none" : "block" }}
           width={this.width}
           height={this.height}
           id={`histogram_${fieldForId}_svg`}
@@ -573,7 +573,7 @@ class HistogramBrush extends React.PureComponent {
           <span
             style={{
               color: unclippedRangeMinColor,
-              display: isSingleValue ? "none" : "auto",
+              display: isSingleValue ? "none" : "block",
             }}
           >
             min {unclippedRangeMin.toPrecision(4)}
@@ -590,7 +590,7 @@ class HistogramBrush extends React.PureComponent {
           <span
             style={{
               color: unclippedRangeMaxColor,
-              display: isSingleValue ? "none" : "auto",
+              display: isSingleValue ? "none" : "block",
             }}
           >
             max {unclippedRangeMax.toPrecision(4)}
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -476,6 +476,8 @@ class HistogramBrush extends React.PureComponent {
     const unclippedRangeMaxColor =
       world.clipQuantiles.max === 1 ? "#bbb" : globals.blue;
 
+    const isSingleValue = unclippedRangeMin === unclippedRangeMax;
+
     return (
       <div
         id={`histogram_${fieldForId}`}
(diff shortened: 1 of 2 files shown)


client/src/components/brushableHistogram/index.js
```diff
@@ -476,6 +476,8 @@ class HistogramBrush extends React.PureComponent {
     const unclippedRangeMaxColor =
       world.clipQuantiles.max === 1 ? "#bbb" : globals.blue;
 
+    const isSingleValue = unclippedRangeMin === unclippedRangeMax;
```
bkmartinjr, line 479: what behavior are you targeting when the user has clipped the range, so that it is all single-valued or NaN (clipped)?   In this case, the NaN doesn't show up in the histogram, and the histogram will end up being single-valued.

So shouldn't the conditional be based upon the min/max of the clipped range, rather than the unclipped range?

client/src/components/continuousLegend/index.js
```diff
@@ -115,7 +116,13 @@ class ContinuousLegend extends React.Component {
       d3.select("#continuous_legend").selectAll("*").remove();
     }
 
-    if (colorAccessor && colorScale && colorScale.range) {
+    if (
```
bkmartinjr, line 119: this would be a lot more readable if it was something like:
```
const range = colorScale?.range;
const [domainMin, domainMax] = colorScale?.domain?.() ?? [0, 0];
if (colorAccessor && colorScale && range && domainMin < domainMax) {
  ...
}
```

and then use those variables further down

client/src/components/continuousLegend/index.js
```diff
@@ -128,7 +135,16 @@ class ContinuousLegend extends React.Component {
   }
 
   render() {
-    const { colorAccessor } = this.props;
+    const { colorAccessor, colorScale } = this.props;
+
+    if (
+      colorScale &&
+      colorScale.domain &&
```
seve, line 142: You can use optional chaining here:
```suggestion
      colorScale?.domain &&
```

PR #1557, opened 2020-06-10
Componentize bar charts

seve:
This PR breaks out the occupancy and mini histogram graphs into `miniHistogram` and `miniStackedBar` components


seve:
@mweiden any idea why the codecov bot didn't fire on this PR?

client/src/components/categorical/value/index.js
```diff
@@ -19,6 +19,8 @@ import Truncate from "../../util/truncate";
 
 import { AnnotationsHelpers } from "../../../util/stateManager";
 import { labelPrompt, isLabelErroneous } from "../labelUtil";
+import MiniHistogram from "../../expressionGraphs/miniHistogram";
```
colinmegill, line 22: These charts are not specific to `expression`. Recommend top level `src/components/miniHistogram/index.js`

client/src/components/categorical/value/index.js
```diff
@@ -336,11 +427,11 @@ class CategoryValue extends React.Component {
       LABEL_MARGIN +
       (isUserAnno ? ANNO_MENU : 0);
 
-    const OCCUPANCY_WIDTH = 100;
+    const GRAPH_WIDTH = 100;
```
colinmegill, line 430: For better or worse, we use `graph` for the main scatterplot. Alternatives include a more specific `CANVAS_WIDTH`, `CHART_WIDTH` or `MINI_HISTO_WIDTH`

client/src/components/categorical/value/index.js
```diff
@@ -494,8 +615,8 @@ class CategoryValue extends React.Component {
               display={isColorBy && categories ? "auto" : "none"}
               style={{
                 marginLeft: 5,
-                width: 11,
-                height: 11,
+                width: VALUE_HEIGHT,
```
colinmegill, line 618: Since this is inside the `value/index.js` component, it could also be `this.height` as a constant set in the constructor

mweiden:
@seve seems that the push tests did not run for this branch/PR. Not sure why since you didn't change the workflow and they seem to run fine on other branches.

Screen Shot 2020-06-11 at 12 40 17 PM

seve:
> @seve seems that the push tests did not run for this branch/PR. Not sure why since you didn't change the workflow and they seem to run fine on other branches.

@mweiden I'm going to push a dummy commit and see if that jump-starts it. Otherwise, I may just abandon this PR.

client/src/components/categorical/value/index.js
seve: @colinmegill, I created expressionGraphs to bundle the occupancy and mini histogram in the same folder.  Do you have a suggestion for a folder name that they would fit in?

seve: I named them expression graphs because they both show expression, whether that be of categorical or continuous data

client/src/components/categorical/value/index.js
seve: I'm going to opt-out of doing this to consolidate all the dimension constants in one location

client/src/components/categorical/value/index.js
seve: Moved to individual component folders

PR #1574, opened 2020-06-17
Check for lassoPolygon before attempting to push

### LINKED ISSUE #1563: TypeError: null is not an object (evaluating 'o.push')
Sentry Issue: [HOSTED-CELLXGENE-R](https://sentry.prod.si.czi.technology/sci-sc/hosted-cellxgene/issues/3458/?referrer=github_integration)

```
TypeError: null is not an object (evaluating 'o.push')
  at apply (./src/components/graph/setupLasso.js:50:7)
...
(1 additional frame(s) were not displayed)
```

colinmegill:
Fixes #1563

### DIFF
--- a/client/src/components/graph/setupLasso.js
+++ b/client/src/components/graph/setupLasso.js
@@ -45,7 +45,10 @@ const Lasso = () => {
 
     const handleDrag = () => {
       const point = d3.mouse(svg.node());
-      lassoPolygon.push(point);
+      if (lassoPolygon) {
+        lassoPolygon.push(point);
+      }
+
       lassoPath.attr("d", polygonToPath(lassoPolygon));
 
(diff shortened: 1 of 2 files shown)


client/src/components/graph/setupLasso.js
```diff
@@ -7,9 +7,15 @@ const Lasso = () => {
...
 
   const lasso = (svg) => {
-    let lassoPolygon;
-    let lassoPath;
-    let closePath;
+    let lassoPolygon = null;
+    let lassoPath = null;
+    let closePath = null;
+
+    const resetState = () => {
```
bkmartinjr, line 14: rather than defining a new function, why not just use the existing `reset()` function?

PR #1576, opened 2020-06-17
check if bin creation returned null before rendering charts

### LINKED ISSUE #1573: MiniStackedBar error
I clipped/subset a data set down to a very small amount of data, then colored by a category.  

```
TypeError: Cannot destructure property 'length' of 'n' as it is undefined.
    at Fp.drawStacks (index.js:30)
    at Fp.componentDidMount (index.js:48)
    at oc (react-dom.production.min.js:212)
    at ds (react-dom.production.min.js:255)
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Bi (react-dom.production.min.js:122)
    at hs (react-dom.production.min.js:248)
    at Jc (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Bi (react-dom.production.min.js:122)
    at qi (react-dom.production.min.js:123)
    at Wi (react-dom.production.min.js:122)
    at I (react-dom.production.min.js:287)
    at Yt (react-dom.production.min.js:68)
    at HTMLDocument.sentryWrapped (helpers.ts:85)
```

Test case:

1. Load https://cellxgene.staging.single-cell.czi.technology/d/krasnow_lab_human_lung_cell_atlas_smartseq2-2.cxg/
2. Open category "comparment", and select only "endothelial"
3. Subset
4. Color-by category "label"

boom....

The stack trace clearly shows that `drawStacks()` is called from `componentDidMount()`.  Setting a breakpoint in this case shows that the component props do not contain `domainValues`, hence the error.

```
this.props
{
  height: 11, 
  width: 100, 
  colorScale: ƒ
}
```

seve:
The mini histogram and mini stacked bar charts were not checking to see if they were receiving null values for props. This PR conditionally renders those components based on their prop values.

---

Fixes #1573

### DIFF
--- a/client/src/components/categorical/value/index.js
+++ b/client/src/components/categorical/value/index.js
@@ -427,9 +427,12 @@ class CategoryValue extends React.Component {
       world,
     } = this.props;
 
+    const isColorBy = metadataField === colorAccessor;
+
     if (
       !this.shouldRenderStackedBarOrHistogram ||
-      !categoricalSelection[colorAccessor]
+      !categoricalSelection[colorAccessor] ||
+      isColorBy
     ) {
       return null;
     }
@@ -692,7 +695,7 @@ class CategoryValue extends React.Component {
           </div>
           <span style={{ flexShrink: 0 }}>
             {this.renderMiniStackedBar(value)}
-            {this.renderMiniMiniHistogram(value)}
+            {this.renderMiniHistogram(value)}
           </span>
         </div>
         <div>
--- a/client/src/components/categorical/value/index.js
+++ b/client/src/components/categorical/value/index.js
@@ -22,11 +22,14 @@ import { labelPrompt, isLabelErroneous } from "../labelUtil";
 import MiniHistogram from "../../miniHistogram";
 import MiniStackedBar from "../../miniStackedBar";
 
+const VALUE_HEIGHT = 11;
+const CHART_WIDTH = 100;
+
 /* this is defined outside of the class so we can use it in connect() */
 function _currentLabelAsString(ownProps) {
   const { categorySummary, categoryIndex } = ownProps;
   // when called as a function, the String() constructor performs type conversion,
-  // and returns a primtive string.
+  // and returns a primitive string.
   return String(categorySummary.categoryValues[categoryIndex]);
 }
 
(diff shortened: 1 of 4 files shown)


tihuan:
Thanks for the fix, @seve !

Echoing @bkmartinjr 's point on encapsulating rendering logic, I created a PR for reference 😄 

https://github.com/chanzuckerberg/cellxgene/pull/1577

PTAL! Thank you!

client/src/components/categorical/value/index.js
```diff
@@ -480,6 +480,37 @@ class CategoryValue extends React.Component {
           CHART_MARGIN
         : globals.leftSidebarWidth - otherElementsWidth;
 
+    let stackedBins;
+    let histogramBins;
+
+    // If coloring by and this isn't the colorAccessor and it isn't being edited
```
tihuan, line 486: Unsure if the comment reflects the condition, since the comment says "if coloring by" while the condition is `&& !isColorBy &&`, which seems to mean it's NOT coloring by?

Thank you!

seve: Mistake of poor verbiage, I meant coloring by in the general sense that _something_ is being colored by not that the present value is being colored by.  I'll try to clear this up with better wording

tihuan: Ah thank you! The misunderstanding is likely because I'm not familiar with the code base yet 😆 But thanks for the clarification!

tihuan:
Ah for the test, can we add it in https://github.com/chanzuckerberg/cellxgene/pull/1548 please? This way I don't have to port it over once this PR is merged lol Thanks a lot! 🙏

seve:
@bkmartinjr 
> * need a smoke test to ensure we don't trip over this again.

Smoke test is now deferred to #1548

client/src/components/categorical/value/index.js
```diff
@@ -408,6 +418,99 @@ class CategoryValue extends React.Component {
     return false;
   }
 
+  renderMiniStackedBar = (categoryValue) => {
```
bkmartinjr, line 421: is there a reason this method is created per-instance, rather than per-class?  Ie, why not declare it:

```
renderMiniStackedBar(categoryValue) {
    ... 
}
```

PR #1577, opened 2020-06-17
feedback-1573

tihuan:


client/src/components/categorical/value/index.js
```diff
@@ -22,11 +22,20 @@ import { labelPrompt, isLabelErroneous } from "../labelUtil";
...
 import MiniStackedBar from "../../miniStackedBar";
 
+const VALUE_HEIGHT = 11;
+const LEFT_MARGIN = 60;
+const CHECKBOX = 26;
+const CELL_NUMBER = 50;
+const ANNO_MENU = 26;
+const LABEL_MARGIN = 16;
+const CHART_WIDTH = 100;
+const CHART_MARGIN = 24;
```
seve, line 32: What is the performance gain from bringing these out of the class?  I opted to leave them in-line to increase readability b/c I thought the visual pros outweighed the performance.

client/src/components/categorical/value/index.js
```diff
@@ -297,7 +313,7 @@ class CategoryValue extends React.Component {
     height
   ) => {
     /*
-      Knowing that colorScale is based off continuous data, 
+      Knowing that colorScale is based off continuous data,
```
seve, line 316: Was this a visual catch or was it an extension? I'd like to be able to catch these too :)

client/src/components/categorical/value/index.js
tihuan: Ah, I lift them out of the `render()` function because the new methods `this.renderMiniStackedBar()` and `this.renderMiniHistogram()` need access to a few of these constants, so they can't be in `render()` function anymore. But I should put the rest back in `render()` for readability!

Thanks for raising this!

client/src/components/categorical/value/index.js
tihuan: Oh great question! VSCode has an option to trim that 😄 

`"files.trimTrailingWhitespace": true,`

https://github.com/prettier/prettier/issues/2422#issuecomment-313843775

client/src/components/categorical/value/index.js
tihuan: Just moved most of the constants back 😄 PTAL again! Thank you 🙏

PR #1592, opened 2020-06-26
FIX-test-target-closed: remove jestPuppeteer.resetBrowser()

tihuan:
This PR fixes some unknown `target closed` errors, likely due to `jestPuppeteer.resetBrowser()` is run before the previous test is done.

`jestPuppeteer.resetBrowser()` was introduced to make sure each test has its own isolated state, so no browser state such as cookies are shared. This was helpful at Meta because we have AuthN, so some tests need to run in logged in mode and some logged out. But setting `browserContext: "incognito"` seems to be enough for cellxgene right now, since at least it will start a new tab for each test, and we don't have AuthN cookies yet.

We can take a look at using `jestPuppeteer.resetBrowser()` in a reliable way once cellxgene has AuthN. But for now I think it's less noise to not reset browser 😄 

Thank you!

### DIFF
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -149,7 +149,7 @@ describe("cell selection", () => {
 describe("gene entry", () => {
   test("search for single gene", async () => {
     await goToPage(appUrlBase);
-    addGeneToSearch(data.genes.search);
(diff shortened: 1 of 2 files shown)


seve:
Cya `target closed` errors 🎉

tihuan:
Thanks for the quick review, @seve!! Merging 🎉

Issue #1590, opened 2020-06-25
multi-dataroot can have different configuration options


bmccandless:
Each dataroot can have different configuration options.
A use case is to allow datasets in one dataroot to require authentication and permit user annotations,
and datasets on another dataroot will not require authentication and not permit user  annotations

PR #1596, opened 2020-06-29
refactor config to support different config options for datasets in different dataroots.

bmccandless:
This PR gives the ability to specify different config options for different dataroots.

The default_config.py shows that the config is now divided into two sections (server, and dataroot).  

When a data adaptor is initialized, it is provided with the app_config, as well as a dataset config.
The dataset config may be different for different datasets depending on the path that was used 
to access to the dataset.

#1590


server/common/default_config.py
```diff
@@ -1,129 +1,148 @@
...
+    server_timing_headers: false
+    csp_directives: null
+
+  multi_dataset:
+    # If dataroot is set, then cellxgene may serve multiple datasets.  This parameter is not
+    # compatible with single_dataset/datapath.
+    # dataroot may be a string, representing the path to a directory or S3 prefix.  In this
+    # case the datasets in that location are accessed from <server>/d/<datasetname>.
+    # example:
+    #     dataroot:  /path/to/datasets/
```
bkmartinjr, line 23: If I understand correctly, this is a shorthand for:

```
dataroot:
    d: /path/to/datasets/
```

If that is true, would it simplify things to EOL the shortcut and only support the dictionary config syntax (ie, the long form)?  It would have the benefit of removing the magic "/d/" from the code base.

bmccandless: Yes, I could do that.   One small issue is that "--dataroot" is a command line option, which is pretty handy.  I could either 1) remove the --dataroot command line option, which means that a to setup a multi dataset local server you need to have a config file or 2).  keep it, and replace it with the dictionary, { "d" : dataroot },  when evaluating the config options.     In either case it gets rid of the special handling in the app.py code.  In the second case it moves the special handling to app_config.py.  The special handling in app_config.py would be pretty trivial.

bkmartinjr: I think the CLI params could be left dirt simple, and only cover a subset of what you can do in the config.  So in this case (`--dataroot`) it would just be a shortcut...ie, your second option.

bmccandless:
I went back and forth on the best way to organize the config file.   At one point I had something similar to what you suggested, and ended up with the current approach mainly because it made the single dataset approach simpler, and the multi dataset with single dataroot case simpler.  I figured those would be the common cases, and the multi dataroot cases would probably have few users outside of us.  

In the current PR, the single dataset and multi dataset / single dataroot case are fairly straight forward:

```
server:
   (options for single or multi-dataset)

dataset:
   (the config parameters for all datasets)
```

I think in your suggest those cases would look like this, which has the extra "default" level.  Maybe that's not a big deal, but one goal I had was to make the common cases simple.

```
server:
   multi-dataset:  true or false

data:
   default:
      (the config parameters for all datasets)
```

Another wrinkle was how to handle things like baseURL and location.  They are used by the server to identify and access datasets, and therefore not really a property of the dataset config, but rather a property of the server.   I think following your suggestion we would have to put in some non intuitive logic around baseURL.   For example if this was a multi dataset but not a multi-dataroot case, the default.baseURL could be filled in or given a default value ("/d").   And for the single dataset case it would be ignored.   But if this was a multi dataset / multi dataroot case then the baseURL should not be set in the "default" section, but must be filled in by the others (set1, set2).   Furthermore, we would need to check that the baseURL across all sections were unique.  This isn't a big deal, but just points to a subtle issues.

I don't really have a strong preference for my approach, but wanted to let you know I did weigh some options. 

Your right that the single-dataset options like "about" and "title" are an oddity.

bkmartinjr:
> Another wrinkle was how to handle things like baseURL and location. They are used by the server to identify and access datasets, and therefore not really a property of the dataset config, but rather a property of the server.

Thinking about this more, my root issue is not the organization as much as the use of the dataset URL path as the name in the config file.  While they can commonly be the same, forcing them to be the same seems like it will make future cases of aliasing, re-naming, etc ripple through the config files.  There is also the issue of wanting to use URL paths for data that are not a single path element (eg, server from /d/ and from /foo/bar/baz/).

Did you consider separating the URL path from the dataset "name" in the config?  Ie, doing something like the following, where each dataset/dataroot has an "ID", and a separate config param for the URL path element?
```
server:
    id_1: /s3/path/to/bucket
    someOtherName: /another/path/to/a/bucket

dataset:
    # defaults for all dataroots

per_dataset_config:
    id_1:
        URL_path: "d"
        about_legal_tos: tos_for_id_1.html

    someOtherName:
        URL_path: "foo/bar/baz"
```

bmccandless:
> Did you consider separating the URL path from the dataset "name" in the config? Ie, doing something like the following, where each dataset/dataroot has an "ID", and a separate config param for the URL path element?
> 
> ```
> server:
>     id_1: /s3/path/to/bucket
>     someOtherName: /another/path/to/a/bucket
> 
> dataset:
>     # defaults for all dataroots
> 
> per_dataset_config:
>     id_1:
>         URL_path: "d"
>         about_legal_tos: tos_for_id_1.html
> 
>     someOtherName:
>         URL_path: "foo/bar/baz"
> ```

Yeah, that looks like a good approach.  I can rework the patch to go with that.

server/common/app_config.py
```diff
@@ -423,13 +574,114 @@ def handle_multi_dataset(self, context):
...
+            return False
+        return value > limit_value
+
+
+class DatasetConfig(BaseConfig):
+    """Manages the config attribute associated with a dataset."""
+
+    def __init__(self, tag, app_config, default_config):
+        super().__init__(app_config, default_config)
+        self.tag = tag
```
bkmartinjr, line 619: the dataroot tag must be a legal URL path element.  I can't find code that validates that this is true, and it is probably worth adding.  Will help catch malformed configuration.

server/common/default_config.py
```diff
@@ -1,129 +1,148 @@
...
+    port : null
+    open_browser: false
+    force_https: false
+    flask_secret_key: null
+    generate_cache_control_headers: false
+    server_timing_headers: false
+    csp_directives: null
+
+  multi_dataset:
+    # If dataroot is set, then cellxgene may serve multiple datasets.  This parameter is not
```
bkmartinjr, line 18: it seems a bit odd that the single/multi-dataset mode is enabled by a sub-parameter of multi_dataset configuration.

Maybe it would be clearer with a top level server param that sets the mode directly, which would allow a config file to have full configs or both modes, but only one enabled at a time.

Eg,
```
server:
    data_serving_mode: "multi"  # or "single"
```

Or alternatively:
```
server:
    serve_multiple_datasets: true
```

server/common/default_config.py
```diff
@@ -1,129 +1,148 @@
 import yaml
 
 default_config = """
```
bkmartinjr, line 3: Future task - we need a light weigh document that describes every config parameter, its default, etc.  Not for this PR, but would be good to file an issue and link them.

server/common/health.py
```diff
@@ -24,10 +24,13 @@ def health_check(config):
     health = {"status": None, "version": "1", "releaseID": cellxgene_version}
 
     checks = False
-    if config.single_dataset__datapath is not None:
-        checks = _is_accessible(config.single_dataset__datapath, config)
-    elif config.multi_dataset__dataroot is not None:
-        checks = all([_is_accessible(datapath, config) for datapath in config.multi_dataset__dataroot.values()])
+    server_config = config.server_config
+    if server_config.single_dataset__datapath is not None:
```
bkmartinjr, line 28: these checks (first check single, then check multi) are the opposite of what the comment in default_config.py says (it says that multi is enabled by setting mult_dataset__dataroot).  This code will assume single if single is set, regardless of multil.

server/app/app.py
```diff
@@ -184,33 +188,38 @@ class SchemaAPI(DatasetResource):
     @cache_control(public=True, max_age=ONE_WEEK)
     @rest_get_data_adaptor
     def get(self, data_adaptor):
-        return common_rest.schema_get(data_adaptor, current_app.annotations)
+        annotations = data_adaptor.dataset_config.user_annotations
```
bkmartinjr, line 191: given that `user_annotations` config is now managed by the data adaptor, why not push this code down into the rest.py layer, and not repeat it for each route handler?  (I may be missing the reason?)   Just seems like repetitive code in this layer, if annotations is always in the data_adaptor without fail.

server/common/app_config.py
bmccandless: The check was added in the original PR for multi data root.  It's in the checks for multi_dataset/dataroot attribute.  There is also a check that the tag in DatasetConfig must be one of the keys in multi_dataset/dataroot.   

I think given the previous discussion that plan is to add a  "URL_path" to the dataset config.  I'd then move the check for legal URL path elements to that config attribute, which would be clearer.

server/common/default_config.py
bmccandless: It is clunky right now, agreed.  I'll take your suggestion to add a top level server param.

server/common/default_config.py
bmccandless: Agreed.  I've started commenting as I add new ones, but there are many that are not commented.

server/common/health.py
bmccandless: The code only lets you have multi_dataset__dataroot set OR single_dataset__datapath set.  If you set both it will give a configuration error.  The rest of the code can then assume that only one is set.  But I'll change that to have a top level server param, like you suggested above, then there won't be any logic like this.

server/app/app.py
bmccandless: Yeah, that's a good idea.    No reason not to do it.

server/common/default_config.py
bmccandless: I remembered my I did it the way I did now.  
In practice I expected either the multi_dataset OR the single_dataset block to appear, not both.
So, if the user wanted multi_dataset, then only that block would appear in the config file, and "dataroot" would be a required parameter.  

I think if we made a change with that it would be orthogonal to this PR, and we can do it separately if needed.

One improvement through would be to have an "is_single_dataset()" or "is_multi_dataset()" function on the AppConfig object, so that the logic in the rest of the code is simpler.

bkmartinjr: I'm OK with it being a separate PR, but I think (in principle) that flags/modes set implicitly, ie, by inferring from other flags, is generally bad form.  It makes the system hard to understand.  I can live with it for small stuff, but this is a very fundamental mode for the system.   Just making the argument about clarity and comprehensibility of the system.

PR #1604, opened 2020-07-02
exclude a tiledb version that has a known bug

bmccandless:


server/requirements.txt
```diff
@@ -18,6 +18,6 @@ pandas>=0.24.2
 PyYAML>=5.3
 scipy>=1.3.0
 requests>=2.22.0
-tiledb>=0.5.9,!=0.6.0
+tiledb>=0.5.9,!=0.6.0,!=0.6.1
```
bkmartinjr, line 21: Isn't this equivalent to >=0.6.2

bmccandless: Yes, it is.  I'll update.

PR #1628, opened 2020-07-14
simply state management

### LINKED ISSUE #1475: Large datasets make scatterplot cells look unselected due to tiny-ness (and actual unselected ones are invisible)
When I plot genes on the scatterplot, the unselected cells are too tiny to show up in the scatterplot (and the selected ones are also very tiny, the size of unselected ones in the main vis. I am testing with a large dataset when this happened, and it is probably related to #1325 Colin made wherein the main vis dots disapper for large dataset in small window. I do not think this one is enough of an issue to stop the release. Here is a screenshot showing how difficult it is to even see the selected cells in the scatterplot, and how they look “unselected” due to their tinyness:
Screen Shot 2020-05-13 at 10 17 51 AM

Probably related to #1325 .

I was testing with this dataset:
https://cellxgene.dev.single-cell.czi.technology/d/HumanTissueTcellActivation.cxg/

### LINKED ISSUE #1624: consider getDerivedStateFromProps()
Some of our code in componentDidUpdate/Mount would be better served by using getDerivedStateFromProps().

See for example the new graph/overlay/centroidLabels.js

Background:

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html just in case!

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change

bkmartinjr:
Variety of changes to simplify state management, as a second step in the redux refactor:
* viewport size tracking simplified for both graph and scatterplot
* as part of scatterplot viewport tracking, fixed long-standing bug #1475 

I did not find any components which would benefit from using getDerivedStateFromProps() now that we have done the async refactor.

Fixes #1475 
Fixes #1624

### DIFF
--- a/client/src/components/graph/graph.js
+++ b/client/src/components/graph/graph.js
@@ -191,7 +191,7 @@ class Graph extends React.Component {
       camera: null,
       modelTF,
       modelInvTF: mat3.invert([], modelTF),
-      projectionTF: null,
+      projectionTF: createProjectionTF(viewport.width, viewport.height),
 
       // regl state
       regl: null,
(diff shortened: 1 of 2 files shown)


seve:
Looks sane to me so far, I'll take a closer look tomorrow morning 👍

Issue #1578, opened 2020-06-17
Design a new dataset for testing automation 


seve:
We would benefit from adding/replacing our current testing dataset(pbmc3k) with a deeper and more diverse dataset.  This could be a joining of multiple existing datasets that stress our typing system. Would require a refactoring of our hard-coded values and snapshots.

signechambers1:
@seve @maniarathi is this still relevant? do you want to include as some part of https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/single-cell/44?

maniarathi:
Good call! I added it to our smoketest epic!

signechambers1:
@maniarathi the smoke test epic is closed, ok to close this ticket?

maniarathi:
Beefing up our set of valid-but-unusual datasets is still a thing we need to work on to be honest and we had to trim our efforts in testing due to time constraints. I'd like to keep this open because the ideas listed above are good, but I'm not sure when to get to it. I'd mark it as a P3.

PR #1637, opened 2020-07-16
fix truncation edge cases

seve:
This PR fixes two issues related to formatting of the Truncation component when given certain forms of input.

1. Input with whitespace at the midpoint
2. Input with only a single character


seve:
> Can we add some tests to cover these edge cases please?

@tihuan 
Connecting this to #1578: Due to the limitation of pbmck I would only be able to test the whitespace case via the normal e2e tests.  Will probably be adding both the whitespace and single-char cases to the e2e_annotations tests

PR #1660, opened 2020-07-22
correctly handle non-string categoricals

bkmartinjr:
Categoricals which are not a string primitive may not rely on Dataframe's weak attempt to guess which type of summarization is required.  Change `summarize()` to `summarizeCategorical()` to request an explicitly categorical summarization.


bkmartinjr:
> Have you tried running user-annos on a dataset with non-string categoricals? I think `src/actions/annotation.js` makes the same summarization type guesses.

Filed as #1661 - suggest we resolve that in a separate PR, as it won't affect the current deployment which has user annotation disabled.

PR #1664, opened 2020-07-23
Fix the app config for reembedding and add a test

bmccandless:


bmccandless:
> LGTM. One question: why is scanpy required? It doesn't show up in the diff. Is this required under the hood for re-embedding?

Yes, the re-embedding uses scanpy for computing the new umap (server/compute/scanpy.py).  If you use the re-embedding feature you need to have it installed:  raise NotImplementedError("Please install scanpy to enable UMAP re-embedding")

bkmartinjr:
> LGTM. One question: why is scanpy required? It doesn't show up in the diff. Is this required under the hood for re-embedding?

Yes, has always been necessary.  But it was always an optional installation, which we did not want to force users to install because reembedding is still experimental.  So the server code would handle the case where it was not yet installed, and just issue a warning if the reembedding route was invoked.

I don't have an opinion if should be in requirements-dev.txt, although I intentionally left it out in the original work (my reason was that I didn't want any other dependencies creeping into the code base).

bmccandless:
> > LGTM. One question: why is scanpy required? It doesn't show up in the diff. Is this required under the hood for re-embedding?
> 
> Yes, has always been necessary. But it was always an optional installation, which we did not want to force users to install because reembedding is still experimental. So the server code would handle the case where it was not yet installed, and just issue a warning if the reembedding route was invoked.
> 
> I don't have an opinion if should be in requirements-dev.txt, although I intentionally left it out in the original work (my reason was that I didn't want any other dependencies creeping into the code base).

I added it for the test case I wrote.   It has to get installed to run the tests now.

bkmartinjr:
> I added it for the test case I wrote. It has to get installed to run the tests now.

Gotcha!  Is there a test to make sure things (other than re-embedding) run without it?  That will be the common case for users.

bmccandless:
> > I added it for the test case I wrote. It has to get installed to run the tests now.
> 
> Gotcha! Is there a test to make sure things (other than re-embedding) run without it? That will be the common case for users.

To run with and without scanpy installed?   That would be a little hard to test.  I think we could do it by creating a virtualenv with different modules installed for our tests to run.   That might be a bit more work than it's worth though in my opinion, but I don't have a better idea.  The other option is to just be careful not to import scanpy anywhere else in the code.

bkmartinjr:
> To run with and without scanpy installed? 

I was mostly interested in making sure we don't accidentally introduce dependencies not listed in requirements.txt.  As long as we run at least smoke tests with just requirements.txt (ie, what non-developer users install), it seems fine.

bmccandless:
> I was mostly interested in making sure we don't accidentally introduce dependencies not listed in requirements.txt. As long as we run at least smoke tests with just requirements.txt (ie, what non-developer users install), it seems fine.

Yes, it looks like we use just use requirements.txt in the push_tests.yml which run the smoke tests.  So, we should catch any accidental dependencies.

PR #1670, opened 2020-07-24
Add basic authentication in the server

bmccandless:
A pattern for creating authentication methods is introduced, with three
authentication types defined:
  none - no authentication
  session - like the current session based auth used for user annotations
  test - used to test the login/logout process end to end

The config endpoint now returns informations about the authentication, like if
the user is authenticated and their username.  The redirect uri's for login and
logout are also returned if the authentication type requires login

This is the first a several PRs for authentication.

### DIFF
--- a/server/app/app.py
+++ b/server/app/app.py
@@ -85,6 +85,7 @@ def dataset_index(url_dataroot=None, dataset=None):
     try:
         cache_manager = current_app.matrix_data_cache_manager
         with cache_manager.data_adaptor(url_dataroot, location, app_config) as data_adaptor:
+            data_adaptor.set_uri_path(f"{url_dataroot}/{dataset}")
             dataset_title = app_config.get_title(data_adaptor)
             return render_template(
                 "index.html", datasetTitle=dataset_title, SCRIPTS=scripts, INLINE_SCRIPTS=inline_scripts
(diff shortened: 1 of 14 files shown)


server/app/app.py
```diff
@@ -160,6 +162,17 @@ def dataroot_test_index():
 
     config = current_app.app_config
     server_config = config.server_config
+
+    auth = server_config.auth
+    if auth.is_valid():
+        if server_config.auth.is_authenticated():
```
MDunitz, line 168:
```suggestion
        if auth.is_authenticated():
```

server/app/app.py
```diff
@@ -160,6 +162,17 @@ def dataroot_test_index():
...
+            data += f"<p>Logged in as {auth.get_userid()} / {auth.get_username()}</p>"
+        if auth.requires_client_login():
+            if server_config.auth.is_authenticated():
```
MDunitz, line 171:
```suggestion
            if auth.is_authenticated():
```

server/common/app_config.py
```diff
@@ -414,8 +432,12 @@ def __init__(self, app_config, default_config):
         # The matrix data cache manager is created during the complete_config and stored here.
         self.matrix_data_cache_manager = None
 
+        # The authentication object (BCM -- better name)
```
MDunitz, line 435: could you write this as ```# TODO @brian better name```

bmccandless: Oops, didn't mean to commit the "BCM".   I'll fix this and the other issue in my next PR.

Issue #1509, opened 2020-06-04
Corpora schema support


bkmartinjr:
add support for importing data which conforms to the corpora schema.  Includes:
- [x] CXG encoding spec, as an accompanying doc to the corpora schema and corpora H5AD encoding specs
- [x] Update CXG spec to indicate how all Corpora fields are encoded
- [x] Update the CXG tool to follow this new encoding spec
- [x] Update cellxgene to know about this new version of CXG
- [x] Tests, as required

PR #1676, opened 2020-07-25
initial support for corpora schema conventions

### LINKED ISSUE #1532: corpora dataset project info populated into CXG for display in project/dataset card

bkmartinjr:
This PR adds:
* specification for CXG, including CXG encoding of Corpora schema conventions
* add support to extract Corpora dataset props, and serve them to the front-end via /config route
* cxgtool.py - add support to extract and encode Corpora dataset props in CXG

This is a partial solution to #1509 
Fixes #1532

### DIFF
--- a/dev_docs/cxg.md
+++ b/dev_docs/cxg.md
@@ -1,0 +1,162 @@
+# CXG Data Format Specification
+
+Document Status: _draft_
+
+Version: 0.2.0 (_DRAFT, not yet approved_)
+
+Date Last Modified: 2020-07-23
+
+## Introduction
+
+CXG is a cellxgene-private data format, used for at-rest storage of annotated matrix data. It is similar to [AnnData](https://anndata.readthedocs.io/en/stable/), but with performance and access characteristics amenable to a multi-dataset, multi-user serving environment.
+
+CXG is built upon the [TileDB](https://tiledb.com/) embedded database. Each CXG is a TileDB [group](https://docs.tiledb.com/main/api-usage/object-management), which in turn includes one or more TileDB multi-dimensional arrays.
+
+This document presumes familiarty with [TileDB terminology and concepts](https://docs.tiledb.com/main/), the [Corpora schema](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema.md) and its [H5AD encoding](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md), and the AnnData/H5AD data model.
+
+This document also leverages the current cellxgene schema, which is documented in the [REST API spec](./REST_API.md).
+
+### Terminology
+
+Unless explicitly noted, the AnnData conventions and terminology are adopted when referring to general annotated matrix characteristics (eg, `n_obs` is the number of observations/rows/cells in the annotated matrix).  Where implied by context, eg, "TileDB array attribute", domain-specific terms are used.
+
+Where capitalized, [IETF RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) conventions are followed (ie, conventions MUST be followed).
+
+_Author's note:_ if you see any ambiguous terms, please call them out for clarification.
+
+### Reserved
+
+The `cxg` prefix is used for CXG-specific names.
+
+### Encoding Data With TileDB Arrays
+
+The TileDB array schema authoritatively defines the characteristics of each array (eg, the type of `X` is defined by [`X.schema`](https://tiledb-inc-tiledb-py.readthedocs-hosted.com/en/stable/python-api.html#tiledb.libtiledb.Array.schema)). In some cases, additional metadata is required for the CXG, and is attched to the array using the TileDB [array metadata](https://docs.tiledb.com/main/basic-concepts/array-metadata) capability.
+
+All TileDB arrays MUST have a uint32 domain, zero based.  All X counts and embedding coordinates SHOULD be coerced to float32, which is ample precision for visualization purposes, and MUST be a numeric type. Dataframe (metadata) types are generally preserved, or where that is not possible, converted to something with equal representative value in the cellxgene application (eg, categorical types are converted to string, bools to uint8, etc).
+
+CXG consumers (readers) MUST be prepared to handle any legal TileDB compression, global layout and tile size.  CXG writers SHOULD attempt to encode data using best-effort heuristics for time and space considerations (eg, dense/sparse encoding tradeoffs).
+
+## Entities
+
+### CXG
+
+The CXG is a TileDB group containing all data and metadata for a single annotated matrix.  The following objects must be present in a CXG:
+* __obs__: a TileDB array, of shape (n_obs,), containing obs annotations, each annotation stored in a separate TileDB array attribute.
+* __var__: a TileDB array, of shape (n_var,), containing var annotations, each annotation stored in a separate TileDB array attribute.
+* __X__: a TileDB array, of shape (n_obs, n_var), with a single TileDB attribute of numeric type.
+* __X_col_shift__: (optional) TilebDB Array used in column shift encoding, shape (n_var,), dtype = X.dtype. Single unnamed numeric attribute.  
+* __emb__: a TileDB group, which in turn contains all (zero or more) embeddings.
+* __emb__/\<embedding_name\>__: a TileDB array, with a single anonymous attribute, of numeric type, and shape (n_obs, N>=2).
+* __cxg_group_metadata__: an empty TileDB array, used to store CXG-wide metadata
+
+### obs and var
+
+All per-observation (obs) and per-feature (var) data is encoded in a TileDB array named `obs` and `var` respectively, with shape (n_obs,) and (n_var,).  Each TileDB array has an array attribute for each obs/var column.  All TileDB array attributes will have the same type and value as the original data, eg, float32, with the following exceptions:
+* bool is encoded as uint8 (1/0)
+* categorical is encoded as string
+* Numeric types are cast to 32-bit equivalents
+
+In addition to the obs/var data, both TileDB arrays contain an optional 'cxg_schema' metadata field that is a JSON string containing per-column (attribute) schema hinting.  This is used where the TileDB native typing information is insufficient to reconstruct useful information such as categorical typing from Pandas DataFrames, and to communicate which column is the preferred human-readable index for obs & var.
+
+The `cxg_schema` JSON string is attached to the TileDB array metadata, and is a dictionary containing the following top-level names:
+* "index": string, containing the name of the index column
+* \<column-name\>: optional, a JSON dict, contain a schema definition using the same format as the cellxgene REST API /schema route
+
+For example:
+```
+{
+    "index": "obs_index",
+    "louvain": { "type": "categorical", "categories": [ "0", "1", "2", "3", "4" ]}
+    "is_useful": { "type": "boolean" }
+}
+```
+
+### X
+
+TileDB array, with a single anonymous attribute, shape (n_obs, n_var), containing the count matrix (equivalent to the AnnData `X` array).  MUST have numeric type, and SHOULD be float32.  The TileDB schema defines type and sparsity, and both dense and sparse encoding are supported.
+
+### X_col_shift
+
+Optional TileDB array, used to encode-per column offsets for column-shift sparse encoding.  The TileDB array will have a single anonymous attribute, of the same type as the X array, and shape (n_var,).  
+
+If the X array is sparse, and X_col_shift exists, then all values in the i'th column were subtracted by X_col_shift[i].
+
+### emb and embedding arrays
+
+A CXG must have a group named `emb`, which will contain all embeddings.  Embeddings are encoded as TileDB arrays, of numeric type and shape (n_obs, >=2).  The arrays SHOULD be coerced to float32, and MUST be a numeric type.  The TileDB array name will be assumed to be the embedding name (conventionally, embedding names in CXG are _not_ prefixed with an `X_` as they are in AnnData).
+
+CXG supports zero or more embeddings. Note that cellxgene currently _requires_ at least one embedding.
+
+### cxg_group_metadata
+
+Required, but empty TileDB array, used to store CXG-wide metadata.  The following fields are defined:
+* __cxg_version__: (required) a semver string identifying the specification version used to encode the CXG.
+* __cxg_properties__: (optional) a dictionary containing dataset wide properties, defined below.
+* __cxg_category_colors__: (optional) a categorical color table, defined below.
+
+#### cxg_properties
+
+The properties metadata dictionary contains dataset-wide properties, encoded as a JSON dictionary. Curently, the following fields are defined:
+* title: string, dataset human name (eg, "Lung Tissue")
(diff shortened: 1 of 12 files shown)


server/common/corpora.py
```diff
@@ -0,0 +1,81 @@
...
+        corpora_schema_version is not None
+        and corpora_encoding_version is not None
+        and corpora_schema_version.startswith("1.")
+        and corpora_encoding_version.startswith("0.1.")
+    )
+
+    if not version_is_supported:
+        raise ValueError("Unsupported Corpora schema version")
+
+    Required_Simple_Fields = [
```
maniarathi, line 55: Just a nit: why is the casing weird on these variables?

server/common/corpora.py
```diff
@@ -0,0 +1,81 @@
...
+    Get Corpora dataset properties from an AnnData
+    """
+    versions = corpora_get_versions_from_anndata(adata)
+    if versions is None:
+        return None
+    [corpora_schema_version, corpora_encoding_version] = versions
+    version_is_supported = (
+        corpora_schema_version is not None
+        and corpora_encoding_version is not None
+        and corpora_schema_version.startswith("1.")
```
maniarathi, line 48: Can we pull out the supported schema versions into a utility file or some sort of constants file. I see that `cxgtool.py` also checks the versions so it would be helpful to have the supported versions in a single location that can get updated instead of having to remember a couple places to check.

server/test/test_diffexp.py
```diff
@@ -105,13 +105,23 @@ def sparse_diffexp(self, apply_col_shift):
             adata = adaptor_anndata.data
 
             sparsename = os.path.join(dirname, "sparse.cxg")
-            write_cxg(adata=adata, container=sparsename, title="sparse", sparse_threshold=11)
+            cxg_group_metadata = create_cxg_group_metadata(
```
maniarathi, line 108: Do we also need a test case here where `corpora_props` is not None since that is new functionality?

dev_docs/cxg.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Entities
+
+### CXG
+
+The CXG is a TileDB group containing all data and metadata for a single annotated matrix.  The following objects must be present in a CXG:
+* __obs__: a TileDB array, of shape (n_obs,), containing obs annotations, each annotation stored in a separate TileDB array attribute.
+* __var__: a TileDB array, of shape (n_var,), containing var annotations, each annotation stored in a separate TileDB array attribute.
+* __X__: a TileDB array, of shape (n_obs, n_var), with a single TileDB attribute of numeric type.
+* __X_col_shift__: (optional) TilebDB Array used in column shift encoding, shape (n_var,), dtype = X.dtype. Single unnamed numeric attribute.
```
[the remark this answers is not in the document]
bkmartinjr: @bmccandless - please add this information to this branch.

server/common/corpora.py
bkmartinjr: casing?

server/common/corpora.py
bkmartinjr: Can you expand on what problem this solves?  I bias away from overly complicated config & code structure unless there is a clear value.   Right now, this file is entirely self-contained wrt semantics and readability.  Given how simple the code is, I'm not sure splitting helps.

LMK.

dev_docs/cxg.md
```diff
@@ -0,0 +1,162 @@
...
+        "version": {
+            "corpora_schema_version": "1.0.0",
+            "corpora_encoding_version": "0.1.0",
+        }
+    }
+}
+```
+
+The `corpora` metadata, if present, MUST contain the version information. Optionality of other values in this object will follow the specifications set forth in the relevant Corpora schema specification (ie, optional fields are optional, required are present, etc), with the following changes:
+* the contents of `corpora_encoding_version` MUST be identical to the `cxg_version`, as this field is defined as the current object encoding version, *NOT* the source data encoding version.
```
bmccandless, line 138: If corpora_encoding_version must be identical to cxg_version, then why have it?

server/test/test_diffexp.py
bkmartinjr: @bmccandless - can you respond.  I'm unclear why this parameter would be necessary to test for differential expression.

dev_docs/cxg.md
```diff
@@ -0,0 +1,162 @@
...
+
+All other Corpora schema fields will be encoded into a CXG using the conventions defined in the [Corpora Schema AnnData Implementation](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md). For example, fields in `AnnData.obs` will be encoded in the CXG `obs`array as defined [above](#obs-and-var).
+
+### Compatibility with CXG 0.1.0
+
+For backwards compatibility and continuity with CXG version 0.1.0, the following MUST be implemented.
+
+#### Presentation Hints
+* The [Corpora `title`](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema.md#presentation-metadata) value MUST be saved in the `cxg_properties.title` field.
+* The [Corpora `color_map`](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema.md#presentation-hints) MUST be saved in the `cxg_category_colors` field.
```
bmccandless, line 152: I think is is listed as optional in the corpora schema.  "Submitters can include a field called {field}_colors..."

server/common/corpora.py
```diff
@@ -0,0 +1,81 @@
...
+from server.cli.upgrade import validate_version_str
+
+
+def corpora_get_versions_from_anndata(adata):
+    """
+    Given an AnnData object, return:
+        * None - if not a Corpora object
+        * [ corpora_schema_version, corpora_encoding_version ] - if a Corpora object
+
+    Implements the identification protocol defined in the spect.
```
bmccandless, line 19: s/spect/spec/

server/common/corpora.py
maniarathi: I figured that because this particular check is written word-for-word twice (lines 551-556) in a completely different file `cxgtool.py`, it seems preferable to consolidate into a single location. Otherwise you run into the issue of having to remember to update in both locations or run into bugs.

server/common/corpora.py
maniarathi: The capitalization of the beginning of each word threw me off :P Haven't seen it before and didn't know what it meant.

bkmartinjr: lolz, no idea - just me trying to imply constants.  I'll remove it on the assumption you are the person who defines code convensions.

server/common/corpora.py
bkmartinjr: Ah, I see - you think both files will always have the same version check.  I was not assuming that.  But for now, I'll DRY it.

dev_docs/cxg.md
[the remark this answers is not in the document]
bmccandless: There is wording in the next section that explains each of these arrays in more detail.  If looks like this section is just meant to be a brief summary of the arrays.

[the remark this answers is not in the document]
bkmartinjr: correct - but I assume @maniarathi wants even more details than the following section?  Arathi?

dev_docs/cxg.md
bkmartinjr: Because cxg_version is a super set.   They evolve independently.

dev_docs/cxg.md
bkmartinjr: correct, but when it is present, it must be int that field, rather than in the corpora field.

I'll revise the wording to make it more obvious what I'm trying to say.

server/test/test_diffexp.py
bmccandless: This test is just to verify that the sparse matrix converter works.   
A test case for corpora_props should probably go in the test_cxgtool.py

dev_docs/cxg.md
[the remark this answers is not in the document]
bmccandless: @maniarathi, there is no link to col shift encoding.  It is our own invention to recapture the inherit sparseness of a matrix that had been previously normalized.   The cxgtool.py automatically can detect this and do the transformation.   A user would not need to know about it (but the details are pretty simple).

server/test/test_diffexp.py
bkmartinjr: write_cxg is not a public interface, and will never have a None value in props.  Matt and Brian pulled  this out simply for testing.

I will add a test case for the corpora.py exported API.

bkmartinjr: added test_corpora.py

server/common/corpora.py
```diff
@@ -0,0 +1,87 @@
...
+    ]
+    # Spec says some values encoded as JSON due to the inability of AnnData to store complex types.
+    required_json_fields = ["contributors", "project_links"]
+    optional_simple_fields = ["preprint_doi", "publication_doi", "default_embedding", "default_field", "tags"]
+
+    corpora_props = {}
+    for key in required_simple_fields:
+        if key not in adata.uns:
+            raise KeyError(f"missing Corpora schema field {key}")
+        corpora_props.update({key: adata.uns[key]})
```
mckinsel, line 78: This is a kind of unusual way to assign to a dict. Any reason not to `corpora_props[key] = adata.uns[key]`?

server/common/corpora.py
```diff
@@ -0,0 +1,87 @@
...
+    for key in required_json_fields:
+        if key not in adata.uns:
+            raise KeyError(f"missing Corpora schema field {key}")
+        corpora_props.update({key: json.loads(adata.uns[key])})
```
mckinsel, line 82: You might want to add a little color to a `json.JSONDecodeError` here, like `try: ... except ...: raise json.JSONDecodeError(f"Corpora schema field {key} is expected to be a valid JSON string")`

server/converters/cxgtool.py
```diff
@@ -601,6 +545,59 @@ def save_metadata(container, metadata_dict):
...
+        # clobber encoding version to be OUR version, not the source H5AD encoding
+        corpora_props["version"].update({"corpora_encoding_version": CXG_VERSION})
+        corpora_project_links = corpora_props.get("project_links", [])
+        corpora_about_link = next(
+            (link for link in corpora_project_links if (link.get("link_type", None) == "SUMMARY")), {}
+        )
+    else:
+        corpora_about_link = {}
+
+    title = title if title is not None else corpora_about_link.get("link_name", basefname)
```
mckinsel, line 579: This is kind of a personal preference, so take it or leave it, but

```python
title = title or corpora_about_link.get("link_name", basefname)
```

dev_docs/cxg.md
```diff
@@ -0,0 +1,162 @@
...
+All implementions MUST ignore unrecognized fields. 
+
+#### cxg_category_colors
+
+This optional field contains a copy of the category color table, which MAY be used to display category-specific color labels. This is a JSON dictionary, containing a per-category color-table.  Each color table is named `{category_name}_colors`, and is itself a dictionary mapping label name to RGB color.  For example:
+
+```
+{
+    "louvain_colors": {
+        "0": "#FFFFFF",
```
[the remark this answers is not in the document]
bkmartinjr: It currently supports several formats, but that is really an app convention (currently unspecified), and I'm not sure if it belongs in this spec or not.  At the moment, it will support either of:
* rgb(DDD,DDD,DDD) -- eg, `rgb(255,255,255)` (base 10)
* #HHHHHH -- eg, `#FF0000` (hex)

server/common/corpora.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Given an AnnData object, return:
+        * None - if not a Corpora object
+        * [ corpora_schema_version, corpora_encoding_version ] - if a Corpora object
+
+    Implements the identification protocol defined in the specification.
+    """
+
+    # per Corpora AnnData spec, this is a corpora file if the following is true
+    if adata.uns is None or "version" not in adata.uns:
```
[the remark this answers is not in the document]
bkmartinjr: per the [AnnData spec](https://anndata.readthedocs.io/en/stable/anndata.AnnData.html#anndata.AnnData), `None` is legal, and is the default value.

server/common/corpora.py
```diff
@@ -0,0 +1,87 @@
...
+        return None
+    version = adata.uns["version"]
+    if not isinstance(version, collections.abc.Mapping) or "corpora_schema_version" not in version:
+        return None
+
+    corpora_schema_version = version.get("corpora_schema_version", None)
+    corpora_encoding_version = version.get("corpora_encoding_version", None)
```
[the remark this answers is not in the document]
bkmartinjr: true, I was just being explicitly explicit :-).   Will remove.

server/common/corpora.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # TODO: spec says these must be SEMVER values, so check.
+    if validate_version_str(corpora_schema_version) and validate_version_str(corpora_encoding_version):
+        return [corpora_schema_version, corpora_encoding_version]
+
+    return None
```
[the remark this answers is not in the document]
bkmartinjr: Can you clarify?   Do you mean returning None has no side-effects?  Or?

server/common/corpora.py
bkmartinjr: I simply lack any style sensibility :-)  Will fix.

dev_docs/cxg.md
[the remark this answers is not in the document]
mckinsel: Okay thanks, was mostly just curious.

server/common/corpora.py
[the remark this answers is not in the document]
mckinsel: I think `None` is the default parameter value in the constructor, but I'm pretty sure it [always sticks something in there](https://github.com/theislab/anndata/blob/master/anndata/_core/anndata.py#L506).

server/common/corpora.py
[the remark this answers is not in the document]
mckinsel: If you delete that line the behavior of the function does not change:
```python
>>> def fn():
        if False: return "something"
>>> fn() is None
True
```
Not a big deal, if you think it's clearer with the explicit `return`.

server/common/corpora.py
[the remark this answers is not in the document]
bkmartinjr: I have previously been bit by not following their documentation (as they feel free to change the code on occasion).  Unless you see a major advantage, I'd prefer to just assume their docs are correct, even if it costs us an extra check....

server/common/corpora.py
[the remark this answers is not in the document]
bkmartinjr: doh...got it.

server/common/corpora.py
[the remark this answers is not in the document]
mckinsel: Extremely possible I'm missing something, but I don't think the docs say that `adata.uns` may be `None`? The docs don't appear to express a super crisp opinion about the question though. Maybe the defensive thing to do is `if "version" not in adata.uns_keys()`?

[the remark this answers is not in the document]
bkmartinjr: The `uns` field definition on that page is:  `uns : Optional[Mapping[str, Any]] (default: None)`

This is confusing, as `None` is not a Mapping.  Hence my conservatism.

So yeah, your approach seems good.

PR #1679, opened 2020-07-28
re-implement re-embeddings

### LINKED ISSUE #1569: Cannot select all cells after re-embedding and labeling a subset in a new category
Related to https://github.com/chanzuckerberg/cellxgene/issues/280

cellxgene version:
`(cellxgene) DNab4216c1:scanpy ktrav$ cellxgene --version`
`[cellxgene] Version 0.15.0`

Launch command used:
`cellxgene launch --experimental-enable-reembedding droplet-infection_1-adata_stroma-cxg.2020-06-15.h5ad`

Steps to reproduce:
1. Select a group of cells, subset, and re-embed
2. Change to new embedding
3. Select another group of cells
4. Create a new category
5. Add label to the category and select middle "Add X number of selected cells to this label"

After following this sequence, I am unable to select any cells not in the group I just labeled. I do not believe this behavior is intended because this does not occur if you select and label cells using the original embedding.

Undoing the subset (clicking the whole circle icon) allows selection again, but is annoying if labeling multiple groups on the new embedding.

### LINKED ISSUE #1606: rewrite reembedding to use annomatrix
Reembedding is currently disabled during the refactor.  Reimplement. Likely requires also resolving #1559.

bkmartinjr:
this re-enables _experimental_ re-embedding support, which was temporarily disabled by the annoMatrix refactoring.

There are two groups of changes, which build on top of the previous work supporting subset embeddings:
* The backend route is slightly changed.  A PUT to /layout/obs will now compute an embedding on the requested set of cells, and store that embedding in anndata.obsm.  All cells which are not in that embedding will be noted with a NaN value.  The route now returns _just_ the new schema for that layout column.  The front-end will now be required to separately perform a GET to fetch the data.
* The front-end has resurrected the menubar UI, and has been adapted to the new route. 

This change has one significant end-user effect:  a refresh of the front-end will not lose the embedding (yay!).   The embedding will persist as long as the back-end is running.

This will almost certainly need to be further refined for any eventual multi-user implementation.

Fixes #1606 
Fixes #1569


server/test/test_anndata_adaptor.py
```diff
@@ -237,14 +237,12 @@ def test_compute_embedding(self):
...
         name = schema["name"]
         self.assertEqual(schema["type"], "float32")
         self.assertEqual(schema["dims"], [f"{name}_0", f"{name}_1"])
 
-        emb = decode_fbs.decode_matrix_FBS(fbs)
-        self.assertEqual(emb["n_rows"], 100)
-        self.assertEqual(emb["n_cols"], 2)
-        self.assertEqual(emb["col_idx"], [f"{name}_0", f"{name}_1"])
+        emb = self.data.data.obsm[f"X_{name}"]
+        self.assertEqual(emb.shape, (2638, 2))
```
maniarathi, line 248: Wouldn't you also want to assert that the re-embedding is only available for the 100 obs you subsetted and the rest are NaNs?

server/test/test_api.py
```diff
@@ -76,21 +76,23 @@ def test_put_layout_fbs(self):
...
         self.assertEqual(result.status_code, HTTPStatus.OK)
-        df = decode_fbs.decode_matrix_FBS(result.content)
-        self.assertEqual(df["n_rows"], 100)
-        self.assertEqual(df["n_cols"], 2)
-        cols = list(df["col_idx"])
-        self.assertTrue(cols[0].startswith("reembed:umap_") and cols[0].endswith("_0"))
-        self.assertTrue(cols[1].startswith("reembed:umap_") and cols[1].endswith("_1"))
+        result_data = result.json()
+        self.assertIsInstance(result_data, dict)
+        self.assertEqual(result_data.get("type", None), "float32")
```
maniarathi, line 89: Are the defaults for the `get` functions there on purpose? It seems like you never actually want the default value to occur since you assert for equality so the `get` function failing is just wrong behavior right? In which case, I'd remove the default function to assert that result_data.get("type") is always expected to return `float32`. Ditto in lines below.

Also.. the default `default` value is `None` anyways which makes it redundant :) 

Ditto in other locations.

client/src/annoMatrix/loader.js
```diff
@@ -194,6 +195,20 @@ export default class AnnoMatrixLoader extends AnnoMatrix {
...
+  addEmbedding(colSchema) {
+    /*
+    add new layout to the obs embeddings
+    */
+    const { name: colName } = colSchema;
+    if (_getColumnSchema(this.schema, "emb", colName)) {
+      throw new Error("column already exists");
+    }
+
+    const o = this._clone();
```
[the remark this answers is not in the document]
bkmartinjr: Sure, will do.  But this is a case where I deliberately choose the name, as anything longer is not informative, and makes the code harder to read.  But I will change.

server/test/test_api.py
bkmartinjr: I'm just not very pythonic and was doing a lot of cutting/pasting.  Will cleanup!

server/test/test_anndata_adaptor.py
bkmartinjr: good idea

server/data_anndata/anndata_adaptor.py
```diff
@@ -310,16 +309,15 @@ def compute_embedding(self, method, obsFilter):
             raise FilterError("Error parsing filter")
         with ServerTiming.time("layout.compute"):
             X_umap = scanpy_umap(self.data, obs_mask)
```
bmccandless, line 311: You are no longer normalizing the embedding here.  Is it no longer needed?

bkmartinjr: correct, not needed.  That happens when the client requests the embedding via the normal GET /layout/obs route (just like all other embeddings).

bmccandless: sounds good.

PR #1681, opened 2020-07-28
add oauth authentication

bmccandless:
Add support for OAuth2.

Change the interface to AuthTypeBase
  - better handling of config parameters
  - add a complete_setup function for additional setup steps

Added a function wrapper to enforce authentication for the
routes that require authenticaiton.


server/auth/auth.py
```diff
@@ -21,6 +16,11 @@ def requires_client_login(self):
         """Return True if the user needs to login from the client (e.g. Login button is shown)"""
         return False
 
+    @abstractmethod
+    def complete_setup(self, app):
```
maniarathi, line 20: Could there be a better name for this function or at least a better docstring? I have very little understanding of what is supposed to happen in this function if I were to use this interface. What are the things that are necessary to "complete setup"?

server/auth/auth_oauth.py
```diff
@@ -0,0 +1,176 @@
+from flask import session, request, redirect, current_app
+from server.auth.auth import AuthTypeClientBase, AuthTypeFactory
+from server.common.errors import AuthenticationError, ConfigurationError
+from urllib.parse import urlencode
+from urllib.request import urlopen
+import json
+
+# It is not required to have authlib or jose.
+# However, it is a configuration error to use this auth type if they are not installed.
+missingimport = []
```
maniarathi, line 10: This is a global variable and therefore should be capitalized to MISSING_IMPORTS or something similar right?

https://www.python.org/dev/peps/pep-0008/#constants

server/auth/auth.py
```diff
@@ -7,11 +7,6 @@ class AuthTypeBase(ABC):
     def __init__(self):
         super().__init__()
 
-    @abstractmethod
-    def set_params(self, params):
-        """Set the parameters from app config.  raise ConfigurationError if any params are invalid"""
-        pass
-
     @abstractmethod
     def is_valid(self):
```
maniarathi, line 11: Why not make this more explicitly named like `can_return_user_info` or something like that? `is_valid` has so many connotations in terms of auth in general so I had to search around to understand exactly what `is_valid` meant in the implementations.

server/auth/auth_oauth.py
```diff
@@ -0,0 +1,176 @@
...
+        if missingimport:
+            raise ConfigurationError(f"oauth requires these modules: {', '.join(missingimport)}")
+        self.algorithms = ["RS256"]
+        self.api_base_url = app_config.authentication__params_oauth__api_base_url
+        self.client_id = app_config.authentication__params_oauth__client_id
+        self.client_secret = app_config.authentication__params_oauth__client_secret
+        self.callback_base_url = app_config.authentication__params_oauth__callback_base_url
+        self.audience = self.client_id
+
+        # load the jwks
```
maniarathi, line 38: What is "jwks"?

server/auth/auth_oauth.py
```diff
@@ -0,0 +1,176 @@
...
+    def add_url_rules(self, app):
+        app.add_url_rule("/login", "login", self.login, methods=["GET"])
+        app.add_url_rule("/logout", "logout", self.logout, methods=["GET"])
+        app.add_url_rule("/oauth2/callback", "callback", self.callback, methods=["GET"])
+
+    def complete_setup(self, flask_app):
+        self.oauth = OAuth(flask_app)
+        if self.callback_base_url is None:
+            # assume this is for a local test, and the oauth provider has been configured
+            # with a callback that understands a localhost callback (e.g. A http://localhost:5005).
```
maniarathi, line 61: Can this not be patched in the test itself instead of having a code related to testing in the production code?

server/auth/auth_oauth.py
```diff
@@ -0,0 +1,176 @@
...
+        return_path = request.args.get("dataset", "")
+        return_to = f"{self.callback_base_url}/{return_path}"
+        params = {'returnTo' : return_to, 'client_id' : self.client_id}
+        return redirect(self.client.api_base_url + '/v2/logout?' + urlencode(params))
+
+    def callback(self):
+        token = self.client.authorize_access_token()
+        id_token = token.get("id_token")
+        session[self.CXG_ID_TOKEN] = id_token
+        r = session.get("oauth_callback_redirect", "/")
```
maniarathi, line 116: Avoid single letter variable names which aren't readable.

server/auth/auth.py
bmccandless: I added to the docstring.  This is just a hook that gets called before the server starts running.   It's entirely up to the authentication type to do something that makes sense here.   The oauth creates a client to the authentication service here, and it's a no-op for the rest of the classes.

server/auth/auth_oauth.py
bmccandless: It a global, but not a constant.

server/auth/auth.py
bmccandless: I don't quite like "can_return_user_info", since that might imply if can return other things.  I changed it to "is_valid_authentication_type".   Is that better?

server/auth/auth_oauth.py
bmccandless: Added the definition

server/auth/auth_oauth.py
bmccandless: I change the comment to remove the part about "for a local test".     It's not for a specific test, but anytime you run the client and server on the same host.

server/auth/auth_oauth.py
bmccandless: Okay, renamed it.

server/auth/auth.py
bmccandless: I also changed the names of a few other functions to make it more clear.
The ones with "user" in the name refer to the current user (or lack of current user): 
  is_user_authenticated, get_user_name, get_user_id, get_user_email.

PR #1683, opened 2020-07-29
Fix server exception classes

### LINKED ISSUE #1675: Missing CLI error message on bad dataset path
If I launch with a bad dataset path, the error message is less than useful.

```
(venv) bruce@neigh:~/projects/cellxgene$ $ cellxgene launch --debug /a/bad/path/10x_mouse.cxg/
[cellxgene] in debug mode, setting verbose=True and open_browser=False
Error: 
(venv) bruce@neigh:~/projects/cellxgene$ 
```

Ideally, we would provide plain english error messages for most common errors.

bmccandless:
str(e) and e.message will both show the error message.
refactored the error.py file to simplify our exception class definitionso

Fix #1675

### DIFF
--- a/server/common/errors.py
+++ b/server/common/errors.py
@@ -1,85 +1,47 @@
 from http import HTTPStatus
 
 
-class RequestException(Exception):
+class CellxgeneException(Exception):
+    """Base class for cellxgene exceptions"""
+
+    def __init__(self, message):
+        self.message = message
+        super().__init__(message)
+
+
+class RequestException(CellxgeneException):
     """Baseclass for exceptions that can be raised from a request."""
 
     # The default status code is 400 (Bad Request)
     default_status_code = HTTPStatus.BAD_REQUEST
 
     def __init__(self, message, status_code=None):
-        Exception.__init__(self)
-        self.message = message
+        super().__init__(message)
         self.status_code = status_code or self.default_status_code
 
 
-class FilterError(RequestException):
-    """Raised when filter is malformed"""
-
-    pass
+def define_exception(name, doc):
+    globals()[name] = type(name, (CellxgeneException,), dict(__doc__=doc))
 
 
-class JSONEncodingValueError(RequestException):
-    """Raised when data cannot be encoded into json"""
-
-    pass
+def define_request_exception(name, doc, default_status_code=HTTPStatus.BAD_REQUEST):
+    globals()[name] = type(name, (RequestException,), dict(__doc__=doc, default_status_code=default_status_code))
 
 
-class MimeTypeError(RequestException):
-    """Raised when incompatible MIME type selected"""
+define_request_exception("FilterError", "Raised when filter is malformed")
+define_request_exception("JSONEncodingValueError", "Raised when data cannot be encoded into json")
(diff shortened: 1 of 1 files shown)


server/common/errors.py
```diff
@@ -1,85 +1,47 @@
...
-    pass
-
+def define_exception(name, doc):
+    globals()[name] = type(name, (CellxgeneException,), dict(__doc__=doc))
 
-class ConfigurationError(Exception):
-    """Raised when checking configuration errors"""
 
-    pass
+def define_request_exception(name, doc, default_status_code=HTTPStatus.BAD_REQUEST):
```
maniarathi, line 27: A bit confused; why not make all of the below exceptions instantiations of `CellxgeneException` where you simply edit the message that is outputted? Do you get any benefits from defining each one separately?

bmccandless: Two benefits.  
1 RequestException is different from CellxgeneException because it has the status_code, which becomes part of the response to the client.
2 The reason for having multiple exception classes instead of one is to give a little bit more control over how you might handle the exception.     There are (a small number of) places that make use of that flexibility.

We could simply our exception handling by just having two classes (RequestException and CellxgeneException), but that would lose some self documenting and the flexibility to handle different types of exceptions differently.  Also, that would require changing about 20 files.
  
In this PR I was fixing the str(e) bug, but while I was here I decided to do some refactoring  since there was a lot of boilerplate in this file.

maniarathi: Ahh ok, thanks for the explanation! 

Yep, I think it makes sense to have the two classes CellxgeneException and RequestException, but the others didn't make sense to me since they aren't handled any differently and the message details exactly what the error is. Creating a new class with documentation doesn't necessarily provide any additional information does it? 

I was thinking that the refactoring here actually obfuscates the definition of the classes and for some of the classes it changes the parent exception from RequestException to CellxgeneException which would be incorrect (e.g. PrepareError, ConfigurationError, and OntologyLoadFailure). 

Are you heavily opposed to editing the 20 files to keep the exceptions clean? :P

bmccandless: Four points:

1. The point of having different exception classes is so you can do something like this:

```
try:
   do_something()
except DisabledFeatureError as e:
  # handle this error one way
except JSONEncodingValueError as e:
  # handle it another way
```

2.  This PR wasn't intended to change the public interface of the exceptions.   Basically I needed to fix how str(e) for the bug I'm fixing, and thought the refactor made the file simpler.
If you don't like my refactoring to generate the exception types, that's something we can talk about for this PR.  
If you want to change the exception handling interface then I'd say we need a good reason to (if it ain't broke...).

3.  I meant to change the parent class of PrepareError, ConfigurationError and OntologyLoadFailure.

4.  By self documenting, I meant not in the errors.py file, but when you raise and catch the exception:
e.g.

```
# current:  This tells you more as a reader, so it's a little more self documenting
raise ConfigurationError(f"unknown config parameter {key}.")
# your proposal
raise CellxgeneError(f"unknown config parameter {key}.")
```

maniarathi: Ahh got it! Thank you so much for the detailed explanation. Makes a lot of sense to me, especially point 1.

PR #1693, opened 2020-07-29
check for bins change for canvas draw

### LINKED ISSUE #1610: Mini histos in left sidebar no longer update on subset

seve:
Added a bins change condition where canvas should be re-rendered

---
fixes #1610

### DIFF
--- a/client/src/components/categorical/value/index.js
+++ b/client/src/components/categorical/value/index.js
@@ -453,10 +453,9 @@ class CategoryValue extends React.Component {
         label,
         CHART_WIDTH,
         VALUE_HEIGHT
-      ) ?? {};
+      ) ?? {}; // if createHistogramBins returns empty object assign null to deconstructed
 
-    if (xScale === undefined || yScale === undefined || bins === undefined)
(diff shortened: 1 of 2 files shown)


client/src/components/categorical/value/index.js
```diff
@@ -455,6 +455,9 @@ class CategoryValue extends React.Component {
         VALUE_HEIGHT
       ) ?? {};
 
+    if (xScale === undefined || yScale === undefined || bins === undefined)
```
colinmegill, line 458: Does it have to be `undefined` — why not: `if (!xScale || !yScale || !bins)`?

seve: It could, I just chose to be a bit more verbose for clarity.  Especially since the assignment is a little nuanced.:
```js
const { xScale, yScale, bins } =
      this.createHistogramBins(
        metadataField,
        categoryData,
        colorAccessor,
        colorData,
        label,
        CHART_WIDTH,
        VALUE_HEIGHT
      ) ?? {};
```

PR #1696, opened 2020-07-29
add support for corpora default_embedding field

### LINKED ISSUE #1570: add support for primary embedding
Add feature to indicate which embedding in a dataset is primary, and should be displayed upon launch.  Wire into CXG and corpora schema.

bkmartinjr:
Fixes #1570 

When a dataset includes a Corpora "default_embedding", that embedding will be displayed upon data load.  If not defined, the preexisting heuristic will choose initial embedding to display.


server/common/app_config.py
```diff
@@ -242,11 +242,17 @@ def get_client_config(self, data_adaptor):
...
+        # corpora dataset_props
         # TODO/Note: putting info from the dataset into the /config is not ideal.
         # However, it is definitely not part of /schema, and we do not have a top-level
         # route for data properties.  Consider creating one at some point.
         corpora_props = data_adaptor.get_corpora_props()
+        if corpora_props and "default_embedding" in corpora_props:
+            default_embedding = corpora_props["default_embedding"]
+            if isinstance(default_embedding, str) and default_embedding.startswith("X_"):
+                default_embedding = default_embedding[2:]  # drop X_ prefix
+            if default_embedding in data_adaptor.get_embedding_names():
```
maniarathi, line 254: Need a test to verify that `default_embedding` is or is not based on the logic you've outlined with the three conditionals?

bkmartinjr: not entirely sure I understand the question...  But I believe that the tests you are asking about are already done inside `get_embedding_names()`.  It returns a list of the embeddings which are legit - ie, which exist, meet our criteria for serving, etc.

maniarathi: Sorry, all I meant was to request modifying `test_app_config.py` to verify that `data_config["config"]["parameters"]["default_embedding"]` gets populated in one of the tests which I don't see today.

bkmartinjr: roger - will do.

bkmartinjr: see test_corpora.py

server/test/test_corpora.py
```diff
@@ -71,3 +77,74 @@ def test_corpora_get_props_from_anndata(self):
...
+        corpora_props = result_data["config"]["corpora_props"]
+        parameters = result_data["config"]["parameters"]
+
+        self.assertEqual(corpora_props["version"]["corpora_schema_version"], "1.0.0")
+        self.assertEqual(corpora_props["organism"], "human")
+        self.assertEqual(parameters["default_embedding"], "tsne")
+
+        import sys
+        print(corpora_props, file=sys.stderr)
+        print(parameters, file=sys.stderr)
```
maniarathi, line 150: Are these three debugging lines or there on purpose?

bkmartinjr: nope - detritus. good catch!

PR #1713, opened 2020-08-04
Update hosted app to get the oauth client secret from the secret manager

bmccandless:


server/eb/app.py
```diff
@@ -40,14 +40,54 @@ def get_secret_key(region_name, secret_name, secret_key):
...
+
+    # need to find the secret manager region.
+    #  1. from CXG_AWS_SECRET_REGION_NAME
+    #  2. discover from dataroot location (if on s3)
+    #  3. discover from config file location (if on s3)
+    secret_region_name = os.getenv("CXG_AWS_SECRET_REGION_NAME")
+    if secret_region_name is None:
+        secret_region_name = discover_s3_region_name(app_config.multi_dataset__dataroot)
+        if not secret_region_name:
+            secret_region_name = discover_s3_region_name(config_file)
```
MDunitz, line 65: could you set it as an env var so its available elsewhere in the app after this runs once?

bmccandless: The secret_region_name is only used here to access the secrets manager

Issue #1524, opened 2020-06-04
authenticated user sessions


bkmartinjr:
1. hosted cellxgene can be configured to use both authenticated and ad hoc user sessions
2. Anonymous sessions continue to be supported on desktop version
3. Corpora deployment integrated with Corpora auth
4. WSGI version demo/docs on how to use with plain OAUTH
5. UI support, eg, when authn challenge required, redirect to login flow

PR #1728, opened 2020-08-10
Separate userinfo from the config endpoint

### LINKED ISSUE #1519: hosted cellxgene can be configured to use both authenticated and ad hoc user sessions

bmccandless:
previously information about if the user was logged in and their username
was part of the config endpoint.
However, the config endpoint was previously static, and has a cache control.
Rather than not caching the config, a new endpoint called "userinfo"
is created to handle that information.

The config endpoint still has the non-changing part of the authentication:

```
   config:
     authentication:
         requires_client_login:  True/False
         login: <uri to login endoint if requires_client_login is True>
         logout: <uri to logout endoint if requires_client_login is True>
```

The userinfo endpoint returns this information:

```
  userinfo:
     is_authenticated:  True/False
     username: <string if is_authenticated>
 
```

if authentication is not enabled then the config does not have an authentication key,
and userinfo returns None.

Also in the PR are a few minor code improvements and bug fixes

 #1519 
 #1523
 #1756


server/auth/auth_oauth.py
```diff
@@ -227,11 +229,12 @@ def get_jwt_payload(self):
...
 
-            except jwt.JWTError as e:
+            except JWTError as e:
                 raise AuthenticationError(f"invalid signature: {str(e)}")
-            except jwt.ExpiredSignatureError as e:
-                raise AuthenticationError(f"token expired: {str(e)}")
-            except jwt.JWTClaimsError as e:
+            except ExpiredSignatureError:
+                # TODO, handle expired sessions by refreshing the token
+                return None
```
MDunitz, line 236: Why none instead of an AuthError?

bmccandless: It's not an error for the token to be expired, it just means the user is no longer logged in.  
It is an error if the token is malformed, or does not have the right claims, audience, etc.

I have a PR in progress that will attempt to refresh the token in this case.

server/test/unit/auth/test_auth.py
```diff
@@ -79,27 +82,27 @@ def test_auth_test(self):
...
             assert r.url == f"{server}/auth/pbmc3k.cxg/"
-            r = session.get(f"{server}/auth/pbmc3k.cxg/api/v0.2/config")
-            data_config = r.json()
-            assert not data_config["config"]["authentication"]["is_authenticated"]
-            assert data_config["config"]["authentication"]["username"] is None
-            assert data_config["config"]["parameters"]["annotations"]
+            config = session.get(f"{server}/auth/pbmc3k.cxg/api/v0.2/config").json()
+            userinfo = session.get(f"{server}/auth/pbmc3k.cxg/api/v0.2/userinfo").json()
+            assert not userinfo["userinfo"]["is_authenticated"]
+            assert userinfo["userinfo"]["username"] is None
```
MDunitz, line 98: could you swap the assert statements for self.assertEqual? its more readable and that seems to be the pattern in the rest of the codebase

client/src/components/menubar/authButtons.js
```diff
@@ -19,10 +19,10 @@ const Auth = React.memo((props) => {
           type="button"
           data-testid="auth-button"
           disabled={false}
-          icon={!auth.is_authenticated ? "log-in" : "log-out"}
-          href={!auth.is_authenticated ? auth.login : auth.logout}
+          icon={!userinfo.is_authenticated ? "log-in" : "log-out"}
```
seve, line 22:
```suggestion
          icon={!userinfo["is_authenticated"] ? "log-in" : "log-out"}
```

We have the convention of subscripting non-camel case attributes

client/src/components/menubar/index.js
```diff
@@ -246,7 +248,7 @@ class MenuBar extends React.PureComponent {
           zIndex: 3,
         }}
       >
-        <AuthButtons auth={auth} />
+        <AuthButtons auth={auth} userinfo={userinfo} />
```
seve, line 251: If you want to, you can do:
```suggestion
        <AuthButtons {...{auth, userinfo}} />
```

client/src/reducers/index.js
```diff
@@ -4,6 +4,7 @@ import thunk from "redux-thunk";
 import cascadeReducers from "./cascade";
 import undoable from "./undoable";
 import config from "./config";
+import userinfo from "./userinfo";
```
seve, line 7: This is breaking the [airbnb style guide](https://airbnb.io/javascript/#naming--filename-matches-export).(lowercase import/filename, PascalCase export)

client/src/reducers/userinfo.js
```diff
@@ -0,0 +1,27 @@
+// jshint esversion: 6
```
seve, line 1: We don't use jshint anymore
```suggestion
```

client/src/components/menubar/authButtons.js
seve: There's a few other instances of this throughout the diff

server/requirements.txt
```diff
@@ -15,7 +15,7 @@ numba>=0.49.1
 numpy>=1.16.0
 packaging>=20.0
 pandas>=0.24.2
-psycopg2==2.7.7
+# psycopg2==2.7.7
```
seve: Why is this commented out?

server/test/unit/auth/test_auth.py
bmccandless: done

client/src/components/menubar/index.js
colinmegill: We've avoided doing any spreading at the component level. Also, this would pass in more properties (is not equiv)

client/src/reducers/userinfo.js
colinmegill: Removed in many other files

client/src/reducers/index.js
colinmegill: Hmm... I was matching the pattern that existed in reducers but also noticed it looked odd. Open an issue to change them all?

client/src/actions/index.js
```diff
@@ -41,6 +41,17 @@ async function configFetch(dispatch) {
   });
 }
 
+async function userInfoFetch(dispatch) {
+  return fetchJson("userinfo").then((response) => {
```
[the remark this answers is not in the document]
colinmegill: cc @bmccandless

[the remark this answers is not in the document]
bmccandless: /userinfo is the standard name for this sort of thing.  It's used by the OpenID connect specification.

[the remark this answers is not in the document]
seve: Ah thanks @bmccandless! Given that they use `UserInfo` for their object name should we reflect the same naming style when not referencing the endpoint? https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

client/src/components/menubar/index.js
seve: In this case what additional properties are being assigned to props?

I thought
```javascript
props = { auth: auth, userinfo: userinfo }
```
was equal to 
```javascript
props = { ...{auth, userinfo} }
```

client/src/reducers/index.js
seve: Will do 👍

client/src/components/menubar/authButtons.js
seve: Was this accidentally resolved? @colinmegill

colinmegill: I committed, and am fixing locally

client/src/actions/index.js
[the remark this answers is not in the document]
bmccandless: @seve @colinmegill, I don't have an opinion on the naming style;  I'll leave it up to you, maybe use whatever coding conventions you normally use.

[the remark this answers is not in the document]
colinmegill: I am comfortable/prefer lowercase, agree that is standard

client/src/actions/index.js
```diff
@@ -41,6 +41,17 @@ async function configFetch(dispatch) {
...
+    const userinfo = { ...response.userinfo };
```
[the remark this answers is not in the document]
seve: I suggested this change based of my assumption that the shape of the response was:
```javascript
response = {
    userinfo: {
         userinfo
    }
}
```
I'm not sure if this is actually the case.

This might be why smoke tests are failing

client/src/actions/index.js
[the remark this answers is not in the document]
colinmegill: Twilio uses PascalCase for things like PhoneNumber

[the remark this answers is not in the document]
colinmegill: Let's agree we won't do Pascal, since all our endpoints begin lowercase. I've looked through all of them and it seems like this is the only, or one of the only, that is multiWord. @bmccandless if you are ok adopting camelCase, let's do that.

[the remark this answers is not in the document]
seve: @colinmegill The proposition is that the endpoint remains `userinfo` but once it is in the front end's hands we will adopt our standard of a camelCase `userInfo`.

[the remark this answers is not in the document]
bmccandless: @colinmegill I'd like endpoint to be /userinfo;  no preference on how you name it in the front end code.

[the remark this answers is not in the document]
colinmegill: @bmccandless let's leave the endpoint as is, I don't mind either way

brianraymor:
@bmccandless - is this PR part of #481? If so, can you connect the PR?

bmccandless:
> @bmccandless - is this PR part of #481? If so, can you connect the PR?

@brianraymor,   #481 is specific to data portal.   This is more like a bug fix for #1519 combined with additional improvements for #1523.  There is actually a few small things left to do on the cellxgene backend, which are part of the #1524 epic.  I'll make some issues for those.

PR #1731, opened 2020-08-11
add CSP sources for obsolete browser prompt

seve:
Adds script hash and explicit domain to `img-src` directive

### DIFF
--- a/client/configuration/webpack/obsoleteHTMLTemplate.html
+++ b/client/configuration/webpack/obsoleteHTMLTemplate.html
@@ -1,4 +1,5 @@
 <script>
+  /** Reflect ANY changes to this script in the script hash in `server/eb/app.py` **/
   var root = document.getElementById("root");
   root.remove();
   var portals = document.getElementsByClassName("bp3-portal");
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -105,7 +105,7 @@ class WSGIServer(Server):
             "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
             + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:", "'https:///cellxgene.cziscience.com'"],
+            "img-src": ["'self'", "'data:'", "'https:///cellxgene.cziscience.com'"],
             "object-src": ["'none'"],
             "base-uri": ["'none'"],
             "frame-ancestors": ["'none'"],
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -105,7 +105,7 @@ class WSGIServer(Server):
             "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
             + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:", "https:///cellxgene.cziscience.com"],
+            "img-src": ["'self'", "data:", "'https:///cellxgene.cziscience.com'"],
             "object-src": ["'none'"],
             "base-uri": ["'none'"],
             "frame-ancestors": ["'none'"],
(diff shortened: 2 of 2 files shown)


server/eb/app.py
```diff
@@ -96,12 +96,16 @@ def __init__(self, app_config):
...
+        obsolete_browser_script_hash = 'SHA25-0028D52E332C015C3ED9929926F4000BB4020B8CB85C1F5769D6AA3BA711F58E'
         csp = {
             "default-src": ["'self'"],
             "connect-src": ["'self'"],
-            "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"] + script_hashes,
+            "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
+            + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:"],
+            "img-src": ["'self'", "data:", "https:///cellxgene.cziscience.com"],
```
seve, line 108: Does order matter here? Should the domain be going before `"data:"`

tihuan: That's a great question! I think the order shouldn't matter and the doc doesn't seem to mention it either 🤔 But definitely worth a double check lol 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src

client/configuration/webpack/obsoleteHTMLTemplate.html
```diff
@@ -1,4 +1,5 @@
 <script>
+  /** Reflect ANY changes to this script in the script hash in `server/eb/app.py` **/
```
tihuan, line 2: Thanks for the comment!

server/eb/app.py
seve, line 108:
```suggestion
            "img-src": ["'self'", "data:", "'https:///cellxgene.cziscience.com'"],
```

Forgot single quotes!

server/eb/app.py
```diff
@@ -105,7 +105,7 @@ def _before_adding_routes(app, app_config):
             "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
             + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:", "https:///cellxgene.cziscience.com"],
+            "img-src": ["'self'", "data:", "'https:///cellxgene.cziscience.com'"],
```
tihuan, line 108: Ahh so tricky 🤦 Sorry for missing that too! Do we need single quotes for `data:`?

seve: Unit tests are still failing, so I guess, not sure why they weren't failing earlier ¯\_(ツ)_/¯

This also could be an issue of the source order.


```suggestion
            "img-src": ["'self'", "'data:'", "'https:///cellxgene.cziscience.com'"],
```

tihuan: Hmm yeah and `data:` doesn't seem to need quotes 🤔 

https://stackoverflow.com/questions/40360109/content-security-policy-img-src-self-data

server/eb/app.py
```diff
@@ -96,12 +96,16 @@ def __init__(self, app_config):
...
+        obsolete_browser_script_hash = 'SHA25-0028D52E332C015C3ED9929926F4000BB4020B8CB85C1F5769D6AA3BA711F58E'
         csp = {
             "default-src": ["'self'"],
             "connect-src": ["'self'"],
-            "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"] + script_hashes,
+            "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
+            + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
-            "img-src": ["'self'", "data:"],
+            "img-src": ["'self'", "'data:'", "'https:///cellxgene.cziscience.com'"],
```
seve, line 108:
```suggestion
            "img-src": ["'self'",  "'https:///cellxgene.cziscience.com'", "data:"],
```

PR #1744, opened 2020-08-13
Release 0.16.1

maniarathi:


.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.16.0
+current_version = 0.16.1
```
seve, line 2: We don't actually want to push the .16.1 version bump to main

PR #1745, opened 2020-08-13
Properly generate hash and provide how-to

seve:
We did it! @tihuan and I were finally able to correctly generate the CSP hash after _a lot_ of tinkering.  I've provided a short how to above the hash, but I'll also write it here for posterity.

Here's how to generate and test a CSP hash without the need to make test deployments:
* Start up local cellxgene
* Install (requestly)[https://app.requestly.io/rules]
* Generate a new rule that modifies the header
* Set it to `Add` `Response` Header: `Content-Security-Policy` `Url` `Contains` `localhost`
* Go to prod and copy the CSP header value
 ```
default-src 'self'; connect-src 'self' sentry.prod.si.czi.technology; script-src 'self' 'unsafe-eval' 'sha256-/rmgOi/skq9MpiZxPv6lPb1PNSN+Uf4NaUHO/IjyfwM=' 'sha256-aFVmiBiEgb+j7fokN003RXrmU/RNz49S53HhxV5fz5k=' 'sha256-VaCbcpE7DqBJBipfdVmR6ctn0HR5HE1em2j+7NuCu2U=' 'sha256-w4uT8Jm+8zCdP4dnxhaN3LSKvtIq4ASVrheKTlJ0UMs=' www.google-analytics.com browser.sentry-cdn.com; style-src 'self' 'unsafe-inline'; img-src 'self' https://cellxgene.cziscience.com data: www.google-analytics.com; object-src 'none'; base-uri 'none'; frame-ancestors 'none'; upgrade-insecure-requests
```
* Paste that into the header value, save and reload the client
* You should see the missing hash in the console
image
* Viola!

Huge h/t to Timmy for helping me along the way and giving me some awesome resources :)

### DIFF
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -101,7 +101,7 @@ class WSGIServer(Server):
 
         # It is _very_ difficult to generate the correct hash manually,
         # consider forcing CSP to fail on the local server by intercepting the response via Requestly
-        # this should print the failing script's hash to console
+        # this should print the failing script's hash to console. See more here: https://github.com/chanzuckerberg/cellxgene/pull/1745
         obsolete_browser_script_hash = ["'sha256-/rmgOi/skq9MpiZxPv6lPb1PNSN+Uf4NaUHO/IjyfwM='"]
         csp = {
             "default-src": ["'self'"],
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -106,7 +106,7 @@ class WSGIServer(Server):
         csp = {
             "default-src": ["'self'"],
             "connect-src": ["'self'"],
-            "script-src": ["'self'", "'unsafe-eval'", "'unsafe-inline'"]
+            "script-src": ["'self'", "'unsafe-eval'"]
             + obsolete_browser_script_hash + script_hashes,
             "style-src": ["'self'", "'unsafe-inline'"],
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: server/eb/app.py [hunk 1]
```diff
@@ -98,7 +98,11 @@ def _before_adding_routes(app, app_config):
         server_config = app_config.server_config
         # This hash should be in sync with the script within
         # `client/configuration/webpack/obsoleteHTMLTemplate.html`
-        obsolete_browser_script_hash = ["'sha256-wl4OlniJEAFM1/VNFG/LBDQyPRkO7P/5kfQWf3+fPWE='"]
+
+        # It is _very_ difficult to generate the correct hash manually,
+        # consider forcing CSP to fail on the local server by intercepting the response via Requestly
```
tihuan, line 103: Super useful comment! Should we add the PR description link here too, so people can find your "How To" steps :D? Thank you!

seve: Good idea!

server/eb/app.py
```diff
@@ -98,7 +98,11 @@ def _before_adding_routes(app, app_config):
...
+        # this should print the failing script's hash to console
```
seve, line 104:
```suggestion
        # this should print the failing script's hash to console. See more here: https://github.com/chanzuckerberg/cellxgene/pull/1745
```

PR #1751, opened 2020-08-14
remove db_uri secret

MDunitz:
Dev environment is down with the error 
`AttributeError: ‘ServerConfig’ object has no attribute ‘user_annotations__hosted_tiledb_array__db_uri’`

This is due to pulling in the db_uri from AWS secrets and adding it to the server config instead of the dataset config

This pr should remove the error and allow the dev environment to deploy successfully
A later pr to refactor secrets retrieval to also assign to the dataset config or to store the db_uri on the server config will be necessary ([ticket 1752](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1752))

## Testing
- added unit test (pre code change) and it failed with an AttributeError
- updated code to remove db_uri and test passes

### DIFF
--- a/server/common/aws_secret_utils.py
+++ b/server/common/aws_secret_utils.py
@@ -1,0 +1,67 @@
+import logging
+import os
+import sys
+
+import boto3
+from flask import json
+
+from server.common.data_locator import discover_s3_region_name
+
+
+def handle_config_from_secret(app_config):
+    """Update configuration from the secret manager"""
+    secret_name = os.getenv("CXG_AWS_SECRET_NAME")
+    if not secret_name:
+        return
+
+    # need to find the secret manager region.
+    #  1. from CXG_AWS_SECRET_REGION_NAME
+    #  2. discover from dataroot location (if on s3)
+    #  3. discover from config file location (if on s3)
+    secret_region_name = os.getenv("CXG_AWS_SECRET_REGION_NAME")
+    if secret_region_name is None:
+        secret_region_name = discover_s3_region_name(app_config.multi_dataset__dataroot)
+        if not secret_region_name:
+            from server.eb.app import config_file
+            secret_region_name = discover_s3_region_name(config_file)
+    if not secret_region_name:
+        logging.error("Could not determine the AWS Secret Manager region")
+        sys.exit(1)
+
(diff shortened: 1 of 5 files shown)


maniarathi:
Hey @MDunitz can you create an issue to track the work you mentioned "to refactor secrets retrieval to also assign to the dataset config or to store the db_uri on the server config"? Also, is there any way we can add a test to verify that this works? (Maybe in `test_api.py`?)

server/common/utils.py
```diff
@@ -11,6 +11,7 @@
 from urllib.parse import urlsplit, urljoin
 import numpy as np
 import pandas as pd
+
```
maniarathi, line 14: This file seems to only have blank lines added? Is this a linting thing?

maniarathi:
This is dope! Thanks!

server/common/utils.py
MDunitz: no I ran into some import errors so I think that is leftover from moving the get_secrets around. Will remove

server/common/aws_secret_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    #  3. discover from config file location (if on s3)
+    secret_region_name = os.getenv("CXG_AWS_SECRET_REGION_NAME")
+    if secret_region_name is None:
+        secret_region_name = discover_s3_region_name(app_config.multi_dataset__dataroot)
+        if not secret_region_name:
+            from server.eb.app import config_file
+            secret_region_name = discover_s3_region_name(config_file)
+    if not secret_region_name:
+        logging.error("Could not determine the AWS Secret Manager region")
+        sys.exit(1)
```
bmccandless, line 29: If this is meant to be a general purpose function, then it should not sys.exit.   Maybe throw a configuration error here instead.   server/eb/app.py could then catch the exception and exit.

bmccandless: Same goes for the exit in get_secret_key

MDunitz: cool made a ticket -- https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1753

PR #1755, opened 2020-08-14
update to get_secrets_key

MDunitz:
## Need
- to pass in the db_uri from AWS secrets manager to the default_dataset_config
- to raise an exception within get_secret_key and call sys.exit in the eb app when that exception is raised. 

## Approach
- split out server_attrs and default_dataset_attrs
- create error and move sys.exit back into eb/app.py

## Testing
- update unit test to include db_uri message

### DIFF
--- a/server/common/aws_secret_utils.py
+++ b/server/common/aws_secret_utils.py
@@ -8,6 +8,7 @@ from flask import json
 from server.common.data_locator import discover_s3_region_name
 from server.common.errors import SecretKeyRetrievalError
 
+
 def handle_config_from_secret(app_config):
     """Update configuration from the secret manager"""
     secret_name = os.getenv("CXG_AWS_SECRET_NAME")
--- a/server/common/aws_secret_utils.py
+++ b/server/common/aws_secret_utils.py
@@ -6,7 +6,7 @@ import boto3
 from flask import json
 
 from server.common.data_locator import discover_s3_region_name
-
+from server.common.errors import SecretKeyRetrievalError
 
 def handle_config_from_secret(app_config):
     """Update configuration from the secret manager"""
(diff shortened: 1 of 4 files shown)


server/common/aws_secret_utils.py
```diff
@@ -49,6 +54,18 @@ def handle_config_from_secret(app_config):
...
+    # update default dataset configuration attributes
+    for key, attr in default_dataset_attrs:
+        cur_val = getattr(app_config.default_dataset_config, attr)
+        if cur_val:
+            continue
+
+        # replace the attr with the secret if it is not set
+        val = secrets.get(key)
+        if val:
+            logging.error(f"set {attr} from secret")
```
bmccandless, line 66: I'm not sure why I had this as logging.error.  It should be "info".  It is definitely not an error.

server/test/unit/common/test_app_config.py
```diff
@@ -121,10 +122,11 @@ def test_get_config_vars_from_aws_secrets(self, mock_get_secret_key):
...
             from server.common.aws_secret_utils import handle_config_from_secret
             # should not throw error
-            # "AttributeError: 'ServerConfig' object has no attribute 'user_annotations__hosted_tiledb_array__db_uri'"
+            # "AttributeError: 'XConfig' object has no attribute 'x'"
             handle_config_from_secret(config)
 
             # should throw 2 errors (one for each var set from a secret)
-            self.assertEqual(len(logger.output), 2)
+            self.assertEqual(len(logger.output), 3)
             self.assertIn('ERROR:root:set app__flask_secret_key from secret', logger.output[0])
```
bmccandless, line 130: You can also directly check for success by checking the config, e.g:
```
self.assertEqual(confg.server_config.app__flask_secret_key, "mock_flask_secret")
```

server/common/aws_secret_utils.py
MDunitz: ok cool, will update

server/test/unit/common/test_app_config.py
MDunitz: oh good call!

PR #1775, opened 2020-08-18
Update Dockerfile

prete:
# Changes
- Update Ubuntu Focal (20.04)
- Add `DEBIAN_FRONTEND=noninteractive` to prevent dialog boxes during installation

# Rationale
Current Dockerfile fails to build for cellxgene version 0.16.0. 
Main error involves `llvlmlite` installation. Update llvm to 10 fixes the issue. I worked around that by adding `llvm-10` to apt-get install packages list and then export `LLVM_CONFIG=/usr/bin/llvm-config-10` before installing cellxgene using pip3. Not sure that's the best way to solve it. Anyway, instead of going that way I believe it's cleaner if the base image is upgraded to Ubuntu Focal which also solves the issue.


seve:
Hi @prete, thanks for the great issue write up and the PR to go with it.

I did some digging around and it looks like the pip version that comes with `apt-get install python3-pip` is a bit outdated and the wheel generated by llvmlite doesn't support it ([See their report here](https://github.com/numba/llvmlite/issues/621)).

I found the solution to be upgrading pip via `pip3 install --upgrade pip` right before we install cellxgene in the Dockerfile. Could you check to see if that fix works for you and update the PR if it does?

seve:
I'm also a bit curious as to how you knew bumping the source image from `bionic` to `focal` could solve this issue :)

prete:
@seve just to be sure, would you want me to only add `pip3 install --upgrade pip` to the original Dockerfile or add that to the current changes I've made?

seve:
> @seve just to be sure, would you want me to only add `pip3 install --upgrade pip` to the original Dockerfile or add that to the current changes I've made?

Just to the original Dockerfile

seve:
This is what I have locally

```Dockerfile
FROM ubuntu:bionic

ENV LC_ALL=C.UTF-8
ENV LANG=C.UTF-8

RUN apt-get update && \
    apt-get install -y build-essential libxml2-dev python3-dev python3-pip zlib1g-dev python3-requests python3-aiohttp
RUN pip3 install --upgrade pip

RUN pip3 install cellxgene

ENTRYPOINT ["cellxgene"]
```

seve:
@prete Looks like our push tests aren't passing, this might be due to the branch not being up-to-date compared to `main`.  Try and see if rebasing/merging main into your branch fixes it.  Otherwise, LGTM!

prete:
Sorry @seve , wasn't checking. Merged main, checks passed.

seve:
@prete Thank you for the great bug report and PR to follow!

Issue #1779, opened 2020-08-18
[BUG] cxgtool type conversions introduce weirdness


mckinsel:
If you have a categorical int64 dtype with some `nan`s. cxgtool will cast to int32, which turns the `nan`s into `-2**31`s

Similarly, if you have a categorical float64, it will cast to float32 and introduce floating point issues, category 1.8 becomes 1.79999...

In both cases, the categories remain the same, so the schema will list a bunch of values with no cells.

PR #1782, opened 2020-08-19
Fixing bugs in cxg conversion tool

maniarathi:


maniarathi:
Screen Shot 2020-08-19 at 2 47 21 PM
Proof!

server/common/utils/type_conversion_utils.py
```diff
@@ -102,3 +129,10 @@ def can_cast_to_int32(dtype, array_values=None):
                 array_values.min() >= ii32.min and array_values.max() <= ii32.max) or array_values.empty:
```
mckinsel, line 129: So this gets called a little earlier than it used to, so now it can get a pd.Series with int categories that is not _ordered_, causing min to raise `TypeError`.

```
>>> s = pd.Series([1, 2, 3, 4, 1, 2, 3], dtype="category")
>>> s.dtype.categories
Int64Index([1, 2, 3, 4], dtype='int64')
>>> s.dtype.categories.dtype.kind
'i'
>>> s.min()
...many line traceback...
TypeError: Categorical is not ordered for operation min
you can use .as_ordered() to change the Categorical to an ordered one
```

I believe you want to check if min raises TypeError and if so, `array_values.cat.as_ordered().min()`

bkmartinjr:
Related to #1779

PR #1789, opened 2020-08-21
Dunitz 1685 hosted annotations

MDunitz:
Fixes for bugs found while deploying hosted annotations in dev env
- dont cache user annotations
- set tiledb default config
- add type category when returning user annotations
- save and retrieve under uri containing dataset name (instead of path) and user name (instead of id)

## Testing
- updated unit tests for new uri format
- updated unit tests to include schema hints to pass in data type
- tested extensively in dev env

### DIFF
--- a/client/src/annoMatrix/loader.js
+++ b/client/src/annoMatrix/loader.js
@@ -88,7 +88,7 @@ export default class AnnoMatrixLoader extends AnnoMatrix {
 
   addObsColumn(colSchema, Ctor, value) {
     /*
-		add a column to field, initializing with value.  Value may 
+		add a column to field, initializing with value.  Value may
     be one of:
       * an array of values
       * a primitive type, including null or undefined.
--- a/client/src/annoMatrix/loader.js
+++ b/client/src/annoMatrix/loader.js
@@ -88,7 +88,7 @@ export default class AnnoMatrixLoader extends AnnoMatrix {
 
   addObsColumn(colSchema, Ctor, value) {
     /*
-		add a column to field, initializing with value.  Value may
+		add a column to field, initializing with value.  Value may 
     be one of:
       * an array of values
       * a primitive type, including null or undefined.
--- a/server/common/annotations/hosted_tiledb.py
+++ b/server/common/annotations/hosted_tiledb.py
@@ -1,13 +1,10 @@
 import json
-import logging
 import os
 import re
 import time
 
 import pandas as pd
-import numpy as np
 import tiledb
-import numpy as np
 from flask import current_app
 from tiledb.libtiledb import TileDBError
 
@@ -16,7 +13,6 @@ from server.common.errors import AnnotationCategoryNameError
 from server.common.utils.sanitization_utils import sanitize_values_in_list
 from server.common.utils.type_conversion_utils import get_dtypes_and_schemas_of_dataframe, get_dtype_of_array
 from server.db.cellxgene_orm import Annotation
-logger = logging.getLogger(__name__)
 
 
 class AnnotationsHostedTileDB(Annotations):
--- a/server/common/annotations/hosted_tiledb.py
+++ b/server/common/annotations/hosted_tiledb.py
@@ -137,3 +137,16 @@ class AnnotationsHostedTileDB(Annotations):
 
         self.db.session.add(annotation)
         self.db.session.commit()
+
+    def update_parameters(self, parameters, data_adaptor):
+        params = {}
+        params["annotations"] = True
+        params["user_annotation_collection_name_enabled"] = False
+
+        if self.ontology_data:
+            params["annotations_cell_ontology_enabled"] = True
+            params["annotations_cell_ontology_terms"] = self.ontology_data
+        else:
+            params["annotations_cell_ontology_enabled"] = False
+
+        parameters.update(params)
--- a/server/common/annotations/hosted_tiledb.py
+++ b/server/common/annotations/hosted_tiledb.py
@@ -1,17 +1,22 @@
 import json
+import logging
 import os
 import re
 import time
 
 import pandas as pd
+import numpy as np
 import tiledb
+import numpy as np
 from flask import current_app
+from tiledb.libtiledb import TileDBError
 
 from server.common.annotations.annotations import Annotations
 from server.common.errors import AnnotationCategoryNameError
 from server.common.utils.sanitization_utils import sanitize_values_in_list
 from server.common.utils.type_conversion_utils import get_dtypes_and_schemas_of_dataframe, get_dtype_of_array
-from server.db.cellxgene_orm import CellxGeneDataset, Annotation
+from server.db.cellxgene_orm import Annotation
+logger = logging.getLogger(__name__)
 
 
 class AnnotationsHostedTileDB(Annotations):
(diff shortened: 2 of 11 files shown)


server/common/annotations/hosted_tiledb.py
```diff
@@ -21,6 +22,13 @@ def __init__(self, directory_path, db):
         super().__init__()
         self.db = db
         self.directory_path = directory_path
+        config = tiledb.Config()
+        config["vfs.s3.region"] = "us-west-2"
+        self.ctx = tiledb.Ctx(config=config)
+        try:
+            tiledb.default_ctx(config)
+        except TileDBError:
+            pass
```
bmccandless, line 31: Don't hard-code the region.

There is already a config for tiledb in the CxgAdaptor which will have the correct region set (along with other configuration options).   Please set that to be the default_ctx in the set_tiledb_context function of CxgAdaptor in server/data_cxg/cxg_adaptor.py.  Then you can remove lines 25-31

server/common/annotations/hosted_tiledb.py
```diff
@@ -93,20 +107,20 @@ def convert_to_pandas_df(self, tileDBArray):
...
         timestamp = time.time()
-        dataset_name = data_adaptor.get_location()
-        dataset_id = self.db.get_or_create_dataset(dataset_name)
+        dataset_location = data_adaptor.get_location()
+        dataset_id = self.db.get_or_create_dataset(dataset_location)
+        dataset_name = data_adaptor.get_title()
         user_id = self.db.get_or_create_user(user_id)
 
-        uri = f"{self.directory_path}-{dataset_name}-{user_id}-{timestamp}"
+        uri = f"{self.directory_path}{dataset_name}/{user_name}/{timestamp}"
```
bmccandless, line 118: is directory path guaranteed to end with a "/".  Might need to check for that.

server/common/annotations/hosted_tiledb.py
```diff
@@ -93,20 +107,20 @@ def convert_to_pandas_df(self, tileDBArray):
         return new_df
 
     def write_labels(self, df, data_adaptor):
-
         user_id = current_app.auth.get_user_id()
+        user_name = current_app.auth.get_user_name().replace(' ', '')
```
bmccandless, line 111: Also why remove spaces?  I'm not sure if there is a current need, but that would make it harder to map from uri to user name.

server/common/annotations/hosted_tiledb.py
MDunitz: it looks like it is currently hard coded to a different region in the cxg adaptor https://github.com/chanzuckerberg/cellxgene/blob/bc150a84698cf7a28900df865ad9daea85dc0c1c/server/data_cxg/cxg_adaptor.py#L25

server/common/annotations/hosted_tiledb.py
MDunitz: It looks like adding spaces to file names in s3 causes some downloading issues/isnt recommended
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html

I can replace it with an underscore if that would be clearer

client/src/annoMatrix/loader.js
```diff
@@ -88,7 +88,7 @@ export default class AnnoMatrixLoader extends AnnoMatrix {
 
   addObsColumn(colSchema, Ctor, value) {
     /*
-		add a column to field, initializing with value.  Value may 
+		add a column to field, initializing with value.  Value may
```
maniarathi, line 91: Weird line break?

server/common/annotations/hosted_tiledb.py
maniarathi, line 118: I had a general question about naming of annotation files; I understand that it might be easier to debug if you have filenames that we can quickly take a look at but to me this is not a good enough reason to add a bunch of rules in the business logic tier of the application that should be at the data access layer. We are starting to blur the lines which inevitably will lead to faulty code. For example, `time.time()` is almost guaranteed to cause an issue at some point because the [clocks on different servers aren’t always set exactly the same](https://docs.python.org/2/library/time.html#time.time). I’d advocate strongly to revise such that the database keep track of the user id, dataset id, etc. and use `uuid4()` to generate filenames. Or at the very least, tag on a `uuid4()` id to the filename.

server/common/annotations/hosted_tiledb.py
MDunitz: discussed over slack

server/common/annotations/hosted_tiledb.py
MDunitz: discussed over slack

client/src/annoMatrix/loader.js
MDunitz: removed

server/common/annotations/hosted_tiledb.py
MDunitz: So the time on the filename is just used as a differentiator between files for a user/dataset. The time the annotation is saved to the db (created_at) is the time we sort by. The database is the place we are keeping track of user id/dataset id etc. When retrieving annotations we read from the db to get the uri. 
I can append a uuid4() to the end of the filename to alleviate any concerns about collisions.

MDunitz: discussed further over slack

server/common/annotations/hosted_tiledb.py
```diff
@@ -6,21 +6,31 @@
...
+        if directory_path[-1] == "/":
+            self.directory_path = directory_path
+        else:
+            self.directory_path = directory_path + "/"
+        tiledb_config = server_config.adaptor__cxg_adaptor__tiledb_ctx
+        self.ctx = tiledb.Ctx(config=tiledb_config)
+        try:
+            tiledb.default_ctx(tiledb_config)
+        except TileDBError:
+            pass
```
bmccandless, line 33: What I meant earlier is you can remove all of these lines.   
We've already created a tiledb_config in the CxgAdaptor class.  
The fix was to  add "tiledb.default_ctx( CxgAdaptor.tiledb_ctx) to the "set_tiledb_context" function in CxgAdaptor.
Once that is done, there is no longer a need to pass in a ctx into any of the tiledb functions.
So, then you can remove self.ctx here, and the places where you used it.

MDunitz: sorry I misunderstood your earlier message, just updated

server/data_cxg/cxg_adaptor.py
```diff
@@ -51,8 +51,13 @@ def set_tiledb_context(context_params):
         """Set the tiledb context.  This should be set before any instances of CxgAdaptor are created"""
         try:
             CxgAdaptor.tiledb_ctx = tiledb.Ctx(context_params)
+            tiledb.default_ctx(context_params)
+
         except tiledb.libtiledb.TileDBError as e:
-            raise ConfigurationError(f"Invalid tiledb context: {str(e)}")
+            if e.message == "Global context already initialized!":
```
bmccandless, line 57: This is likely only triggered in unit tests where we would initialize more than once, which would be okay.  But it would be a bug if this happened outside of testing,  and in this implementation it would be a silent bug.  

The only real problem is if we try to change the tiledb config after we've already set the default_ctx.  

I suggest doing this:

```
if e.message == "Global context already initialized!":
   if tiledb.default_ctx().config().dict() != CxgAdaptor.tiledb_ctx.config().dict():
      raise ConfigurationError(f"Cannot change tiledb configuration once it is set")
else:
    raise ConfigurationError(f"Invalid tiledb context: {str(e)}")
```

Now the only problem would be if we had a test file where we wanted to set the tiledb config differently in two different tests.  I don't think we have that, but if we did we could split those into different test files, or find some other workaround.

MDunitz: 👍

PR #1791, opened 2020-08-24
thuang-fix-static-asset-font

tihuan:
This seems to fix the font URL path, at least locally for both `:3000` and `:5005`

Screen Shot 2020-08-24 at 4 01 07 PM

Screen Shot 2020-08-24 at 4 01 28 PM

### DIFF
--- a/client/configuration/webpack/webpack.config.dev.js
+++ b/client/configuration/webpack/webpack.config.dev.js
@@ -31,7 +31,11 @@ const devConfig = {
         test: /\.(jpg|png|gif|eot|svg|ttf|woff|woff2|otf)$/i,
         loader: "file-loader",
         include: [nodeModules, fonts],
-        query: { name: "static/assets/[name].[ext]" },
+        query: {
+          name: "static/assets/[name].[ext]",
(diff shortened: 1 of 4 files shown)


client/configuration/webpack/webpack.config.dev.js
```diff
@@ -31,7 +31,11 @@ const devConfig = {
         test: /\.(jpg|png|gif|eot|svg|ttf|woff|woff2|otf)$/i,
         loader: "file-loader",
         include: [nodeModules, fonts],
-        query: { name: "static/assets/[name].[ext]" },
+        query: {
+          name: "static/assets/[name].[ext]",
+          // (thuang): This is needed to make sure @font url path is '/static/assets/'
+          publicPath: "/",
```
maniarathi, line 37: Does this not cause the proxy bug again since it's an absolute path?

tihuan: This is for local development only, so all is well!

maniarathi:
lol guess it was good to go; czibotmerge was eager to get it in haha

PR #1793, opened 2020-08-25
fix pca call in reembeddings

bmccandless:
This had the wrong dim passed into n_comps,
and so failed when the number of genes was less than 50.


maniarathi:
Seconding on adding a test please.

bmccandless:
> lgtm -- should we add a test to catch this in the future?

I've added an issue to add more testing: #1794 
We should definitely stress it out more before removing the "experimental" disclaimer.

Issue #1780, opened 2020-08-18
[BUG] FE attempts to access userinfo when authentication is disabled.


bmccandless:
Run cellxgene with authentication_enabled: false
And user_annotations/enabled : false
The FE will get an error and lot load the page:

Here is the error message
```
controls.js:30 TypeError: Cannot read property 'userinfo' of null
    at index.js:46
    at async Promise.all (:5005/index 3)
```

When authentication is not enabled, the config endpoint does not return an authentication section.

seve:
The userinfo endpoint is always fetched on initial data load:

```javascript
const doInitialDataLoad = () =>
  catchErrorsWrap(async (dispatch) => {
    dispatch({ type: "initial data load start" });

    try {
      const [config, schema] = await Promise.all([
        configFetch(dispatch),
        schemaFetch(dispatch),
        userColorsFetchAndLoad(dispatch),
        userInfoFetch(dispatch),
      ]);
...
```

bmccandless:
> The userinfo endpoint is always fetched on initial data load:

Thanks @seve.  So some options:

1.  either the FE needs to handle the case when the userinfo endpoint return null (when the server is configured to have authentication disabled)
2. or,  the FE needs to wait for the config endpoint to complete, and check if an authentication is enabled before before fetching the userinfo.

bmccandless:
Once this issue is complete, then the dev deployment config file needs to be updated.
authentication type -> oauth
default dataset config:  authentication -> disable

PR #1795, opened 2020-08-26
Fix frontend mishandling of null userinfo

bmccandless:
If the authentication is disabled, the userinfo endpoint returns null.
This case needs to be handled.

 #1780

### DIFF
--- a/client/src/actions/index.js
+++ b/client/src/actions/index.js
@@ -43,7 +43,7 @@ async function configFetch(dispatch) {
 
 async function userInfoFetch(dispatch) {
   return fetchJson("userinfo").then((response) => {
-    const userinfo = { ...response.userinfo };
+    const userinfo = { ...(response ? response.userinfo : null) };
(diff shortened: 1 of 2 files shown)


client/src/actions/index.js
```diff
@@ -43,7 +43,7 @@ async function configFetch(dispatch) {
 
 async function userInfoFetch(dispatch) {
   return fetchJson("userinfo").then((response) => {
-    const userinfo = { ...response.userinfo };
+    const userinfo = { ...(response ? response.userinfo : null) };
```
seve, line 46: We can simplify this to a conditional destructure like this:
```suggestion
    const { userinfo } = response || {};
```

seve: (The line should have been: `const { userinfo } = response` to begin with.)

bmccandless: Thanks, fixed.

PR #1809, opened 2020-09-01
Correctly check if mini histograms shouldn't be rendered

### LINKED ISSUE #1808: [PROD Bug] Mini histograms not rendering on color-by gene
Mini histograms are toggled, but not rendered when a user colors by a gene

seve:
Previously mini histogram rendering would abort if 
```javascript
AnnotationsHelpers.isContinuousAnnotation(schema, colorAccessor)  
 ```

returned a falsey value.  Since this function parsed the schema for `colorAccessor`, it would correctly return `undefined` when `colorAccessor` was a gene name, and incorrectly abort the render since `undefined` is falsey.

This PR swaps out our call of `isContinousAnnotation` for a check if `isCategoricalAnnotation` strictly equal to `true`

---

Fixes #1808

### DIFF
--- a/client/src/components/categorical/value/index.js
+++ b/client/src/components/categorical/value/index.js
@@ -439,7 +439,9 @@ class CategoryValue extends React.Component {
 
     if (
       !this.shouldRenderStackedBarOrHistogram ||
-      !AnnotationsHelpers.isContinuousAnnotation(schema, colorAccessor)
+      // This function returns true on categorical annotations(when stacked bar should not render),
+      //  in cases where the colorAccessor is a gene this function will return undefined since genes do not live on the schema
+      AnnotationsHelpers.isCategoricalAnnotation(schema, colorAccessor) === true
     ) {
       return null;
     }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: client/src/components/categorical/value/index.js [hunk 1]
```diff
@@ -439,7 +439,9 @@ class CategoryValue extends React.Component {
 
     if (
       !this.shouldRenderStackedBarOrHistogram ||
-      !AnnotationsHelpers.isContinuousAnnotation(schema, colorAccessor)
+      // This function returns true on categorical annotations(when stacked bar should not render),
+      //  in cases where the colorAccessor is a gene this function will return undefined since genes do not live on the schema
+      AnnotationsHelpers.isCategoricalAnnotation(schema, colorAccessor) === true
```
maniarathi, line 444: If `isCategoricalAnnotations` returns `undefined` then we _do_ want to render right? In which case wouldn't undefined not equal true thus not rendering?

maniarathi: Ahh nvm. I see. `return null` means that it doesn't get rendered and since the conditional will evaluate to false, we'll be clear.

seve: If the return value of `isCategoricalAnnotations` is strictly equal to `true` this render function returns null not rendering. `undefined !== true` and doesn't enter the if block

seve: Would the alternative of `isContinousAnnotations === false` be more clear?

maniarathi: Nah, I'm good! Just took me a second to parse it.

Issue #1778, opened 2020-08-18
[Integration][E2E] Authentication in the deployment


bmccandless:
A few minor issues have already come up with authentication working on the dev deployment, and other issues
may arise.  This is a placeholder for work that might need to be done to get authentication working end to end on
the dev deployment.

mweiden:
@bmccandless seems like this should incorporated in our estimates as part of the standard ~20-30% estimate buffer, no? Can we close this or am I missing a critical bit of info here?

bmccandless:
It doesn't really matter to me how the work is split into tickets.   I can reopen one of the other tickets, maybe #1519.   Or I can treat anything that comes up as a bug.   I created this ticket just to capture troubleshooting work that I know will need to be done, so we can plan for it next sprint.

tihuan:
Thanks for the bug ticket, @bmccandless !

It seems to me that there are two components to this ticket:

1. There are already minor bugs discovered
2. Likely there will be some more bugs we will discover in Sprint

I think for easier estimation, maybe we can dedicate this bug ticket to Item 1, and we'll create a another AuthN bug ticket for Item 2 if they do arise?

If that sounds good, then we just need Brian's help to add more details about the minor bugs, so it's easier for the Team to estimate the scale/complexity of the bugs

What do you both think? Thank you!

tihuan:
Adding #502 (api.cellxgene.cziscience.com) as a blocker, since I remember that's what we said in the Sprint Planning lol But if it's not, please feel free to take that out! Thank you!

@bmccandless

lesliecodes:
deploy away!

PR #1819, opened 2020-09-09
separate backend base url from frontend

bmccandless:
There is a plan to have the API for both cellxgene and dataportal at the same domain ( https://api.cellxgene.cziscience.com). And for the frontend to be served from https://cellxgene.cziscience.com.
For auth specifically, the login provides a callback to the oauth server (which is now a backend url), and the callback api redirects the user back to a frontend page. So, frontend and backend needed to be two distinct concepts.

 part of chanzuckerberg/cellxgene#1778


server/auth/auth_oauth.py
```diff
@@ -255,8 +254,8 @@ def check_jwt_payload(self, id_token):
                 }
         if rsa_key:
             options = {}
-            if not rsa_key["n"] or not rsa_key["e"]:
-                # this is a mock auth server, do not validate
+            if self.client_id == "mock_client_id" and (not rsa_key["n"] or not rsa_key["e"]):
```
maniarathi, line 257: Can you remove the testing logic from the production code and instead mock out the expected value for `options`?

server/app/app.py
```diff
@@ -353,32 +363,40 @@ def __init__(self, app_config):
...
                     lambda dataset, url_dataroot=url_dataroot: dataset_index(url_dataroot, dataset),
                     methods=["GET"],
                 )
                 self.app.add_url_rule(
-                    f"/{url_dataroot}/<dataset>/static/<path:filename>",
+                    f"{backend_path}/{url_dataroot}/<dataset>/static/<path:filename>",
                     f"static_assets_{url_dataroot}",
                     view_func=lambda dataset, filename: send_from_directory("../common/web/static", filename),
                     methods=["GET"]
+
```
maniarathi, line 395:
```suggestion
```

server/app/app.py
```diff
@@ -353,32 +363,40 @@ def __init__(self, app_config):
...
                 self.app.add_url_rule(
-                    f"/{url_dataroot}/<dataset>/",
+                    f"{backend_path}/{url_dataroot}/<dataset>/",
                     f"dataset_index_{url_dataroot}",
                     lambda dataset, url_dataroot=url_dataroot: dataset_index(url_dataroot, dataset),
                     methods=["GET"],
                 )
                 self.app.add_url_rule(
-                    f"/{url_dataroot}/<dataset>/static/<path:filename>",
+                    f"{backend_path}/{url_dataroot}/<dataset>/static/<path:filename>",
```
maniarathi, line 391: Are you sure that the backend_path should be used here? Are the static assets served from the API URL?

maniarathi (commented):
Curious, why are APIs served from separate dataroots? The API doesn't change per dataroot no?

maniarathi:
Hey @bmccandless , I'm curious. What specific problem in auth led you to needing to separate out the backend and frontend URLs?

bmccandless:
> Hey @bmccandless , I'm curious. What specific problem in auth led you to needing to separate out the backend and frontend URLs?

@maniarathi.    It's due to the plan to have the API for both cellxgene and dataportal at the same domain (https://api.dev.corpora.cziscience.com).  And for the frontend to be served from https://cellxgene.cziscience.com.
For auth specifically, the login provides a callback to the oauth server (which is now a backend url), and the callback api redirects the user back to a frontend page.   So, frontend and backend needed to be two distinct concepts.

server/app/app.py
bmccandless: Thanks, yes this is a mistake.   Actually not really sure we even need this rule, since static assets shouldn't depend on the dataset.   We server static assets out of /static/cellgene/....
I'll revert this.

maniarathi: Ah, we definitely need this rule. This is for folks who are hosting cellxgene on their own and without this, the static assets are not accessible if they use a proxy.

server/app/app.py
```diff
@@ -353,32 +363,40 @@ def __init__(self, app_config):
...
                 bp_api = Blueprint(
-                    f"api_dataset_{url_dataroot}", __name__, url_prefix=f"/{url_dataroot}/<dataset>" + api_version
+                    f"api_dataset_{url_dataroot}", __name__,
+                    url_prefix=f"{backend_path}/{url_dataroot}/<dataset>" + api_version
                 )
                 resources = get_api_resources(bp_api, url_dataroot)
                 self.app.register_blueprint(resources.blueprint)
                 self.app.add_url_rule(
-                    f"/{url_dataroot}/<dataset>/",
+                    f"{backend_path}/{url_dataroot}/<dataset>/",
```
maniarathi, line 385: Actually ditto here: you want the dataset to be accessible right at the frontend path, not the API path right?

MDunitz:
tiny correction we want the dataportal and cellxgene api to be served from https://api.cellxgene.cziscience.com

server/app/app.py
bmccandless: Yes, I got carried away :). fixed.

server/auth/auth_oauth.py
bmccandless: It might be a little tricky to mock out because the code here is running in a different process than the test case.

server/test/unit/common/test_app_config.py
```diff
@@ -133,3 +134,23 @@ def test_get_config_vars_from_aws_secrets(self, mock_get_secret_key):
             self.assertEqual(config.server_config.app__flask_secret_key, "mock_flask_secret")
             self.assertEqual(config.server_config.authentication__params_oauth__client_secret, "mock_oauth_secret")
             self.assertEqual(config.default_dataset_config.user_annotations__hosted_tiledb_array__db_uri, "mock_db_uri")
+
+    def test_backend_base_url(self):
+
+        # test the backend_base_url feature, and that it can contain a path
+        c = AppConfig()
```
maniarathi, line 141: For readability, please avoid using single character variable names.

server/common/app_config.py
```diff
@@ -743,6 +747,20 @@ def exceeds_limit(self, limit_name, value):
...
+            return f"http://{self.app__host}:{self.app__port}"
+        else:
+            return self.app__backend_base_url
+
+    def get_frontend_base_url(self):
+        if self.app__frontend_base_url == "local":
+            return f"http://{self.app__host}:{self.app__port}"
+        elif self.app__frontend_base_url is None:
+            return self.get_backend_url()
+        else:
```
maniarathi, line 761: Pythonic nit: don't need the else here since the previous conditionals return.

server/common/default_config.py
```diff
@@ -14,6 +14,20 @@
...
+    # There are two reasons to set these parameters:
+    #  1. Oauth authentication is used; the oauth server will redirect back to the backend_base_url after login,
+    #     which then redirects back to the frontend_base_url.  If the frontend_base_url is not set, it will default to
+    #     the backend_base_url.  If oauth authentication is used, the backend_base_url must be set.
+    #     For a local test (where the server runs on "http://localhost:<port>"), then the backend_base_url may be
+    #     set to the string "local".
+    #  2. The cellxgene deploymnent is in an environment where the frontend and backend have
+    #     different base urls.  In this case both backend_base_url and frontend_base_url must be set.
+    #     It is up to the server admin to ensure that the networking is setup correctly for this environment.
+    backend_base_url: null
```
maniarathi, line 28: How do you feel about renaming this `api_base_url` and `web_base_url`? I think it can get a little confusing since the frontend_base_url is _technically_ still backend. Having `api` in what you now name `backend_base_url` makes it much more clear what it is actually serving.

server/auth/auth_oauth.py
maniarathi: Can you try please and let me know. It is really not a good practice to have unit testing conditionals in production code.

maniarathi (changes requested):

maniarathi:
Friendly ping about this question: 
> Curious, why are APIs served from separate dataroots? The API doesn't change per dataroot no?

server/auth/auth_oauth.py
bmccandless: One option is to have a config parameter for this.  E.g. jwt_code_options, which is a dictionary of key/values.
Another option is to switch on an environment variable.

server/common/default_config.py
bmccandless: Sure, sounds good.

server/auth/auth_oauth.py
maniarathi: You are not able to `patch` the function `check_jwt_payload` and define it's return value?

maniarathi:
> > Hey @bmccandless , I'm curious. What specific problem in auth led you to needing to separate out the backend and frontend URLs?
> 
> @maniarathi. It's due to the plan to have the API for both cellxgene and dataportal at the same domain (https://api.dev.corpora.cziscience.com). And for the frontend to be served from https://cellxgene.cziscience.com.
> For auth specifically, the login provides a callback to the oauth server (which is now a backend url), and the callback api redirects the user back to a frontend page. So, frontend and backend needed to be two distinct concepts.

Can you add the above explanation to the top level comment of this commit? Otherwise it seems very mysterious why this PR was made :)

client/index_template.html
```diff
@@ -38,9 +38,11 @@
   <body>
     <script type="text/javascript">
       window.CELLXGENE = {};
+      backend_base_url = "{{BACKEND_BASE_URL}}";
+      prefix = backend_base_url ? backend_base_url : location.origin;
       window.CELLXGENE.API = {
-        prefix: `${location.origin}${location.pathname}api/`,
-        version: "v0.2/",
+          prefix: `${prefix}${location.pathname}api/`,
+          version: "v0.2/",
```
seve, line 45: Do we need to recompute any CSP hashes as a result of altering our script?

server/test/unit/common/test_app_config.py
bmccandless: > Friendly ping about this question:
> 
> > Curious, why are APIs served from separate dataroots? The API doesn't change per dataroot no?

I think it's mostly historical reasons for that.
It could be different, something like  e.g. /api/v0.2/schema?dataset=datasetname.cxg.
Maybe now that we are separating out the frontend and backend it makes sense to revisit this.

client/index_template.html
```diff
@@ -38,9 +38,11 @@
   <body>
     <script type="text/javascript">
       window.CELLXGENE = {};
+      backend_base_url = "{{BACKEND_BASE_URL}}";
```
seve, line 41: Prefer JS convention of camelCase `backendBaseURL`.

Is this supposed to be a string literal?  `backend_base_url` will _always_ `===` the string`{{BACKEND_BASE_URL}}`

This means that the following ternary will never assign `location.origin` to `prefix`

seve (changes requested):

client/index_template.html
bmccandless: No, I don't think so.

client/index_template.html
bmccandless: This is a template.  The BACKEND_BASE_URL gets replaced with the what the server passed when it rendered the template.  Other examples of this are SCRIPTS and INLINE_SCRIPTS, which are handled differently because they are lists.

server/auth/auth_oauth.py
bmccandless: The function is run in a different process than the test.    I don't think patching will work in this case.

client/index_template.html
seve: Ah I thought we were using a different templating engine and not familiar with lodash'. I thought that it wouldn't replace any text inside the literal string.  LGTM!

client/index_template.html
seve: You might want to try to test the CSP rules.  I wrote something up here on how to do that: https://github.com/chanzuckerberg/cellxgene/pull/1745

seve: The CSP header might have changed since then,  its probably best to grab it from a dev instance

server/test/unit/common/test_app_config.py
maniarathi: yeah, the URL is super bizarre and non intuitive. we should be aiming for something like:
`https://api.cellxgene.dev.single-cell.czi.technology/cellxgene/userinfo` or
`https://api.cellxgene.dev.single-cell.czi.technology/cellxgene/schema?dataset=datasetname.cxg`. Or even a dataset UUID.

There are lots of way to handle versioning but one clean way is for the versions to go into the Accept and Content-Type headers.

client/index_template.html
bmccandless: @seve,  I'm not familiar with how we handle hashes, but don't think we hash the index_template.html.  I see it for scripts and for main but not for this.

server/auth/auth_oauth.py
bmccandless: I added a config parameter to handle the decode options.

client/index_template.html
bmccandless: I changed to camelCase

server/test/unit/common/test_app_config.py
```diff
@@ -19,46 +20,46 @@ def mockenv(**envvars):
 
 class AppConfigTest(unittest.TestCase):
     def test_update(self):
-        c = AppConfig()
-        c.update_server_config(app__verbose=True, multi_dataset__dataroot="datadir")
-        v = c.server_config.changes_from_default()
+        config = AppConfig()
+        config.update_server_config(app__verbose=True, multi_dataset__dataroot="datadir")
+        v = config.server_config.changes_from_default()
```
MDunitz, line 25: nit -- can we avoid 1 letter var names?

maniarathi (approved):

MDunitz (approved):
lgtm

seve (approved):
LGTM, thanks for the changes.  CSP _shouldn't_ be an issue, but if we see any signal we know where to look.

client/index_template.html
tihuan: Ooh thanks for pointing that out, @seve ! Yeah I have a feeling we'll need to update the CSP hash 👍 

@seve can you help us with that for this PR? Thank you!

PR #1829, opened 2020-09-11
Fixes from frontend/backend url separation

bmccandless:
This fixes the CORS and CSP headers.

Also, in thie commit, I removed the cors_supports_credentials config parameter,
which was recently introduced.
Instead, the logic determines the need to use CORS headers if the
web_page_url is set.

 #1778

### DIFF
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -4,10 +4,11 @@ import sys
 import os
 import hashlib
 import base64
+from urllib.parse import urlparse
 from flask import json
 import logging
 from flask_talisman import Talisman
-
+from flask_cors import CORS
 from server.common.aws_secret_utils import handle_config_from_secret
 from server.common.errors import SecretKeyRetrievalError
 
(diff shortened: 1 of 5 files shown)


server/eb/app.py
```diff
@@ -41,6 +42,16 @@ def __init__(self, app_config):
...
         server_config = app_config.server_config
+
+        # add the api_base_url to the connect_src csp header.
+        extra_connect_src = []
+        api_base_url = server_config.get_api_base_url()
+        if api_base_url:
+            parse_api_base_url = urlparse(api_base_url)
+            extra_connect_src = [f"{parse_api_base_url.scheme}://{parse_api_base_url.netloc}"]
+
+        print("EXTRA CONNECT SRC", extra_connect_src)
```
MDunitz, line 53: do you need this print statement?

MDunitz (approved):
lgtm

server/eb/app.py
bmccandless: no, thanks, I'll remove

PR #1831, opened 2020-09-11
Fix InfoFormat parameter checking

### LINKED ISSUE #1825: Info Drawer failing on dev
Dataset: https://cellxgene.dev.single-cell.czi.technology/d/krasnow_lab_human_lung_cell_atlas_smartseq2-2.cxg/

Error: 
```
react-dom.production.min.js:209 TypeError: Cannot read property 'map' of undefined
    at infoFormat.js:11
    at infoFormat.js:183
    at Yi (react-dom.production.min.js:153)
    at Va (react-dom.production.min.js:175)
    at Ha (react-dom.production.min.js:175)
    at Pa (react-dom.production.min.js:174)
    at mc (react-dom.production.min.js:268)
    at ls (react-dom.production.min.js:246)
    at cs (react-dom.production.min.js:246)
    at Jc (react-dom.production.min.js:239)
ec @ react-dom.production.min.js:209
react-dom.production.min.js:123 Uncaught TypeError: Cannot read property 'map' of undefined
    at infoFormat.js:11
    at infoFormat.js:183
    at Yi (react-dom.production.min.js:153)
    at Va (react-dom.production.min.js:175)
    at Ha (react-dom.production.min.js:175)
    at Pa (react-dom.production.min.js:174)
    at mc (react-dom.production.min.js:268)
    at ls (react-dom.production.min.js:246)
    at cs (react-dom.production.min.js:246)
    at Jc (react-dom.production.min.js:239)
```

EDIT(9/11): Conditional render functions in `InfoFormat` component are not handling render cases where their arguments are undefined.

`InfoFormat`'s render functions need to be checked and updated where needed to handle undefined arguments

seve:
Went through and ensured that undefined/null values were caught and handled correctly in render functions.  Also documented some of the more complicated functions.

---
Closes #1825

### DIFF
--- a/client/src/components/infoDrawer/infoFormat.js
+++ b/client/src/components/infoDrawer/infoFormat.js
@@ -3,7 +3,7 @@ import React from "react";
 
 const renderContributors = (contributors, affiliations, skeleton) => {
   // eslint-disable-next-line no-constant-condition --  Temp removed contributor section to avoid publishing PII
-  if (contributors?.length === 0 && true) return null;
+  if (!contributors || (contributors.length === 0 && true)) return null;
   return (
     <>
       <H3 className={skeleton ? Classes.SKELETON : null}>Contributors</H3>
--- a/client/src/components/infoDrawer/infoFormat.js
+++ b/client/src/components/infoDrawer/infoFormat.js
@@ -3,7 +3,7 @@ import React from "react";
 
 const renderContributors = (contributors, affiliations, skeleton) => {
   // eslint-disable-next-line no-constant-condition --  Temp removed contributor section to avoid publishing PII
-  if (!contributors || (contributors.length === 0 && true)) return null;
+  if (!contributors || contributors.length === 0 || true) return null;
   return (
     <>
       <H3 className={skeleton ? Classes.SKELETON : null}>Contributors</H3>
--- a/client/src/components/infoDrawer/infoFormat.js
+++ b/client/src/components/infoDrawer/infoFormat.js
@@ -25,6 +25,7 @@ const renderContributors = (contributors, affiliations, skeleton) => {
   );
 };
 
+// generates a list of unique institutions by order of appearance in contributors
 const buildAffiliations = (contributors = []) => {
   const affiliations = [];
   contributors.forEach((contributor) => {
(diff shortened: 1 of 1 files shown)


seve:
Looking at it now, I'm not a huge fan of not sticking to a single practice of nullish handling.  Some are caught in the params of `InfoFormat` some are caught in the individual render functions. I can justify `dataPortalProps` and `skeleton` but not sure about the others.

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -43,11 +44,7 @@ const renderAffiliations = (affiliations, skeleton) => {
       <H3 className={skeleton ? Classes.SKELETON : null}>Affiliations</H3>
       <UL>
         {affiliations.map((item, index) => (
-          <div
-            id={`#afil${index}`}
-            key={item}
-            className={skeleton ? Classes.SKELETON : null}
-          >
+          <div key={item} className={skeleton ? Classes.SKELETON : null}>
```
maniarathi, line 47: Is the `id` not necessary here?

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 
 const renderContributors = (contributors, affiliations, skeleton) => {
   // eslint-disable-next-line no-constant-condition --  Temp removed contributor section to avoid publishing PII
-  if (contributors?.length === 0 && true) return null;
+  if (!contributors || (contributors.length === 0 && true)) return null;
```
maniarathi, line 6: Would you want to continue leaving the `?` in contributors.length? I get confused by undefined vs null but it seems like `!contributors` would catch the null case but not when contributors is undefined?

maniarathi (commented):
I mostly had a couple questions

client/src/components/infoDrawer/infoFormat.js
seve: `!contributors` will be checking for falsely values which includes undefined and null

client/src/components/infoDrawer/infoFormat.js
seve: The id was left over from a previous attempt to reference it from the contributor list

maniarathi (approved):

PR #1832, opened 2020-09-12
Pass in the previous crossfilter when creating a new annomatrix for a switched embedding in order to retain the previous selection of cells.

maniarathi:
Screen Shot 2020-09-11 at 6 31 38 PM
Screen Shot 2020-09-11 at 6 31 17 PM

### DIFF
--- a/client/src/actions/embedding.js
+++ b/client/src/actions/embedding.js
@@ -30,8 +30,7 @@ export const layoutChoiceAction = (newLayoutChoice) => async (
   On layout choice, make sure we have selected all on the previous layout, AND the new
   layout.
   */
-  const { annoMatrix: prevAnnoMatrix } = getState();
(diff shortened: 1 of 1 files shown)


client/src/actions/embedding.js
```diff
@@ -31,8 +31,10 @@ export const layoutChoiceAction = (newLayoutChoice) => async (
   layout.
   */
   const { annoMatrix: prevAnnoMatrix } = getState();
+  const { obsCrossfilter: prevCrossfilter } = getState();
```
bkmartinjr, line 34: nit: you could combine lines 33 & 34 into a single assignment, avoiding the extra call to getState().

bkmartinjr (approved):

client/src/actions/embedding.js
maniarathi: Good call! Done!

PR #1833, opened 2020-09-12
app config bug fix:

bmccandless:
When reading a config file that included per_dataset_config,
the dataroot specializations were applied, but not the default config.
This PR fixes that and also includes a test for this case.


MDunitz (approved):
lgtm

PR #1834, opened 2020-09-14
Fix import of anndata from master so that there aren't issues with scanpy version checking.

maniarathi:
Proof that it works! https://github.com/chanzuckerberg/cellxgene/runs/1114912928?check_suite_focus=true


maniarathi:
Note of course that the cellxgene release X anndata master is legitimately failing now 😆

MDunitz (approved):
lgtm

PR #1836, opened 2020-09-14
dont cache schema

MDunitz:
## Need
- to return a schema containing recently created user annotations when a user logs out and then logs back in (without requiring a hard refresh)

## Approach
- dont cache the schema route

## Future Todos
- separate the user schema and the dataset schema so we can cache the dataset schema


bmccandless (approved):
LGTM

PR #1853, opened 2020-09-17
Info Drawer format adjustments

seve:
This PR tweaks the look and feel of the info drawer in response to QA from @signechambers1


seve:
Updating snapshots to fix smoke-tests

client/src/components/util/truncate.js
```diff
@@ -40,7 +40,7 @@ export default (props) => {
   ) {
     throw Error("Only pass a single child with text to Truncate");
   }
-  const originalString = children.props.children;
+  const originalString = children.props.children.toString();
```
tihuan, line 43: suggest `String(children.props.children)` over `toString()`, since:

`String(undefined) // undefined` while `undefined.toString()` blows up the program and throws a runtime error!

seve: Nice catch but shouldn't
```javascript
  if (
    React.Children.count(children) !== 1 ||
    React.Children.count(children.props?.children) !== 1
  ) {
    throw Error("Only pass a single child with text to Truncate");
  }
```
catch that before it can error?

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -80,20 +82,47 @@ const renderOrganism = (organism, skeleton) => {
...
-          if (!pair[1] || pair[1] === "") return null;
+        {Array.from(singleValueCategories).map((pair, index, array) => {
+          const [category, value] = pair;
+          let ontologyTermID = "";
+          // If the value is empty skip it
+          if (!value || value === "") return null;
+
+          // If the next category is a ontology term, let's save it's value to render now
+          if (array[index + 1]?.[0]?.toString().includes("ontology_term_id"))
+            // ontologyTermID = array[index + 1][1];
```
tihuan, line 99: is this dead code we can remove?

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -80,20 +82,47 @@ const renderOrganism = (organism, skeleton) => {
...
       <UL>
-        {Array.from(singleValueCategories).map((pair) => {
-          if (!pair[1] || pair[1] === "") return null;
+        {Array.from(singleValueCategories).map((pair, index, array) => {
+          const [category, value] = pair;
+          let ontologyTermID = "";
+          // If the value is empty skip it
+          if (!value || value === "") return null;
+
+          // If the next category is a ontology term, let's save it's value to render now
```
tihuan, line 97: what's the rationale behind checking the next item in the current iteration instead of just wait for the next iteration to check it?

client/src/components/infoDrawer/infoFormat.js
seve: Oh! Lol I was actually looking for that comment and couldn't find it! I ended up writing it right below it ahahhaha

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -80,20 +82,47 @@ const renderOrganism = (organism, skeleton) => {
...
+          if (array[index + 1]?.[0]?.toString().includes("ontology_term_id"))
+            // ontologyTermID = array[index + 1][1];
+            ontologyTermID = array[index + 1]?.[1];
+          // If this category is an ontology term skip it since we already used it
+          if (category.toString().includes("ontology_term_id")) {
+            return null;
+          }
+
+          const catWidth = "30%";
```
tihuan, line 106: seems like we can lift these constants out of the function declaration and put them right above the function?

E.g., 

```tsx
const CAT_WIDTH = "30%";
const VAL_WIDTH = "35%";
const renderSingleValueCategories = (singleValueCategories, skeleton) => { ... }
```

client/src/components/infoDrawer/infoFormat.js
```diff
@@ -80,20 +82,47 @@ const renderOrganism = (organism, skeleton) => {
...
+          let ontologyTermID = "";
+          // If the value is empty skip it
+          if (!value || value === "") return null;
+
+          // If the next category is a ontology term, let's save it's value to render now
+          if (array[index + 1]?.[0]?.toString().includes("ontology_term_id"))
+            // ontologyTermID = array[index + 1][1];
+            ontologyTermID = array[index + 1]?.[1];
+          // If this category is an ontology term skip it since we already used it
+          if (category.toString().includes("ontology_term_id")) {
```
tihuan, line 102: tiny nit: Maybe create a constant for `"ontology_term_id"`?

E.g.,

```ts
const ONTOLOGY_TERM_ID = "ontology_term_id";
```

client/src/components/infoDrawer/infoFormat.js
seve: If we tried to handle it on the next iteration we wouldn't be able to iterate of the list using Map since we have to append the value of that checked item to the element we're currently constructing.  AFAIK we don't have access to the resulting array in `Array.prototype.map()`

client/src/components/infoDrawer/infoFormat.js
seve: Good idea, I'll put it right next to the width constants

tihuan (commented):
Thanks for the really great fix, @seve ! Just some tiny comments 😄 👌 PTAL! Thank you!

client/src/components/util/truncate.js
tihuan: Yes! But for consistency, personally I find it easier to always use `String()` and haven't found a need to use `toString()` 😆

client/src/components/infoDrawer/infoFormat.js
tihuan: ooh should we use `.reduce()`?

seve: Oh great idea!

client/src/components/util/truncate.js
tihuan: Oh! Just double checking, did we need to use `toString()` here or it's just not updated yet 😆 ? Thank you!

tihuan (approved):
Looks suuuper good!! Thanks for the update, Seve! Only two tiny questions, otherwise LGTMMM! Happy Friday 😄 🎉 👯 !

PR #1854, opened 2020-09-17
config refactor

MDunitz:
## Need
- to begin to break up the config file to improve readability/testability

## Approach
- broke config file into four separate files/classes (base, app, server and dataset)
- create helper functions for testing the config including the ability to quickly write custom yml config files
- created unit tests for each config class

## Out of scope/intended for future refactoring
- major rewrites/simplifications of the code
- additional documentation/example config files
- testing all possible configuration options


server/cli/launch.py
```diff
@@ -354,7 +352,7 @@ def launch(
     # > cellxgene launch --dataroot <url>
 
     if dump_default_config:
-        print(default_config)
+        print(DEFAULT_CONFIG.default_config)
```
bmccandless, line 355: cellxgene launch --dump-default-config

should print out yaml, so that users can then modify.  
DEFAULT_CONFIG.default_config is a python dictionary.

Maybe we should also have a test for this?

bmccandless (commented):
I made a first pass at the source.  I'll look at tests next.

server/common/config/app_config.py
```diff
@@ -0,0 +1,171 @@
...
+        # attribute of the server_config.
+        self.dataroot_config = {}
+
+        # Set to true when config_completed is called
+        self.is_completed = False
+
+    @staticmethod
+    def get_default_config():
+        try:
+            with open(f"{os.getenv('PROJECT_ROOT')}/default_config.yml", 'r') as default_config:
```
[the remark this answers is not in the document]
MDunitz: Moved the project root definition into the setup.py file
we could also put in in environment.defaults or use a relative path

I prefer to have the config as a separate file at the root so it is easy for users to find. Also if we use a yaml file, it makes it simple for the user to use it as an example to create their own custom config.

server/common/config/app_config.py
```diff
@@ -0,0 +1,171 @@
...
+    presentations of a dataset.
+    The dataset config attributes can be overridden depending on the url by which the
+    dataset was accessed.  These are stored in dataroot_config.
+    AppConfig has methods to initialize, modify, and access the configuration.
+    """
+
+    def __init__(self):
+
+        # the default configuration (see default_config.py)
+        ## TODO @madison -- if we always read from the default config (hard coded path) can we set those values as defaults within the config class?
```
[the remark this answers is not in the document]
MDunitz: I meant instead of using the default config file we could just set the default vars. We might need to rewrite the dump function but I think it would simplify quite a bit. 
More of a future optimization...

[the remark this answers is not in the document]
bmccandless: Okay, as long as we maintain the comments that are currently in the config file, and probably also want to maintain the order that the config parameters appear in the file.

server/common/config/app_config.py
```diff
@@ -0,0 +1,172 @@
...
+    presentations of a dataset.
+    The dataset config attributes can be overridden depending on the url by which the
+    dataset was accessed.  These are stored in dataroot_config.
+    AppConfig has methods to initialize, modify, and access the configuration.
+    """
+
+    def __init__(self):
+
+        # the default configuration (see default_config.py)
+        # TODO @madison -- if we always read from the default config (hard coded path) can we set those values as
```
maniarathi, line 23: I like this idea - @bmccandless is there a reason why the hardcoded default config values can't be set as default values in code?

server/common/config/app_config.py
```diff
@@ -0,0 +1,172 @@
...
+        #  defaults within the config class?
+        self.default_config = self.get_default_config()
+        # the server configuration
+        self.server_config = ServerConfig(self, self.default_config["server"])
+        # the dataset config, unless overridden by an entry in dataroot_config
+        self.default_dataset_config = DatasetConfig(None, self, self.default_config["dataset"])
+        # a dictionary of keys to DatasetConfig objects.  Each key must exist in the multi_dataset__dataroot
```
maniarathi, line 30: Can you add a statement here on how the DatasetConfig objects in the dictionary may or may not override the default_dataset_config that you just initialized?

server/common/config/app_config.py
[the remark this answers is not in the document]
bmccandless: We already have a solution for helping a user get started with a config file:
```
  cellxgene launch --dump-default-config > myconfig.py
```

server/common/config/app_config.py
```diff
@@ -0,0 +1,172 @@
...
+
+        # Set to true when config_completed is called
+        self.is_completed = False
+
+    @staticmethod
+    def get_default_config():
+        try:
+            with open(f"{os.getenv('PROJECT_ROOT')}/default_config.yml", 'r') as default_config:
+                return yaml.safe_load(default_config)
+        except yaml.YAMLError as e:
```
maniarathi, line 42: Maybe also toss in a `except OSError as err:` to catch the case that the file doesn't exist and explain that it wasn't able to find the file?

server/common/config/app_config.py
```diff
@@ -0,0 +1,172 @@
...
+            raise ConfigurationError(f"Unable to import default configuration yaml: {e}")
+
+    def get_dataset_config(self, dataroot_key):
+        if self.server_config.single_dataset__datapath:
+            return self.default_dataset_config
+        else:
+            return self.dataroot_config.get(dataroot_key, self.default_dataset_config)
+
+    def check_config(self):
+        """Verify all the attributes have been checked"""
```
maniarathi, line 52: What does it mean for an attribute to have been "checked"? Maybe add some clarity here in the docstring?

server/common/config/base_config.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class BaseConfig(object):
+    """This class handles the mechanics of updating and checking attributes.
+    Derived classes are expected to store the actual attributes"""
+
+    def __init__(self, app_config, default_config, dictval_cases={}):
+        # reference back to the app_config
+        self.app_config = app_config
+        # the complete set of attribute and their default values (unflattened)
```
maniarathi, line 14:
```suggestion
        # the complete set of attributes and their default values (unflattened)
```

server/common/config/base_config.py
```diff
@@ -0,0 +1,104 @@
...
+from flatten_dict import flatten
+from server.common.errors import ConfigurationError
+
+
+class BaseConfig(object):
+    """This class handles the mechanics of updating and checking attributes.
+    Derived classes are expected to store the actual attributes"""
+
+    def __init__(self, app_config, default_config, dictval_cases={}):
+        # reference back to the app_config
```
maniarathi, line 12: Can you explain here what the relationship of BaseConfig and AppConfig is here? AppConfig contains BaseConfig?

server/common/config/base_config.py
```diff
@@ -0,0 +1,104 @@
...
+        self.app_config = app_config
+        # the complete set of attribute and their default values (unflattened)
+        self.default_config = default_config
+        # attributes where the value may be a dict (and therefore are not flattened)
+        self.dictval_cases = dictval_cases
```
maniarathi, line 17: Why does this need to be a separate member variable? During flattening time, could you not just check one variable and based on the type, triage whether it needs to be flattened or not. The the flattening logic isn't sprawled beyond the flattening function itself.

server/common/config/base_config.py
```diff
@@ -0,0 +1,104 @@
...
+
+        flat_config = flatten(config_copy)
+        for key, value in flat_config.items():
+            # name of the attribute
+            attr = "__".join(key)
+            mapping[attr] = (key, value)
+
+        return mapping
+
+    def check_attr(self, attrname, vtype):
```
maniarathi, line 47: I'd rename this to be more descriptive of what the check actually is: `validate_correct_type_of_configuration_attribute`. I generally don't shy away from using long names because a) they disambiguate the function and b) we all use IDEs anyways so we rarely need to type out the full function name.

server/common/config/base_config.py
```diff
@@ -0,0 +1,104 @@
...
+        self.app_config = app_config
+        # the complete set of attribute and their default values (unflattened)
+        self.default_config = default_config
+        # attributes where the value may be a dict (and therefore are not flattened)
+        self.dictval_cases = dictval_cases
+        # used to make sure every attribute value is checked
+        self.attr_checked = {key_name: False for key_name in self.create_mapping(default_config).keys()}
+
+    def create_mapping(self, config):
+        """Create a mapping from attribute names to (location in the config tree, value)"""
```
maniarathi, line 22: I actually have no idea what this is doing, still. What is "location in the config tree". Why is the config a tree and what do the nodes in the tree represent?

server/common/config/client_config.py
```diff
@@ -0,0 +1,125 @@
...
+    if dataset_config.app__authentication_enable and auth.is_valid_authentication_type():
+        userinfo = {}
+        userinfo["userinfo"] = {
+            "is_authenticated": auth.is_user_authenticated(),
+            "username": auth.get_user_name(),
+            "user_id": auth.get_user_id(),
+            "email": auth.get_user_email()
+        }
+        return userinfo
+    else:
```
maniarathi, line 124: Don't need an `else` here. the `return` from the if statement guarantees that this code path never hit if the `if` condition is met.

server/common/config/dataset_config.py
```diff
@@ -0,0 +1,214 @@
...
+        self.handle_diffexp(context)
+
+    def handle_app(self):
+        self.check_attr("app__scripts", list)
+        self.check_attr("app__inline_scripts", list)
+        self.check_attr("app__about_legal_tos", (type(None), str))
+        self.check_attr("app__about_legal_privacy", (type(None), str))
+        self.check_attr("app__authentication_enable", bool)
+
+        # scripts can be string (filename) or dict (attributes).   Convert string to dict.
```
maniarathi, line 65:
```suggestion
        # scripts can be string (filename) or dict (attributes). Convert string to dict.
```

maniarathi: Sorry, my OCD kicked in.

server/common/config/dataset_config.py
```diff
@@ -0,0 +1,214 @@
...
+        scripts = []
+        for script in self.app__scripts:
+            try:
+                if isinstance(script, str):
+                    scripts.append({"src": script})
+                elif isinstance(script, dict) and isinstance(script["src"], str):
+                    scripts.append(script)
+                else:
+                    raise Exception
```
maniarathi, line 74: Why throw the error later? Can you just raise the ConfigurationError here?

server/common/config/dataset_config.py
```diff
@@ -0,0 +1,214 @@
...
+from server.common.errors import ConfigurationError, OntologyLoadFailure
+from server.compute.scanpy import get_scanpy_module
+from server.data_common.matrix_loader import MatrixDataLoader, MatrixDataType
+from server.db.db_utils import DbUtils
+
+
+class DatasetConfig(BaseConfig):
+    """Manages the config attribute associated with a dataset."""
+
+    def __init__(self, tag, app_config, default_config):
```
maniarathi, line 16: Can you add types here? So something like `def __init__(self, tag: str, app_config: AppConfig, default_config: BaseConfig):`

(I'm just guessing at what these might be..)

server/common/config/app_config.py
MDunitz: created a ticket --  https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1880

server/data_common/data_adaptor.py
```diff
@@ -388,3 +388,18 @@ def get_last_mod_time(self):
...
             lastmod = None
         return lastmod
+
+
+class AppFeature(object):
+    def __init__(self, path, available=False, method="POST", extra={}):
+        self.path = path
+        self.available = available
+        self.method = method
+        self.extra = extra
```
maniarathi, line 398: Tbh, all 4 variable names mean absolutely nothing to me. Is there a way these names can be more descriptive?

server/data_common/data_adaptor.py
```diff
@@ -388,3 +388,18 @@ def get_last_mod_time(self):
...
+        for k, v in extra.items():
+            setattr(self, k, v)
```
maniarathi, line 400:
```suggestion
        [setattr(self, key, value) for key, value in extra.items()]
```
List comprehension to make it more Pythonic. Also I'd add more descriptive variable names than `key` and `value` once I understand what is actually in `extra`.

server/test/fixtures/invalid_config.yml
```diff
@@ -0,0 +1,204 @@
...
+    verbose: false
+    debug: false
+    host: localhost
+    port : null
+    open_browser: false
+    force_https: false
+    flask_secret_key: null
+    generate_cache_control_headers: false
+    server_timing_headers: false
+#    csp_directives: null  ## should throw an error
```
maniarathi, line 13: It's super difficult to understand what in this config is wrong. Is it possible to simply have the error itself only and layer it on the default config to generate the error?

server/test/unit/common/config/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+                             client_secret="null", jwt_decode_options="null", session_cookie="true", cookie="null",
+                             dataroot="null", index="false", allowed_matrix_types=[], max_cached_datasets=5,
+                             timelimit_s=5, dataset_datapath="null", obs_names="null", var_names="null", about="null",
+                             title="null", diffexp_max_workers=64, cpu_multiplier=4, target_workunit="16_000_000",
+                             data_locater_region_name="us-east-1", cxg_tile_cache_size=8589934592,
+                             cxg_num_reader_threads=32, anndata_backed="false", column_request_max=32,
+                             diffexp_cellcount_max="null", config_file_name="server_config.yaml"):
+        configfile = os.path.join(self.tmp_fixtures_directory, config_file_name)
+        with open(configfile, "w") as server_config:
+            config = f"""server:
```
maniarathi, line 38: Can this be a fixture file that is read?

server/test/unit/common/config/__init__.py
```diff
@@ -0,0 +1,218 @@
...
+    def custom_dataset_config(self, scripts=[], inline_scripts=[], about_legal_tos="null", about_legal_privacy="null",
+                              authentication_enable="true", max_categories=1000, custom_colors="true",
+                              enable_users_annotations="true", annotation_type="local_file_csv", db_uri="null",
+                              hosted_file_directory="null", local_file_csv_directory="null", local_file_csv_file="null",
+                              ontology_enabled="false", obo_location="null", embedding_names=[],
+                              enable_reembedding="false", enable_difexp="true", lfc_cutoff=0.01, top_n=10,
+                              config_file_name="dataset_config.yml"):
+        configfile = os.path.join(self.tmp_fixtures_directory, config_file_name)
+        with open(configfile, "w") as dataset_config:
+            config = f"""
```
maniarathi, line 179: Can this be a fixture file that is read?

server/common/config/dataset_config.py
```diff
@@ -0,0 +1,217 @@
...
+                )
+                if matrix_data_loader.matrix_data_type != MatrixDataType.H5AD:
+                    raise ConfigurationError("enable-reembedding is only supported with H5AD files.")
+                if server_config.adaptor__anndata_adaptor__backed:
+                    raise ConfigurationError("enable-reembedding is not supported when run in --backed mode.")
+
+            try:
+                get_scanpy_module()
+            except NotImplementedError:
+                # Todo why isnt this in the reqs file/why is this check necessary
```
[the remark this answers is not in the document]
maniarathi: We initially didn't require scanpy because it used to be a pretty flaky module. At this point though, scanpy is pretty solid so there really isn't much of a reason not to include it, even if we haven't fully supported re-embeddings. It's just one more thing that we're putting on the shoulders of the users at this point.

maniarathi (commented):
Pass 1

server/test/fixtures/invalid_config.yml
bmccandless: Yeah, it is not an error have missing config parameters.  A user's config file need only have the parameters they want to change from the defaults.

server/common/config/app_config.py
bmccandless: That ideas fine if we remove the default config file, in fact it would be basically required.

But I'm unsure what the need or benefit is of removing the config file
The config file servers a few purposes.
1.  a file to document the attributes
2. a file to see the whole structure in one place 
3. a file to provide an order for the attributes
5. a file, whose contents we can present when "cellxgene launch --dump-default-config" (related to 3).
4. a file that is versioned, so you can easily see what changes PRs

We can find workarounds for all these, but that would just move complexity to other places.  
I don't think there is a good return on investment.

If we want to keep improving the app config, 
here are a few items that I think have more value:
1.  better handling for aws secrets 
2. consolidate all the single_dataset testing that in spread around the app config code into one place
3. better documentation for the config parameters

bmccandless: two more idea for improvement

4.  The ability to check that a config file is valid: e.g. `cellxgene launch --check-config config.yaml`
5. The ability to pass in more than one config file.  This would be super useful for the hosted cellxgene deployments.

I think @MDunitz has made an issue for the 4th one.  I can make issues for the others.

server/common/config/base_config.py
bmccandless: AppConfig contains one or more DatasetConfig and one ServerConfig.   DatasetConfig and ServerConfig both inherit from BaseConfig.

server/common/config/app_config.py
MDunitz: I think 2 and 3 would be really helpful. I've already created a ticket for 3 https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1879

server/test/fixtures/invalid_config.yml
MDunitz: removed this file for clarity, instead I'll just generate an invalid config by passing in a var of the wrong type

server/common/config/dataset_config.py
MDunitz: Because AppConfig creates the DatasetConfig/ServerConfig objects adding types seems to lead to circular import errors (even when done locally)

server/common/config/base_config.py
MDunitz: discussed offline

server/data_common/data_adaptor.py
bmccandless: This looks like some old cruft.  I think everything about "extra" in the AppFeature class can be removed.
I checked and it is not set by the server.  
The AppFeature are used to communicate which endpoints are enabled to the client through the /config:
For example (some output from the config endpoint):
```
    "features": [
      {
        "available": false,
        "method": "POST",
        "path": "/cluster/"
      },
      {
        "available": true,
        "method": "GET",
        "path": "/layout/obs"
      },
```
"extra" was a way to add more key/values to that json blob.  I think diffexp might have set something here at one point.
This kind of pre-dates me.  But in any case we can remove "extra"

Also, a quick check of the client source, and it does not appear that the client even looks at "features".
I think we can safely remove the whole concept.

(As a side note, there are plenty of ways to improve our API)

server/common/config/base_config.py
bmccandless: Yes, the comment definitely can be greatly improved on.

The purpose of this function is to create a dictionary that looks like this:

The keys are the name of attributes, using the double underscore convention.
For example:  authentication__type.

The values are a tuple. 
The first item of the tuple is a tuple of path elements.   I'm calling this a "location in the config tree" in the comment.
The second item of the tuple is the value of the config parameter.
For example:   (('authentication', 'type'), 'session'))

So, basically the keys are the attribute names.  
The values says where in config hierarchy you'd find that config parameter, and it's current value.

server/common/config/dataset_config.py
MDunitz: Tried refactoring, but if script is a dict that doesnt contain `src` it will throw a key error which wont be caught

server/data_common/data_adaptor.py
MDunitz: I'm going to remove all of the code related to features in a separate pr just to make the diff clear if anything breaks.

server/data_common/data_adaptor.py
MDunitz: it sounds like this code isnt being used by the client at all. I'm going to remove all of the code related to it in a separate pr just to make the diff clear if anything breaks.

server/common/config/base_config.py
MDunitz: The whole config is passed in as a dict of dicts. Most of the time the config value is a a string, but when the value is intended to be a dict we dont want to flatten it and turn the dict keys into the path. @bmccandless does this align with your understanding?

server/test/unit/common/config/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+from unittest import mock
+
+from server.test import FIXTURES_ROOT
+
+
+def mockenv(**envvars):
+    return mock.patch.dict(os.environ, envvars)
+
+
+class ConfigTests(unittest.TestCase):
```
maniarathi, line 14: There aren't any tests in this class are there? I see helper functions to return configs but not really any assertions?

The equivalent file has the function `handle_config_from_secret`. It might be nice to test that while mocking out AWS SecretsManager?

server/common/config/base_config.py
bmccandless: Yes, that's right.   These attributes are generally for interfacing to external modules.  One example is tiledb context parameters.   We don't processes them in cellxgene, but rather pass them (as a dictionary of key/value) to a tiledb interface.

server/test/fixtures/valid_default_config.yml
```diff
@@ -0,0 +1,203 @@
+server:
```
bmccandless, line 1: This file doesn't appear to be used.   Can it be removed?

server/test/unit/cli/test_launch.py
```diff
@@ -0,0 +1,29 @@
...
+    @classmethod
+    def tearDownClass(cls) -> None:
+        shutil.rmtree(cls.tmp_dir)
+
+
+def test_dump_default_config(self):
+    os.system(f"cellxgene launch --dump-default-config > {self.tmp_dir}/test_config_dump.txt")
+    config = AppConfig().default_config
+    with open(f"{self.tmp_dir}/expected_config_dump.txt", "w") as expected_config:
+        expected_config.write(yaml.dump(config))
```
bmccandless, line 28: I expect that the output of dump default config should be identical to the config file (comments, ordering of parameters).

server/test/unit/common/config/__init__.py
MDunitz: I have a test for `handle_config_from_secret` in `test_server_config` https://github.com/chanzuckerberg/cellxgene/pull/1854/files#diff-4cf6f8eb2af82657133e9e9cf28dedccR297
I can move it into the init file if you think that would be more clear. 

I typically think of the init file for test directories as a place to store helper functions shared across many tests. Do you prefer to store those elsewhere?

bmccandless (changes requested):
Looks pretty good.
Two things to change:
 1.  a reliable way to find the default_config.yml, or revert back to the current way.
 2. dump-default-confg should preserve comments and parameters order.

server/cli/launch.py
```diff
@@ -354,7 +353,7 @@ def launch(
     # > cellxgene launch --dataroot <url>
 
     if dump_default_config:
-        print(default_config)
+        print(yaml.dump(DEFAULT_CONFIG.default_config))
```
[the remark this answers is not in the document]
MDunitz: I thought it would be better to just print the config vars without any cruft. So it is a clean example file?

maniarathi (commented):
Nit comment: did you run `black` on this for formatting purposes. There seems to be a few weirdnesses around formatting.

server/test/unit/common/config/test_app_config.py
```diff
@@ -0,0 +1,144 @@
...
+
+            app_config = AppConfig()
+            app_config.update_from_config_file(configfile)
+            server_changes = app_config.server_config.changes_from_default()
+            dataset_changes = app_config.default_dataset_config.changes_from_default()
+            self.assertEqual(server_changes, [("multi_dataset__dataroot", "test_dataroot", None)])
+            self.assertEqual(dataset_changes, [])
+
+    def test_configfile_no_server_section(self):
+        # test a config file without a dataset section
```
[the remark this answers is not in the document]
bmccandless: Every parameters in the config file is optional.  If you don't specify it, then you get the default value.
So you can write a really minimal config file.  
e.g. this is a totally valid config file:  equivalent to `cellxgene launch --dataroot directory`

```
server:
   multi_dataset:
      dataroot:  <directory>
```

The second one is also a valid test file, but you couldn't run cellxgene with just this config file.
You'd need to pass in a command line argument specifying the datapath or dataroot to actually run.
But this test is a testing for valid config files, not complete configuration.

```
dataset:
    user_annotations:
         enable: false
```

server/test/unit/common/config/test_app_config.py
```diff
@@ -0,0 +1,144 @@
...
+        config = AppConfig()
+        config.update_from_config_file(file_name)
+        self.assertEqual(config.server_config.single_dataset__datapath, datapath)
+
+    def test_update_server_config_updates_server_config_and_config_status(self):
+        config = self.get_config()
+        config.complete_config()
+        config.check_config()
+        config.update_server_config(multi_dataset__dataroot=FIXTURES_ROOT)
+        with self.assertRaises(ConfigurationError):
```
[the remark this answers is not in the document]
MDunitz: The config error is because when you update the config the attribute has no longer been type checked and you need to call complete_config before check_config will be successful

server/cli/launch.py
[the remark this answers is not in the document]
bmccandless: "cruft" may be in the eye of the beholder :)
The comments are the documentation for the config file.
It's meant to tell the user what each config parameter is used for.
So far that's our only documentation for the config file -- unless the user looks through our git repo or the cellxgene distribution.
Maybe we can change that if we had some online documentation to describe the parameters.
But even then it might be useful to have a comment in the yaml that says "see http://... for documentation on this file"

The other issue I raised was ordering of the parameters.   I chose the ordering intentionally to have the more important ones first.   The behavior here is to alphabetize them.  Also, there are no blank lines separating sections.

server/__init__.py
```diff
@@ -12,3 +12,10 @@
     # Make sure to exit in this case, as the server may not be configured as expected.
     logging.critical(f"Error in import_plugins: {str(e)}")
     sys.exit(1)
+
+try:
+    PROJECT_ROOT = os.popen("git rev-parse --show-toplevel").read().strip()
```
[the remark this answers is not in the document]
MDunitz: alright I give up, I'm going to revert back to the python string with yaml.load

server/cli/launch.py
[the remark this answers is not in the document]
MDunitz: ok that makes sense. I was thinking we just wanted to give the user a clean example they could edit but if its the only way they can access the config documentation its probably better to print it all out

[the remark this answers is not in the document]
MDunitz: But also we should prioritize moving the config documentation

default_config.yml
```diff
@@ -1,6 +1,3 @@
-import yaml
-
-default_config = """
```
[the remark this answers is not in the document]
MDunitz: it looks like importlib has some issues with 3.4-3.6. I'm going to revert back to the current approach

MDunitz:
> Nit comment: did you run `black` on this for formatting purposes. There seems to be a few weirdnesses around formatting.

@maniarathi I just ran black on the repo but it seems to have added quite a few other formatting changes

bmccandless (approved):
LGTM!

server/test/unit/common/config/__init__.py
maniarathi: Ah I was just confused since it inherits `unittest.TestCase`

maniarathi (approved):
Amazing work, Madison!

PR #1857, opened 2020-09-18
Bug reading the config file.

bmccandless:
The config file had a bug where it expected both a "server" and "dataset" section.
If one didn't exist, then it would raise an exception.
It should use the default server config or the defaul dataset config in those cases.
Added a test case that would have caught this.


MDunitz (approved):
lgtm

PR #1871, opened 2020-09-24
Make sure there are more than 1 values in a category before rendering it

### LINKED ISSUE #1869: Left side bar displays all metadata first and then removes singletons from the rendering
When a dataset is loading in cellxgene, all of the singleton metadata first shows up in the left side bar and then disappears after a filtering step removes the singletons. Instead the singletons should never store the left side bar.

A possible mechanism to fix this is to store the singletons separately in the annomatrix.

maniarathi:
I also ran prettier which automatically changed `embedding.js`. There are no functional changes there.

Closes #1869


client/src/components/categorical/index.js
```diff
@@ -197,7 +198,8 @@ class Categories extends React.Component {
         )}
         {/* WRITEABLE FIELDS */}
         {allCategoryNames.map((catName) =>
-          schema.annotations.obsByName[catName].writable ? (
+          schema.annotations.obsByName[catName].writable &&
+          schema.annotations.obsByName[catName].categories.length > 1 ? (
```
seve, line 202:
```suggestion
          schema.annotations.obsByName[catName].writable ? (
```

We still want to display single value user annos in the left sidebar

client/src/components/categorical/index.js
```diff
@@ -185,7 +185,8 @@ class Categories extends React.Component {
         {/* READ ONLY CATEGORICAL FIELDS */}
         {/* this is duplicative but flat, could be abstracted */}
         {allCategoryNames.map((catName) =>
-          !schema.annotations.obsByName[catName].writable ? (
+          !schema.annotations.obsByName[catName].writable &&
+          schema.annotations.obsByName[catName].categories.length > 1 ? (
```
seve, line 189: Categories might not be defined for the category

```suggestion
          !schema.annotations.obsByName[catName].writable &&
          schema.annotations.obsByName[catName].categories?.length > 1 ? (
```

seve (changes requested):

client/src/components/categorical/index.js
seve: Actually my fix is not correct, we want categories with undefined `categories` values to pass through to render:

```suggestion
          !schema.annotations.obsByName[catName].writable &&
          (schema.annotations.obsByName[catName].categories?.length > 1 ||
            !schema.annotations.obsByName[catName].categories) ? (
```

maniarathi:
Addressed the suggested changes!

maniarathi:
I also tested these changes on a 4M dataset to ensure that it still worked.

seve (approved):
LGTM!

Issue #1876, opened 2020-09-25
[BUG]  When the user logs out they should stay on the page.


bmccandless:
This was a request from @ambrosejcarr and others.

signechambers1:
Confirmed this is fixed on dev, i stay on the same page when i log out. Thanks @bmccandless!

PR #1877, opened 2020-09-25
auth:  logging out should keep the user on the same page

bmccandless:
previous behavior is that logout would redirect to the index page.

 #1876


MDunitz (approved):
lgtm

maniarathi (approved):

PR #1882, opened 2020-09-28
Fix license years and add CZI

maniarathi:


LICENSE.txt
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013
+Copyright (c) 2013-2020 Chan Zuckerberg Initiative
```
mckinsel, line 3: I think we started working on cellxgene in 2017.

maniarathi: Done!

mckinsel (approved):

PR #1893, opened 2020-09-29
remove AppFeature and all references to it in the code/tests

MDunitz:


seve (approved):
LGTM! Front end makes no use of these values on the config

bmccandless (approved):
LGTM

PR #1898, opened 2020-09-30
1807 AuthN smoke test

tihuan:
This PR does the following:

1. Add `login` and `logout` helper functions in `client/__tests__/e2e/cellxgeneActions.js`
2. Add conditional AuthN integration test in `client/__tests__/e2e/e2e.test.js`. The test will only run if env variable `TEST_AUTH_INTEGRATION` is `"true"`, which is only set in `single-cell-infra`'s Github Action flow. Corresponding PR [here](https://github.com/chanzuckerberg/single-cell-infra/pull/198)

### DIFF
--- a/client/__tests__/e2e/cellxgeneActions.js
+++ b/client/__tests__/e2e/cellxgeneActions.js
@@ -13,7 +13,10 @@ import {
   getTestClass,
   getTestId,
   isElementPresent,
+  goToPage,
 } from "./puppeteerUtils";
+
+import { appUrlBase } from "./config";
 
 export async function drag(testId, start, end, lasso = false) {
   const layout = await waitByID(testId);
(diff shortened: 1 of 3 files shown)


client/__tests__/e2e/e2e.test.js
```diff
@@ -518,4 +520,20 @@ test("lasso moves after pan", async () => {
 
   expect(panCount).toBe(initialCount);
 });
+
+describe("AuthN Integration", () => {
+  if (process.env.TEST_AUTH_INTEGRATION !== "true") {
```
tihuan, line 525: if `TEST_AUTH_INTEGRATION` is not `"true"`, `it.only()` will be triggered and skip all the tests in this `describe` block

client/__tests__/e2e/puppeteer.setup.js
```diff
@@ -17,7 +17,9 @@ setDefaultOptions({ timeout: 20 * 1000 });
 
 jest.retryTimes(ENV_DEFAULT.RETRY_ATTEMPTS);
 
-(async () => {
+beforeEach(async () => {
+  await jestPuppeteer.resetBrowser();
```
tihuan, line 21: Adds `resetBrowser()`, so Auth0 session is not persisted from one test to another, which breaks Jest auto-retries and other tests that rely on logging in before assertion

client/__tests__/e2e/e2e.test.js
```diff
@@ -518,4 +520,18 @@ test("lasso moves after pan", async () => {
 
   expect(panCount).toBe(initialCount);
 });
+
+const conditionalDescribe =
```
tihuan, line 524: this `conditionalDescribe` will skip the describe block if `TEST_AUTH_INTEGRATION` is NOT `"true"`, which is needed since we only want to run the integration tests in `single-cell-infra` GH flow!

client/__tests__/e2e/cellxgeneActions.js
```diff
@@ -312,4 +315,75 @@ export async function assertCategoryDoesNotExist(categoryName) {
 
   await expect(result).toBe(false);
 }
+
+export async function login() {
+  const email = `cellxgene-smoke-test+${process.env.DEPLOYMENT_STAGE}@chanzuckerberg.com`;
+  const password = "Test1111";
```
seve, line 321: We just have to make sure the permissions for these test accounts are barebones since anyone who looks at this file will have access to the credentials

seve (approved):
LGTM!

client/__tests__/e2e/cellxgeneActions.js
Bento007: we only have one permissions state and that is logged in or not. We can create more in future. This test doesn't run on prod does it?

tihuan: oh the test is intended to run on prod too! But I haven't created a prod Auth0 account yet, tentatively `cellxgene-smoke-test+prod@chanzuckerberg.com`, because I wanna wait till prod deploy and see if https://cellxgene.cziscience.com/experimental/krasnow_lab_human_lung_cell_atlas_10x-1.cxg becomes available!

tihuan: Since the infra PR is not merged yet, the integration test won't run for all envs at the moment: https://github.com/chanzuckerberg/single-cell-infra/pull/198/files

tihuan:
re-opening this ticket since Zenhub closed this on accident due to me manually linking the PR to the issue

PR #1903, opened 2020-10-05
Create design_principles.md

colinmegill:


ambrosejcarr (approved):
This looks great. Two suggested edits, and if you agree with them, then I think this is good to merge. Thanks for writing this up.

Issue #1859, opened 2020-09-22
Improve how the secret manager is used in cellxgene


bmccandless:
The secret manager is currently used in a limited way:
  - only used to overwrite a few pre-selected configuration parameters:  
      flask_secret_key, the oauth client secret, and the database uri.
      these parameters are in the source code (so it's sort of opaque how this works).
      It requites the keys in the secret to have a pre-defined name.
  - only used in the hosted eb server.  
  - The name of the secret is passed as an environment variable.
  
A refactoring could enable
  - any config parameter to be overwritten
  - the location of the secret, the keys it looks for, and the config parameters they set could all be in the config file.  
    This makes it very clear how the parameters are being set.
  - could pull config from more than one secret.
  - this could make it easier to test a configuration before it is deployed.
     https://github.com/chanzuckerberg/single-cell/issues/63

The following is a possible proposal that meets the above requirements.

Within the config file, add a section for the secret manager.
Map between key/values in the secret and config parameters.
Here the "!param" is making use of yaml custom tags.
The loader will combine the param sequence to define the location of a config param:

```
aws_secret_manager:   # section to describe the secrets
   secrets_shared_with_portal:   # name of the secret (I just made this secret name up).
      region:  us-west-2  # region of the secret (if needed)
      values:  
          # map secret keys to cellxgene config parameters
          # It will find the value in the secret at the give key, and use it to set the give config param, given its path.
          client_secret : !param [ "server", "authentication", "params_oauth", "client_secret"]
          client_id : !param [ "server", "authentication", "params_oauth", "client_id"]
   
   secrets_cellxgene_dev:  # another example, showing we can grab from more than one secret
      region: us-west-2
      values:
         db _uri:  !param [ "dataset", "user_annotations", "hosted_tiledb_array", "db_uri" ]
```

With this plan, everything is explicit.  We can get rid of the environment variables, and make 
the secret handling, which is currently embedded into code, be part of the configuration.

bmccandless:
This is done.

PR #1904, opened 2020-10-05
Enhance the AppConfig with external config sources.

bmccandless:
The external config sources are currently environment variables
and AWS secrets manager.

The config file can be augmented with a section describing how
environmen variables and secrets can update config parameters.

benefits:
 - it will enable the config to draw from more than one secret.  This is useful
   for shared secrets between cellxgene and data portal, as well as auth0 secrets.
 - it will make it very straightforward to check the config before a deployment.

 Part of #1859


server/test/unit/common/config/test_external_config.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(ExternalConfig.convert_value("false"), False)
+        self.assertEqual(ExternalConfig.convert_value("False"), False)
+        self.assertEqual(ExternalConfig.convert_value("False"), False)
+        self.assertEqual(ExternalConfig.convert_value("null"), None)
+        self.assertEqual(ExternalConfig.convert_value("None"), None)
+        self.assertEqual(ExternalConfig.convert_value("{'a':10, 'b':'string'}"), dict(a=int(10), b="string"))
+
+    def test_environment_variable(self):
+        with tempfile.TemporaryDirectory() as tempdir:
+            configfile = os.path.join(tempdir, "config.yaml")
```
MDunitz, line 32: Could you create a helper method for creating an external config for testing and add it to the others in config/__init__?

server/common/config/app_config.py
```diff
@@ -73,6 +78,47 @@ def update_default_dataset_config(self, **kw):
             value.update(**kw)
         self.is_complete = False
 
+    def update_single_config_from_path_and_value(self, path, value):
+        """Update a single config parameter with the value.
+        path is a list of string, that gives a path to the config parameter to be udpated.
```
MDunitz, line 83:
```suggestion
        path is a list of string, that gives a path to the config parameter to be updated.
```

server/common/config/external_config.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            self.environment = default_config["environment"]
+            self.aws_secrets_manager__region = default_config["aws_secrets_manager"]["region"]
+            self.aws_secrets_manager__secrets = default_config["aws_secrets_manager"]["secrets"]
+
+        except KeyError as e:
+            raise ConfigurationError(f"Unexpected config: {str(e)}")
+
+    @staticmethod
+    def convert_value(value: str):
```
maniarathi, line 24: To keep this class as clean as possible, I'd move this to the `util` directory. Probably `type_conversion_utils.py` seems like a good fit.

maniarathi: Also should this be hardened to ensure no issues with capitalization (TRUE or FALSE)?

server/common/config/external_config.py
```diff
@@ -0,0 +1,103 @@
...
+        """convert the external value (a string), to the most appropriate type"""
+        if value == "true":
+            return True
+        if value == "false":
+            return False
+        if value == "null":
+            return None
+        try:
+            return eval(value)
```
maniarathi, line 33: What case is this catching?

server/common/config/external_config.py
```diff
@@ -0,0 +1,103 @@
...
+        except (NameError, SyntaxError):
+            pass
+
+        return value
+
+    def complete_config(self, context):
+        self.handle_environment(context)
+        self.handle_aws_secrets_manager(context)
+
+    def handle_environment(self, context):
```
maniarathi, line 43: Add docstring please. I think this function is complex enough that a brief docstring would be helpful.

server/common/config/external_config.py
```diff
@@ -0,0 +1,103 @@
...
+                    raise ConfigurationError(f"required environment variable '{name}' not set")
+            else:
+                value = self.convert_value(value)
+                self.app_config.update_single_config_from_path_and_value(path, value)
+
+    def handle_aws_secrets_manager(self, context):
+        self.validate_correct_type_of_configuration_attribute("aws_secrets_manager__region", (type(None), str))
+        self.validate_correct_type_of_configuration_attribute("aws_secrets_manager__secrets", list)
+
+        if len(self.aws_secrets_manager__secrets) == 0:
```
maniarathi, line 68:
```suggestion
        if not len(self.aws_secrets_manager__secrets):
```
Generally more Pythonic.

server/common/config/external_config.py
```diff
@@ -0,0 +1,103 @@
...
+            return
+
+        self.validate_correct_type_of_configuration_attribute("aws_secrets_manager__region", str)
```
maniarathi, line 71: Duplicate of line 65?

server/test/unit/common/config/test_external_config.py
```diff
@@ -0,0 +1,107 @@
...
+
+import requests
+
+from server.common.config.app_config import ExternalConfig
+from server.common.config.app_config import AppConfig
+from server.test import test_server, FIXTURES_ROOT
+
+
+class TestExternalConfig(unittest.TestCase):
+    def test_type_convert(self):
```
maniarathi, line 14: Add a negative test case? What circumstances under which will this function fail?

server/test/unit/common/config/test_external_config.py
```diff
@@ -0,0 +1,107 @@
...
+            with test_server(command_line_args=["-c", configfile], env=env) as server:
+                session = requests.Session()
+                response = session.get(f"{server}/api/v0.2/config")
+                data_config = response.json()
+                self.assertEqual(
+                    data_config["config"]["displayNames"]["dataset"], "a95c59b4-7f5d-4b80-ad53-a694834ca18b"
+                )
+                self.assertFalse(data_config["config"]["parameters"]["disable-diffexp"])
+
+    @patch("server.common.config.external_config.get_secret_key")
```
maniarathi, line 67: An option - you could setup mocked AWS secrets manager using moto.

server/test/unit/common/config/test_external_config.py
```diff
@@ -0,0 +1,107 @@
...
+            app_config.server_config.app__flask_secret_key = "original"
+            app_config.server_config.single_dataset__datapath = f"{FIXTURES_ROOT}/pbmc3k.cxg"
+
+            app_config.complete_config()
+
+            self.assertEqual(app_config.server_config.app__flask_secret_key, "original")
+            self.assertEqual(app_config.server_config.authentication__params_oauth__client_secret, "mock_oauth_secret")
+            self.assertEqual(
+                app_config.default_dataset_config.user_annotations__hosted_tiledb_array__db_uri, "mock_db_uri"
+            )
```
maniarathi, line 107: I think this test suite is missing a bunch of negative cases. There's a good chunk of logic in `ExternalConfig` around ensuring that wrong config errors are raised and I'd like to see test cases that cover those to ensure that we're not letting through configs that we expect to raise errors.

server/common/config/external_config.py
bmccandless: I can add TRUE and FALSE.  I was mainly trying to hit the python and yaml casing.  But I can make this case insensitive.

server/common/config/external_config.py
bmccandless: The eval is converting to ints or double or dictionaries, etc.
E.g. "1" will convert to int(1).
If the value doesn't eval, like "localhost:5005", then it's caught and we assume the wanted a string.

server/test/unit/common/config/test_external_config.py
bmccandless: The function doesn't fail.  It will return the input value if it doesn't convert to anything else.

server/common/config/external_config.py
bmccandless: Actually, I suppose someone could construct a string that failed in a different way than NameError or SyntaxError.  I'll just catch any exception here.

server/common/config/external_config.py
bmccandless: I think the pythonic way is `if not self.aws_secrets_manager__secrets` .
I kind of disagree with that since I like the clarity of being explicit about the list.
But I can change it.

server/test/unit/common/config/test_external_config.py
bmccandless: I just looked into it.  It's a bit more overhead than what I would need it for here.   I'd have to create the secrets using moto first, then access the secrets.   Will keep it in mind for the future.

server/common/config/app_config.py
```diff
@@ -73,6 +78,47 @@ def update_default_dataset_config(self, **kw):
             value.update(**kw)
         self.is_complete = False
 
+    def update_single_config_from_path_and_value(self, path, value):
+        """Update a single config parameter with the value.
+        path is a list of string, that gives a path to the config parameter to be updated.
```
MDunitz, line 83:
```suggestion
        Path is a list of string, that gives a path to the config parameter to be updated.
```

MDunitz: nit

server/common/config/app_config.py
```diff
@@ -73,6 +78,47 @@ def update_default_dataset_config(self, **kw):
...
+        elif path[0] == "dataset":
+            attr = "__".join(path[1:])
+            try:
+                self.update_default_dataset_config(**{attr: value})
+            except ConfigurationError:
+                raise ConfigurationError(f"unknown config parameter at path: '{str(path)}'")
+
+        elif path[0] == "per_dataset_config":
+            if len(path) < 2:
+                raise ConfigurationError(f"path is invalid: got '{path}'")
```
MDunitz, line 111: Would it be possible to make this more specific? This issue is that they are trying to add a per_dataset_config without giving a dataroot/settings right?

server/common/config/app_config.py
```diff
@@ -73,6 +78,47 @@ def update_default_dataset_config(self, **kw):
...
+            dataroot = path[1]
+            if dataroot not in self.dataroot_config:
+                raise ConfigurationError(f"unknown dataroot in path: got '{path}'")
```
MDunitz, line 114: Could this be more specific as well? 
```suggestion
                raise ConfigurationError(f"unknown dataroot in path: got '{path}', dataroots specified in config are {self.dataroot_config}")
```

server/common/config/app_config.py
```diff
@@ -94,12 +140,16 @@ def update_from_config_file(self, config_file):
...
             self.dataroot_config[key].update_from_config(dataroot_config, f"per_dataset_config__{key}")
 
+        if config.get("external"):
+            self.external_config.update_from_config(config["external"], "external")
+
         self.is_complete = False
 
-    def write_config(self, config_file):
+    def config_to_dict(self):
         """output the config to a yaml file"""
```
MDunitz, line 149: can you update this for the new function?

server/common/config/app_config.py
```diff
@@ -111,15 +161,22 @@ def write_config(self, config_file):
...
             for attrname in dataset.keys():
                 config[f"per_dataset_config__{dataroot_tag}__" + attrname] = getattr(dataroot_config, attrname)
+        for attrname in external.keys():
+            config["external__" + attrname] = getattr(self.external_config, attrname)
 
         config = unflatten(config, splitter=lambda key: key.split("__"))
+        return config
+
+    def write_config(self, config_file):
+        config = self.config_to_dict()
```
MDunitz, line 171:
```suggestion
 """output the config to a yaml file"""
        config = self.config_to_dict()
```

MDunitz (approved):
lgtm - added a few nits around around error messages but otherwise looks good!

PR #1905, opened 2020-10-05
Allow columns encoded in float64 to be rendered as part of continuous value histograms.

maniarathi:
This is interestingly a regression. We previously encoded everything in float32 but because of complications around encoding, needed to allow float64 values to pass through. We didn't update this filter to ensure that some columns that were previously encoded in float32 but are now in float64 to pass through.


seve (approved):
LGTM! Could you provide a QA dataset so we can see a non-working example?

maniarathi:
Yep! `cellxgene launch s3://cellxgene-data-wrangling-prod/strand_lab_human_prostate_atlas-60/strand_lab_human_prostate_atlas-60-curated.h5ad --open`

PR #1907, opened 2020-10-06
create e2e test for auth buttons

seve:
This PR adds a few helpful additions regarding authentication.

Changes:
* e2e tests are now run on test_oauth via a passed config.yaml 
* node dev server correctly handles `/login` and `/logout` endpoints to make developing for auth easier
* Introduced auth e2e tests to check that buttons display and work

### DIFF
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -17,6 +17,7 @@ import {
   goToPage,
   typeInto,
   waitByID,
+  clickOnUntil,
 } from "./puppeteerUtils";
 
 import {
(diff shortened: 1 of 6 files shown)


bmccandless (approved):
LGTM

client/__tests__/e2e/e2e.test.js
```diff
@@ -521,6 +522,17 @@ test("lasso moves after pan", async () => {
   expect(panCount).toBe(initialCount);
 });
 
+describe("auth buttons", () => {
+  test("login then logout", async () => {
```
lesliecodes, line 526: This is more a general question, I was looking at the authButton component and don't see any unit tests... and then I saw that we don't have tests for any components. Am I looking in the wrong place? We should add a case for displaying the `logged in as` section (but that doesn't need to be a e2e layer test)

tihuan: Yeah we don't have React component related unit tests yet. But we should start adding them! 😄

tihuan:
@seve do you have time today to resolve merge conflicts and merge this PR? Thanks a lot!

tihuan:
Whoohoo! GO Seve! Thanks so much!

PR #1908, opened 2020-10-06
Performance test annotations

MDunitz:
## Need
- to determine the limits of performance for annotations feature

## Approach
- create a test for the creation of annotations for datasets of varying size (~10k, ~100k and the largest dataset currently in prod)
- use locust to test the get endpoints for annotations schema and obs 

See [1738](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1738) for details on testing results


bmccandless (commented):
Suggest moving these files to:  server/test/performance.
There are other similar sorts of things there.

scripts/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,200 @@
...
+                    self.test_notes[dataset_name][f"num_categories_{category_count}"][f"max_label_{unique_labels}"][
+                        "put_request"
+                    ] = {"response_status": "failed", "request_time": str(response_time)}
+                else:
+                    self.test_notes[dataset_name][f"num_categories_{category_count}"][f"max_label_{unique_labels}"][
+                        "put_request"
+                    ] = {"response_status": response.status_code, "request_time": str(response_time)}
+
+
+def test_all_datasets():
```
bmccandless, line 187: What calls test_all_datasets?
I think there is something missing to run the tests.

MDunitz: For now I think the tests will only be run manually. You still need to generate a cookie and update that in the code. I tried using the machine to machine tokens from auth0 but they run into issue bc they dont have a user profile. 

I can add a command to the Makefile, I just wanted to be sure people knew they needed to create a cookie

scripts/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,200 @@
...
+        request_info = {}
+        for dataset in self.test_datasets.keys():
+            request_info[dataset] = {}
+            for cat_count in self.annotations_category_count:
+                request_info[dataset][f"num_categories_{cat_count}"] = {}
+                for unique_labels in self.max_labels:
+                    request_info[dataset][f"num_categories_{cat_count}"][f"max_label_{unique_labels}"] = {}
+        return request_info
+
+    def create_annotations_dict_threaded(self, dataset_name, category_count, label_max):
```
bmccandless, line 73: You are using a process pool, so maybe "create_annotations_dict_multi_process"

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,205 @@
...
+
+test_datasets = {
+    "smallest": {
+        "dataset_url": "kampmann_lab_human_AD_snRNAseq_EC_inhibitoryNeurons-53-remixed.cxg",
+        "name": "smallest",
+        "num_cells": 5270,
```
bmccandless, line 24: Instead of hard-coding the cell count, you can retrieve that from the the /schema endpoint.

It's in the schema["dataframe"]["nObs"].

That might make the script easier to use in a more general setting, where the script might accepts a list of dataset locations.

bmccandless:
I think we should consider how to keep the generic open source cellxgene separate from the CZI single-cell hosted cellxgene.   This PR adds a few things about our servers and datasets that wouldn't apply to any one except for us.   There are also a few things that got added to the server/eb directory too.  I think we ought to keep a clean separation, and put the specific coding in a different repo (maybe single-cell-infra, or a new repo).

Other than that this PR looks good to me.

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,205 @@
...
...kyVnNiSGhuWlc1bExYTnRiMnRsTFhSbGMzUXJaR1YyUUdOb1lXNTZkV05yWlhKaVpYSm5MbU52YlNJc0ltVnRZV2xzWDNabGNtbG1hV1ZrSWpwbVlXeHpaU3dpYVhOeklqb2lhSFIwY0hNNkx5OXNiMmRwYmk1alpXeHNlR2RsYm1VdVpHVjJMbk5wYm1kc1pTMWpaV3hzTG1ONmFTNTBaV05vYm05c2IyZDVMeUlzSW5OMVlpSTZJbUYxZEdnd2ZEVm1Oekk0TVRnd05EZGlNVGhpTURBM05tRTBZakV6WmlJc0ltRjFaQ0k2SW1NelR6VjJkelpzVkdWMVREVjFhRkozWWpWR2RVc3hTbmR4TWpKdFVqVlNJaXdpYVdGMElqb3hOakF4TkRnNU1UY3pMQ0psZUhBaU9qRTJNREUxTWpVeE56TXNJbTV2Ym1ObElqb2lTSE51WkVvNFVGZFpTRlJ5YTBkTVVWcDBhM0lpZlEuaEZCcDM3VnJUVi1vX3RMR0p1SW1ieURyRXVZU0tBSlBjT0dxdW9YWm5QZ1Q5UW9PX2RGZGdBZzVnQ2RtVTlUa2dQTkxSXzJsczlPX0lxX2ZBX2NMbTNpVW5rY0ZnOFBsUVByMmh6RzVlN1AzLUVmX3RxZVdOTGw5LW9RTWJwekUyUTlaMXJyaTgyQklJMndzcEFnVWNJVHRYcDdLOTloOTVoYmM5a3U3REI0ZEtYUHNmQ0M0ZDFBc1cwcGNtcXJyMElEeHhPOVZ2LVB6cVNNdm5PQ2t2WmRza3ctMEpHOElsTm9ZOXBBWkxNWG9wMTIwNkZKdU95WHBCVWxiTUxJOVE5aVNhbEQ1dzduU0VQUzhjUl9YTHZXTzhiQ1ZxaXRURTZsajhnRlR1Rlc3emxfY005WVpqbVRwbUcweE95Q3R4VXdfQ0NCMi1NUmx2dU9GN3NmOHRnIiwgInJlZnJlc2hfdG9rZW4iOiAiTTlKakdLSnV2eklPMjRIbW9wQkhreEJwem5jcmxTSU4zaVZCMWxZaWxQRFh4IiwgImV4cGlyZXNfYXQiOiAxNjAxNTc1NTczfQ==; session=eyJfYXV0aDBfYXV0aGxpYl9ub25jZV8iOiJIc25kSjhQV1lIVHJrR0xRWnRrciJ9.X3TJFQ.1aZy0qSeclvinL6PESQl6rNuoU"  # noqa E501
```
bmccandless, line 18: It feels wrong to put a credential into the code, even though it is expired and for a test user.
Also the format depends (in some ways) on how the server was configured.
I suggest removing the value here, and replacing it with "<replace with the actual cookie>".

I think we can come up with a way to generate a temporary token for test scripts like this.

MDunitz: This is the code recommended by auth0 for a machine to machine token, but it errors out when we try to grab the user info on the request

`     def get_token(self):
         import http.client
         dev_payload = "{\"client_id\":\"*****\",\"client_secret\":\"*****\",\"audience\":\"cellxgene.dev.single-cell.czi.technology/\",\"grant_type\":\"client_credentials\"}"
         conn = http.client.HTTPSConnection("czi-cellxgene-dev.us.auth0.com")
         headers = {'content-type': "application/json"}
         conn.request("POST", "/oauth/token", dev_payload, headers)
         res = conn.getresponse()
         data = json.loads(res.read().decode("utf-8"))
         return f"{data['token_type']} {data['access_token']}"`

MDunitz:
> I think we should consider how to keep the generic open source cellxgene separate from the CZI single-cell hosted cellxgene. This PR adds a few things about our servers and datasets that wouldn't apply to any one except for us. There are also a few things that got added to the server/eb directory too. I think we ought to keep a clean separation, and put the specific coding in a different repo (maybe single-cell-infra, or a new repo).
> 
> Other than that this PR looks good to me.

@bmccandless I'll create a ticket for this

server/test/performance/performance_test_annotations_backend.py
MDunitz: added a function for grabbing the cell count from the schema

bmccandless (approved):
LGTM

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+                    request_info[dataset][f"num_categories_{cat_count}"][f"max_label_{unique_labels}"] = {}
+        return request_info
+
+    def create_annotations_dict_multi_process(self, dataset_name, category_count, label_max):
+        annotation_dict = {}
+        futures = []
+        categories = [f"Category{i}" for i in range(category_count)]
+        if not self.test_datasets[dataset_name]["num_cells"]:
+            self.set_cell_count(dataset_name)
+        with concurrent.futures.ProcessPoolExecutor(max_workers=5) as executor:
```
Bento007, line 85: is this run in a container or github actions. I'm not sure if you will have access to multiple processors in those environments. If you are going to parallelize this, why not use all of the cores in the system?

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+        rng.shuffle(cells)
+        return category_name, cells
+
+    @staticmethod
+    def convert_to_fbs(annotation_dict):
+        df = pd.DataFrame(annotation_dict)
+        return encode_matrix_fbs(matrix=df, row_idx=None, col_idx=df.columns)
+
+    @staticmethod
+    def generate_labels(unique_label_count):
```
Bento007, line 123: Does this function produce a consistent number of labels? Line 126, makes me think the number of labels produces is variables. This could affect the reproducibility of the performance test.

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+        for i in range(unique_label_count):
+            length = random.randrange(10, 20)
+            labels.append(f"{i}__" + "".join(random.choice(string.ascii_letters) for z in range(length)))
+        return labels
+
+    @contextmanager
+    def elapsed_timer(self):
+        start = default_timer()
+        elapser = lambda: default_timer() - start  # noqa E731
+        yield lambda: elapser()
```
Bento007, line 134: Can you explain what is happening on line 133-134, what's with the two lambdas?

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+        end = default_timer()
+        elapser = lambda: end - start  # noqa E731
```
Bento007, line 136: is this used?

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+                    else:
+                        self.test_notes[dataset_name][f"num_categories_{category_count}"][f"max_label_{unique_labels}"][
+                            "put_request"
+                        ] = {"response_status": response.status_code, "request_time": str(response_time)}
+
+
+def test_all_datasets():
+    """
+    Run time is dependent on number of datasets, dataset size, number of categories/number being tested and number of
+     unique label counts being tested. However it generally takes a long time. I recommend running this in tmux
```
Bento007, line 197: Can you put a number to it. Is a long time 1 hour, 30min, 10min, ect.... It will help to know what a long time roughly means.

Bento007 (commented):
I have questions

server/test/performance/performance_test_annotations_backend.py
```diff
@@ -0,0 +1,215 @@
...
+        "num_cells": 9409,
+    },
+    "80k": {
+        "dataset_url": "Single_cell_gene_expression_profiling_of_SARS_CoV_2_infected_human_cell_lines_H1299-27-remixed.cxg",  # noqa E501
+        "name": "80k",
+        "num_cells": 81736,
+    },
+    "140k": {"dataset_url": "Single_cell_drug_screening_a549-42-remixed.cxg", "name": "140k", "num_cells": 143015},
+    "largest": {"dataset_url": "human_cell_landscape.cxg", "name": "largest", "num_cells": 599926},
+    "1million": {"dataset_url": None, "name": "1million", "num_cells": 1000000},
```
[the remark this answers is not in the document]
MDunitz: For cell counts without datasets the intention was just to determine how long it would take to generate the matrix/how large the matrix would be

server/test/performance/performance_test_annotations_backend.py
MDunitz: https://github.com/chanzuckerberg/cellxgene/pull/1908#discussion_r500574719

server/test/performance/performance_test_annotations_backend.py
MDunitz: it generates a consistent number of labels, but the labels are of varying length

server/test/performance/performance_test_annotations_backend.py
MDunitz: this is just a helper function to time how long different function within the test take

server/test/performance/performance_test_annotations_backend.py
MDunitz: yes

server/test/performance/performance_test_annotations_backend.py
MDunitz: its completely dependent on the number of datasets the size of the datasets etc it can take hours. This is just a suggestion that people use tmux if they are testing lots of datasets/larger datasets/a lot of combos of max labels and num of categories

MDunitz (commented):
> I have questions

Happy to follow up with your questions on slack/over vc

server/test/performance/performance_test_annotations_backend.py
Bento007: you can get CPU count using:
```
import multiprocessing

multiprocessing.cpu_count()
```
and set that to you max number of workers.

server/test/performance/performance_test_annotations_backend.py
Bento007: got it. I had to parse the () closer.

server/test/performance/performance_test_annotations_backend.py
Bento007: Where does it get used? From what I can see the function returns None after the yield.

server/test/performance/performance_test_annotations_backend.py
Bento007: I suggest changing to ```yield elapser``` adding the lambda is extra in this case.

server/test/performance/performance_test_annotations_backend.py
Bento007: I don't have a better suggestion of phrasing, but "a long time" is not particularly helpful. Maybe remove it all together.

PR #1911, opened 2020-10-07
thuang-1840-authn-prompt

tihuan:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1840undefined

1. Adds `/* eslint-disable camelcase -- Allowing snake_case from BE */` in `client/src/components/menubar/authButtons.js` <--- NO LONGER NEEDED!
2. Extracts `LocalStorage` helper functions into its own file `client/src/components/util/localStorage.js`, and reuse it in both cookie prompt and login prompt
3. Adds `Popover` for login prompt
4. Adds checkbox option to stop showing the login prompt next time the user visits the app

NOTE:
1. I did a slight variation of what the ticket describes, because now we're showing the `<Popover />` over the Login button, so I thought adding another `<Login />` button inside the `<Popover />` would look weird. But please let me know what you think. Thank you!

QA steps:
1. Get `<AuthButtons />` to show by changing the following in `client/src/components/leftSidebar/topLeftLogoAndTitle.js`:

BEFORE
```tsx
{!userinfo.is_authenticated ? (
    <AuthButtons auth={auth} userinfo={userinfo} />
  ) : null}
```

AFTER
```tsx
{true? (
    <AuthButtons auth={auth} userinfo={userinfo} />
  ) : null}
```

2. Play with the prompt!

Expected behaviors:
1. Check the checkbox and click "OK" button, refresh the page, and you should no longer see the prompt if you're not logged in

2. Any other conditions should still show the prompt after refreshing the page

3. Prompt should only show if authN is on, not logged in, and prompt LocalStorage `cxg.LOGIN_PROMPT` value is not `"off"`

**UPDATE: NEW COPY**

Screen Shot 2020-10-08 at 4 40 47 PM

gif

PTAL! Thank you!

### DIFF
--- a/client/src/components/menubar/authButtons.js
+++ b/client/src/components/menubar/authButtons.js
@@ -1,4 +1,5 @@
-import React from "react";
+import React, { useState } from "react";
+
 import {
   AnchorButton,
   Button,
@@ -6,19 +7,35 @@ import {
   Tooltip,
   Popover,
   Menu,
+  Elevation,
+  PopoverPosition,
+  Checkbox,
+  Card,
 } from "@blueprintjs/core";
+
 import { IconNames } from "@blueprintjs/icons";
 
 import * as globals from "../../globals";
+
 import styles from "./menubar.css";
+
+import { storageGet, storageSet, KEYS } from "../util/localStorage";
 
 const BASE_EMOJI = [0x1f9d1, 0x1f468, 0x1f469];
 const SKIN_TONES = [0x1f3fb, 0x1f3fc, 0x1f3fd, 0x1f3fe, 0x1f3ff];
 const MICROSCOPE = 0x1f52c;
 const ZERO_WIDTH_JOINER = 0x0200d;
 
+const LOGIN_PROMPT_OFF = "off";
+
 const Auth = React.memo((props) => {
+  const [isPromptOpen, setIsPromptOpen] = useState(shouldShowPrompt());
+
   const { auth, userinfo } = props;
+
+  const isAuthenticated = userinfo && userinfo.is_authenticated;
+
+  window.userinfo = userinfo;
 
   const randomInt = Math.random() * 15;
   const sexIndex = Math.floor(randomInt / 5);
@@ -31,9 +48,9 @@ const Auth = React.memo((props) => {
     MICROSCOPE
   );
 
-  if (!auth?.["requires_client_login"]) return null;
+  if (!shouldShowAuth()) return null;
 
-  if (userinfo?.["is_authenticated"]) {
+  if (isAuthenticated) {
     const PopoverContent = (
       <Menu>
         <MenuItem
@@ -67,7 +84,7 @@ const Auth = React.memo((props) => {
     );
   }
 
-  return (
+  const LoginButton = (
     <Tooltip
       content="Log in to cellxgene"
       position="bottom"
@@ -76,7 +93,6 @@ const Auth = React.memo((props) => {
       <AnchorButton
         type="button"
         data-testid="log-in"
-        disabled={false}
         href={auth.login}
         className={styles.menubarButton}
       >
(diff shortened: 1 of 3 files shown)


client/src/components/menubar/authButtons.js
```diff
@@ -1,31 +1,122 @@
...
-        content="Log in to cellxgene"
-        position="bottom"
-        hoverOpenDelay={globals.tooltipHoverOpenDelay}
-      >
-        <AnchorButton
-          type="button"
-          data-testid="auth-button"
-          disabled={false}
-          href={!userinfo.is_authenticated ? auth.login : auth.logout}
+      <div style={{ position: "relative" }}>
```
tihuan, line 44: Add a relative parent to wrap siblings `<AuthButton />` and `<div style={{ position: "absolute" }} />`, so the "absolute" sibling can overlap with `<AuthButton />` and position `<Popover />` correctly

We can't just wrap Popover > Tooltip > AuthButton, because that disables `<AuthButton />, which means clicking on the `<AuthButton />` will only trigger the `<Popover />` instead of navigating to the login/logout link

bmccandless (approved):
Looks good to me!  Thanks @tihuan .

tihuan:
Thanks so much for the quick review, Brian!!

PTAL for styling and cellxgene code best practices @pdugan20 @colinmegill @seve 😄 🙏 

Thank you!

colinmegill:
I think we need to clarify the copy.

Does this mean: 

1. Log in now to save your annotations, but regardless of whether you do they'll be here when you come back
2. Log in now to save your annotations, so that they are here when you come back

image

client/src/components/menubar/authButtons.js
seve: > which means clicking on the ``will only trigger the`` instead of navigating to the login/logout link

 typo :)

seve (approved):
Local Storage, exciting!  One small nit about ESLint. LGTM otherwise!

colinmegill (approved):

client/src/components/menubar/authButtons.js
tihuan: Ah thank you! Missing the closing ` 😆 Corrected!

client/src/components/menubar/authButtons.js
```diff
@@ -1,31 +1,122 @@
...
-  if (!auth || (auth && !auth.requires_client_login)) return null;
+  if (shouldShowAuth()) return null;
+
+  const AuthButton = () => {
+    return (
+      <AnchorButton
+        type="button"
+        data-testid="auth-button"
+        disabled={false}
+        href={!userinfo?.is_authenticated ? auth.login : auth.logout}
```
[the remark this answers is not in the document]
tihuan: Ahh thank you! TIL! Will update!

client/src/components/leftSidebar/topLeftLogoAndTitle.js
```diff
@@ -107,9 +107,9 @@ class LeftSideBar extends React.Component {
               userinfo,
             }}
           />
-          {!userinfo.is_authenticated ? (
+          {!userinfo.is_authenticated && (
```
lesliecodes, line 110: 🎉  I think we can even get rid of the ()

```suggestion
          {!userinfo.is_authenticated && <AuthButtons auth={auth} userinfo={userinfo} />}
```

client/src/components/menubar/authButtons.js
```diff
@@ -1,31 +1,122 @@
...
+  const isAuthenticated = userinfo && userinfo.is_authenticated;
+
+  if (shouldShowAuth()) return null;
+
+  const AuthButton = () => {
+    return (
+      <AnchorButton
+        type="button"
+        data-testid="auth-button"
+        disabled={false}
```
lesliecodes, line 35: nit: I think we could just omit this if it's false

bmccandless:
> I think we need to clarify the copy.
> 
> Does this mean:
> 
> 1. Log in now to save your annotations, but regardless of whether you do they'll be here when you come back
> 2. Log in now to save your annotations, so that they are here when you come back
> 
> image

How about:  "Log in to create your own annotations.  Please note, logging in later will refresh this page and you may lose progress."

lesliecodes (changes requested):
Looks great! Can we add a unit test for this? 🙏

client/src/components/menubar/authButtons.js
tihuan: Good point! Done 😆

client/src/components/leftSidebar/topLeftLogoAndTitle.js
tihuan: Ah good point! In this case, since `prettier` wraps code when the line is too long, it automatically does the following instead!

```ts
{!userinfo.is_authenticated && (
  <AuthButtons auth={auth} userinfo={userinfo} />
)}
```

tihuan:
Thanks @lesliecodes ! cellxgene doesn't have unit test, but we can add an integration test after Seve's https://github.com/chanzuckerberg/cellxgene/pull/1907 is merged, since it has the infrastructure code for running AuthN features locally!

client/src/components/leftSidebar/topLeftLogoAndTitle.js
lesliecodes: ooh nvm then!

lesliecodes (approved):
Thanks Timmy!

tihuan:
Thanks Leslie!

I think we're close to being able to merge. Last bit on the buttons, pending decision in Slack thread 😆

tihuan:
Latest iteration!

Screen Shot 2020-10-07 at 7 09 09 PM


gif

Issue #1900, opened 2020-10-02
Layout Issues with Auth Buttons


seve:
Introduction of the auth buttons to the top left corner has led to overflow onto multiple lines and skewing of the LSB.

@signechambers1 LMK if we want to move around any of the buttons or labels to increase visibility

lesliecodes:
@colinmegill please update the description after the decisions are made :)

colinmegill:
User profile photo if avail, gender neutral scientist emoji if not

seve:
Are we moving the log in + info menu buttons back to the menubar?

bmccandless:
I'll add "picture" to the /userinfo endpoint.  This will be a URL.  If it doesn't exist / can't be retrieved, I'll set picture to null.  Sound good?

lesliecodes:
That sounds good!

Mock from Colin for scientist emoji:
![image.png](https://images.zenhubusercontent.com/5f5fb4df4fd147847ef7b38c/7ea73d28-6046-4e72-bda0-e45015d9bf1d)

colinmegill:
That's perfect @bmccandless.

It's important that we keep the title in the upper left sidebar. It's a feature in that it is quick reference and appears in every screenshot. Because it's truncated, it doesn't matter that some titles may be 100's of characters long. It also makes sense for the info menu to be on the left side bar:

`This is cellxgene, you're on this dataset, here's more info about the tool and the dataset you're working with`

The usual pattern for login is that it's more towards the left. Thanks to Harley for bringing up that we could claim the space above genesets for login info — we may want to do that. The quickest change for this PR is to put login back in the menu bar in the graph, and we can discuss whether that is its permanent home later. Graph then becomes:

`Here are actions you can take, one of which is logging in`

lesliecodes:
@signechambers1 this is ready for QA!

signechambers1:
@lesliecodes or @seve the profile picture isn't working for me ![Screen Shot 2020-10-07 at 6.37.51 PM.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/afb6f9d7-5af1-4138-86e8-c2f015227af2)

lesliecodes:
on staging: 
![image.png](https://images.zenhubusercontent.com/5f5fb4df4fd147847ef7b38c/5231662d-a096-44bd-ac6b-69285f5fb9e1)

lesliecodes:
[...]

PR #1914, opened 2020-10-07
Add "picture" to the /userinfo endpoint.

bmccandless:
This may be null or a URL

 #1900

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ lint: lint-server lint-client
 
 .PHONY: lint-server
 lint-server: fmt-py
-	flake8 server --per-file-ignores='server/test/fixtures/dataset_config_outline.py:F821 server/test/fixtures/server_config_outline.py:F821'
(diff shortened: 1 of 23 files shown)


Bento007 (approved):
LGTM

seve (changes requested):
Could you add a picture to the test_oauth type for testing purposes?

bmccandless:
> Could you add a picture to the test_oauth type for testing purposes?

Okay, do you also need a way to test the "null" case?

seve:
> Okay, do you also need a way to test the "null" case?

If it's easy on your end sure! Otherwise, I could mock it in the e2e testing.

bmccandless:
> > Okay, do you also need a way to test the "null" case?
> 
> If it's easy on your end sure! Otherwise, I could mock it in the e2e testing.

How about a way for you to specify the picture when you login.
   localhost:5005/api/v0.2/login?picture=mypicture.png

This will login the test user will a picture, else picture will be null.
(I already have this setup for username, userid, email, so it's easy)

seve (approved):
Thanks Brian

PR #1919, opened 2020-10-08
Add a function to check the configuration for errors.

bmccandless:
This can be used as a sanity check before a deployment:

  chanzuckerberg/single-cell#63


bmccandless:
A lot of the changes in this PR are due to flask_secret_key, which wasn't properly being checked.   It is now required to be set, so a lot of test cases had to be updated.
The actual new code here is the server/eb/check_config.py script that verifies that the config file passes all the checks.

server/eb/check_config.py
```diff
@@ -0,0 +1,39 @@
+import sys
+import argparse
+import yaml
+
+from server.common.config.app_config import AppConfig
+
+
+def main():
+    parser = argparse.ArgumentParser("A script to check hosted configuration files")
```
maniarathi, line 9: Just curious -- is it possible to generalize this so that any config could be checked? We could move the logic into a separate class and then pull the checker in multiple locations -- for example here and in the CLI potentially for other folks who host.

maniarathi (approved):
Thank you for updating the single character variable names!

PR #1924, opened 2020-10-09
add and check system arg to state auth type

seve:
Turns out that our e2e tests are called in more places than just our GitHub actions. 

PR checks if `CXG_AUTH_TYPE ` is test (case-insensitive) to conditionally smoke test the auth buttons.

### DIFF
--- a/client/__tests__/e2e/e2e.test.js
+++ b/client/__tests__/e2e/e2e.test.js
@@ -523,7 +523,9 @@ test("lasso moves after pan", async () => {
 });
 
 const describeIfCalledByMakeFileTarget =
-  process.env.CXG_AUTH_TYPE === "test" ? describe : describe.skip;
+  process.env.CXG_AUTH_TYPE?.toLowerCase() === "test"
+    ? describe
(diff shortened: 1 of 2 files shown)


client/__tests__/e2e/e2e.test.js
```diff
@@ -522,7 +522,12 @@ test("lasso moves after pan", async () => {
   expect(panCount).toBe(initialCount);
 });
 
-describe("auth buttons", () => {
+const describeIfCalledByMakeFileTarget =
```
maniarathi, line 525: Question - are you able to get the configuration values on the frontend? Then you could check whether the oauth `type==none` which would effectively be the same thing without needing to create the additional environment variable?

seve: I can make an API fetch from the e2e tests for the config endpoint.  How does that sound?

maniarathi: SGTM 👍

maniarathi (approved):

PR #1925, opened 2020-10-09
Revert "Allow columns encoded in float64 to be rendered as part of continuous value histograms."

maniarathi:
Reverts chanzuckerberg/cellxgene#1905


maniarathi:
This change breaks RSB functionality where some histograms are not brushable so reverting.

colinmegill (approved):

seve (approved):

PR #1927, opened 2020-10-12
Remove Continuous vars with 1 value from histogram, add to info drawer

MDunitz:
## Need
- If an observation is an int type but it only has one value it shouldnt appear in a histogram

## Approach
- check if min/max match, if they match display in the infodrawer instead

Note that timepoint now appears in the info drawer rather than as a histogram

Screen Shot 2020-10-12 at 18 41 43
Screen Shot 2020-10-12 at 18 41 25

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -650,7 +650,8 @@ class HistogramBrush extends React.PureComponent {
       Number.isNaN(summary.min) ||
       Number.isNaN(summary.max);
 
-    const OK2Render = !summary.categorical && !nonFiniteExtent;
+    const OK2Render =
+      !summary.categorical && !nonFiniteExtent && !isSingleValue;
 
     return {
       histogram,
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -450,6 +450,7 @@ const Histogram = ({
     isScatterplotYYaccessor: state.controls.scatterplotYYaccessor === field,
     continuousSelectionRange: state.continuousSelection[myName],
     isColorAccessor: state.colors.colorAccessor === field,
+    singleContinuousValues: state.singleContinuousValue.singleContinuousValues,
   };
 })
 class HistogramBrush extends React.PureComponent {
@@ -608,7 +609,7 @@ class HistogramBrush extends React.PureComponent {
   };
 
   fetchAsyncProps = async () => {
-    const { annoMatrix, field, dispatch } = this.props;
+    const { annoMatrix, field, dispatch, singleContinuousValues } = this.props;
     const { isClipped } = annoMatrix;
 
     const query = this.createQuery();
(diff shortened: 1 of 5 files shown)


client/src/components/infoDrawer/infoDrawer.js
```diff
@@ -49,14 +80,46 @@ class InfoDrawer extends PureComponent {
...
-            datasetTitle,
-            aboutURL,
-            singleValueCategories,
-            dataPortalProps,
-          }}
-        />
+        <Async
+          watchFn={InfoDrawer.watchAsync}
+          promiseFn={this.fetchAsyncProps}
+          watchProps={{ schema }}
```
seve, line 86: Will the schema ever change?  There's a minor edge case where a user clips to a single continuous value, how will the drawer react?

client/src/components/infoDrawer/infoDrawer.js
```diff
@@ -49,14 +80,46 @@ class InfoDrawer extends PureComponent {
...
+        >
+          <Async.Pending>
+            <InfoFormat
+              {...{
+                datasetTitle,
+                aboutURL,
+                singleValueCategories,
+                dataPortalProps,
+              }}
+            />
```
seve, line 96: We need to reintroduce the loading state for the InfoFormat component (the skeleton + placeholder generation)

client/src/components/brushableHistogram/index.js
```diff
@@ -650,7 +650,8 @@ class HistogramBrush extends React.PureComponent {
       Number.isNaN(summary.min) ||
       Number.isNaN(summary.max);
 
-    const OK2Render = !summary.categorical && !nonFiniteExtent;
+    const OK2Render =
+      !summary.categorical && !nonFiniteExtent && !isSingleValue;
```
seve, line 654: Can we make the `isSingleValue` check as early as possible and instead of assigning the value to `OK2Render` just do an early exit with the return value of `{OK2Render: false}` to avoid unused computes

seve (changes requested):
Few questions about performance and edge cases

Annotation smoke tests are timing out for some reason, sometimes this means that the page is crashed or that something isn't where its expected and a querySelector fails to find its target.

seve:
I just caused a little mischief, if you clip far down enough using the subset button and brushing the RSB histograms you can get all the continuous data to disappear and show in the drawer

MDunitz:
Summary of slack discussion 
- We dont want to update the info drawer past the first render to avoid issues with continuous vars selected down to one value appearing in it
- We do need to make sure that the info drawer is populated by non selected (original) annoMatrix
- We dont want to rerun summarize or reload data
Solution
-  create a reducer, make the checks in the RSB, add the removed histograms to the redux state, and then in the info drawer just pull the values from the redux state 
- this puts all the checks in one place and allows us to (re)remove the async code from the info drawer

client/src/components/brushableHistogram/index.js
```diff
@@ -620,7 +621,10 @@ class HistogramBrush extends React.PureComponent {
     const summary = column.summarize();
     const range = [summary.min, summary.max];
 
-    if (summary.min === summary.max && !isClipped) {
+    if (
+      (summary.min === summary.max && !isClipped) ||
+      singleContinuousValues.has(field)
```
seve, line 626: If the field is already in the singleContiousValues, we probably don't want to call dispatch again since the key/value are already in the map.  Might want to instead add an else if that just returns the same nulled object to avoid the unnecessary dispatch.

seve (changes requested):
Nice fix for the edge case, one small piece of feedback

seve (approved):
LGTM!

PR #1935, opened 2020-10-16
Remove errornous checking for converting float64 to float32. In reality the slight difference by downcasting is totally fine.

maniarathi:
Closes #1934


server/common/utils/type_conversion_utils.py
```diff
@@ -97,15 +97,7 @@ def can_cast_to_float32(dtype, array_values):
...
-        # Try to convert the array to float32
-        converted_float32_values = array_values.to_numpy(np.float32)
-        original_values = array_values.to_numpy()
-
-        # Verify that the two arrays are equal except for NaNs (which will equate to be unequal).
-        if not ((converted_float32_values != original_values) == np.isnan(original_values)).all():
-            return False
-
-        if dtype != np.float32:
+        if not np.can_cast(dtype, np.float32):
```
seve, line 100: I think the docstring needs to be change to reflect the "same output" characteristic of the function.

seve (changes requested):
One small doc change needed.

Could you also help give a sense of how minor the precision loss is?

seve:
Edited so it would close my PR on merge, hope you don't mind!

server/common/utils/type_conversion_utils.py
maniarathi: Ooh good catch! Fixed!

seve (approved):
LGTM!

server/common/utils/type_conversion_utils.py
```diff
@@ -130,9 +129,9 @@ def can_cast_to_int32(dtype, array_values=None):
             return True
         ii32 = np.iinfo(np.int32)
         if (
-            not ordered_array_values.empty
-            and (ordered_array_values.min() >= ii32.min and ordered_array_values.max() <= ii32.max)
-            or ordered_array_values.empty
+                not ordered_array_values.empty
+                and (ordered_array_values.min() >= ii32.min and ordered_array_values.max() <= ii32.max)
+                or ordered_array_values.empty
         ):
```
seve, line 135: @maniarathi - These are the only new code changes and the smoke-tests are timing out now.

PR #1936, opened 2020-10-16
Add in missing previous crossfilter which was causing the re-embedding feature to fail.

maniarathi:


bmccandless:
I think we need an e2e test of re-embedding.

seve (approved):
LGTM, nice catch!

maniarathi:
> I think we need an e2e test of re-embedding.

I've enabled re-embeddings on the E2E tests to ensure that we never make a mistake such that the cellxgene server won't launch with the flag enabled. We'll need additional tests once we decide to harden re-embeddings though and I've created a ticket [here](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/1937) to track that work.

maniarathi:
Apologies -- I need to actually remove turning on re-embeddings in the E2E tests altogether for now until we figure out if we can add scanpy to the requirements.txt file. There is some weirdness around the licenses (transitive GPL-3.0) that need to be sorted out with Legal first and I'd like for this to go through without being blocked by Legal so that we can fix the bug and push out a release.

Issue #1866, opened 2020-09-23
Dockerize cellxgene


maniarathi:
Need: Dockerizing cellxgene gives a few nice things. First, it provides an easy way for users who want to host their own cellxgene to fetch and deploy the cellxgene docker image. Second, it allows us to easy switch our deployment orchestration service. Currently we are using Elastic Beanstalk but this won't be true in the medium-term future as we will likely switch over to one of ECS or EKS once we get some guidance from Infra. 

Definition of Done:
- (REMOVED) Dockerized cellxgene is available on Elastic Container Service.
- (REMOVED) Deployment of cellxgene is updated so that the docker image is uploaded to ECS.
- [ ] Deployment of cellxgene is updated so that EB uses the uploaded docker image.

Open question: how will this play with the compute tier that is being built?

bmccandless:
Update the definition of done:

We will hold off on using ECS directly, and use EB with a Dockerfile instead.

bmccandless:
Progress:
The Dockerized EB is in main.
Next step is to deploy in dev, apply terraform, and re-deploy.  Verify there is no downtime, and dockers are actually used.

PR #1938, opened 2020-10-20
Minor changes to eb server to use Docker

bmccandless:
part of #1866

### DIFF
--- a/server/eb/app.py
+++ b/server/eb/app.py
@@ -185,7 +185,7 @@ else:
 
 if __name__ == "__main__":
     try:
-        application.run(debug=debug, threaded=not debug, use_debugger=False)
+        application.run(host=app_config.server_config.app__host, debug=debug, threaded=not debug, use_debugger=False)
     except Exception:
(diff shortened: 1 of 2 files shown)


server/eb/app.py
```diff
@@ -185,7 +185,7 @@ def get_csp_hashes(app, app_config):
 
 if __name__ == "__main__":
     try:
-        application.run(debug=debug, threaded=not debug, use_debugger=False)
+        application.run(host=app_config.server_config.app__host, debug=debug, threaded=not debug, use_debugger=False)
```
MDunitz, line 188: is this dependent on changes in a different pr?

bmccandless: No this is needed for docker.    I need to pass in "0.0.0.0" for the host so that it can server outside requests.

bmccandless: There is a change in single-cell-infra that passes in host as 0.0.0.0

MDunitz (approved):
lgtm

PR #1944, opened 2020-10-22
small fix to the test suite.

bmccandless:
I noticed a few tests failed when run individually, but not as a suite.

 #1942


seve (approved):
LGTM!

PR #1948, opened 2020-10-23
Fix compatibility test

MDunitz:
## Need
- to have working compatibility tests so when they fail we are given meaningful information

## Approach
- require anndata >= 0.7
Scanpy also has the requirement.
- added a version check/warning to ensure users are aware
- updated versions used in compatibility testing


server/requirements.txt
```diff
@@ -1,4 +1,4 @@
-anndata>=0.6.20
+anndata>=0.7
```
maniarathi, line 1: Not sure if this is legit or not, but does this need to be the full semvar in order for the comparison to work? i.e. 0.7.0?

MDunitz: it worked locally but I can update it just to be safe

MDunitz, line 1:
```suggestion
anndata>=0.7.0
```

.github/workflows/compatibility_tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        anndata-version: [0.6.22.post1, 0.7.1]
+        anndata-version: [0.7, 0.7.1, 0.7.2, 0.7.3, 0.7.4]
```
MDunitz, line 32:
```suggestion
        anndata-version: [0.7.0, 0.7.1, 0.7.2, 0.7.3, 0.7.4]
```

server/data_anndata/anndata_adaptor.py
```diff
@@ -177,10 +177,10 @@ def _load_data(self, data_locator):
             )
 
     def _validate_and_initialize(self):
-        if anndata_version_is_pre_070() and self.server_config.adaptor__anndata_adaptor__backed:
+        if anndata_version_is_pre_070():
             warnings.warn(
-                "Use of --backed mode with anndata versions older than 0.7 will have serious "
-                "performance issues. Please update to at least anndata 0.7 or later."
+                "Use of anndata versions older than 0.7 will have serious "
+                "issues. Please update to at least anndata 0.7 or later."
```
maniarathi, line 183: Nit: I think there's a weird line break -- more words can go on line 182 :P

maniarathi (approved):
LGTM, thanks!

server/data_anndata/anndata_adaptor.py
```diff
@@ -177,10 +177,10 @@ def _load_data(self, data_locator):
             )
 
     def _validate_and_initialize(self):
-        if anndata_version_is_pre_070() and self.server_config.adaptor__anndata_adaptor__backed:
+        if anndata_version_is_pre_070():
             warnings.warn(
-                "Use of --backed mode with anndata versions older than 0.7 will have serious "
-                "performance issues. Please update to at least anndata 0.7 or later."
+                "Use of anndata versions older than 0.7 will have serious "
```
MDunitz, line 182:
```suggestion
                "Use of anndata versions older than 0.7 will have serious issues. Please update to at "
```

server/data_anndata/anndata_adaptor.py
MDunitz, line 183:
```suggestion
                "least anndata 0.7 or later."
```

Issue #1947, opened 2020-10-23
[BUG] Server errors when passed an invalid cookie


seve:
Repro: Prepend `TEST_` to the `cxguser` cookie and refresh

Server will crash and return a 500

PR #1949, opened 2020-10-23
Fix bug in oauth.

bmccandless:
The error checking was too specific, and missed a case.
Make the error checking catch all exceptions.

  #1947

### DIFF
--- a/server/auth/auth_oauth.py
+++ b/server/auth/auth_oauth.py
@@ -226,14 +226,13 @@ class AuthTypeOAuth(AuthTypeClientBase):
             else:
                 value = request.cookies.get(self.cookie_params["key"])
                 value = base64.b64decode(value)
-                try:
-                    tokensdict = json.loads(value)
-                    g.tokens = Tokens(**tokensdict)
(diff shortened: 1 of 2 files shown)


server/auth/auth_oauth.py
```diff
@@ -226,14 +226,13 @@ def get_tokens(self):
...
-                    g.pop("tokens", None)
-                    return None
+                tokensdict = json.loads(value)
+                g.tokens = Tokens(**tokensdict)
 
-        except (TypeError, KeyError):
+        except Exception:
+            # there are many types of exceptions that can be raise in the above section.
+            # It is impractical to list all the exceptions here, since that would be brittle.
+            # If an exception occurs, then return None, meaning that no token could be retrieved.
```
seve, line 235: Should we be logging the error to expose if our users are somehow being served invalid cookies?

bmccandless: Good idea, I'll add that.

seve (approved):
LGTM!

PR #1952, opened 2020-10-27
remove conditional rendering cases from color legend

seve:
The color legend component was overly protective of showing/rendering itself when those cases were already handled with D3 drawing code. 

PR also removes pointer interactions with the legend and address a bit of cleanup.


seve:
#1707

client/src/components/continuousLegend/index.js
```diff
@@ -110,24 +110,7 @@ const continuous = (selectorId, colorscale, colorAccessor) => {
   colors: state.colors,
 }))
 class ContinuousLegend extends React.Component {
```
seve, line 112: Maybe this now calls for PureComponent?

client/src/components/continuousLegend/index.js
```diff
@@ -11,20 +11,20 @@ import {
 
 // create continuous color legend
 // http://bl.ocks.org/syntagmatic/e8ccca52559796be775553b467593a9f
-const continuous = (selectorId, colorscale, colorAccessor) => {
-  const legendheight = 200;
-  const legendwidth = 80;
+const continuous = (selectorId, colorScale, colorAccessor) => {
+  const legendHeight = 200;
```
tihuan, line 15: thanks for cleaning up the variable names!

client/src/components/continuousLegend/index.js
```diff
@@ -110,24 +110,7 @@ const continuous = (selectorId, colorscale, colorAccessor) => {
...
-  componentDidMount() {
-    this.updateState(null);
-  }
-
-  componentDidUpdate(prevProps) {
-    this.updateState(prevProps);
-  }
-
-  async updateState(prevProps) {
+  async componentDidUpdate(prevProps) {
```
tihuan, line 113: I see that we no longer fetch the colors on DidMount, and only rely on DidUpdate. Does that mean we don't need to render the legend between DidMount and the first DidUpdate call?

tihuan (commented):
Thanks for fixing the bug, Seve! One quick question otherwise LGTM!

Will probably still be better to wait for @colinmegill to review, since I'm not super familiar with cellxgene code here 😆 Thank you!

client/src/components/continuousLegend/index.js
seve: Yes, the legend is one of the first components to be rendered(it's top level in `App.js`), and cellxgene initializes with `colorBy` set to undefined.  We won't need to render via D3 on DidMount

tihuan: Got it! Thank you 😄  SGTM!

colinmegill (approved):

PR #1957, opened 2020-10-30
Add link to dev_docs in Contributing section

ihnorton:
It took me a while to track down how to run the unit tests, then realized they were nicely documented under `dev_docs` (was looking at the README directly so didn't see that folder in the file listing at first).


maniarathi (approved):
Thank you so much for this addition @ihnorton!

Issue #1959, opened 2020-10-31
[BUG] unit-tests started failing on github


bmccandless:
**Describe the bug**

Several tests started failing yesterday.

**To Reproduce**

This can be reproduced by cloning cellxgene in a new virtual environment, building and running.

**Additional context**

There a few possible issues.  semver and numpy recently upgraded.
and pip may have just changed how it resolves dependency conflict.

So far I'm able to work through the issues with minor code changes.

bmccandless:
Another issue is h5py which just had a major version change yesterday to 3.0.0.
This one got us:

`Variable-length UTF-8 strings in datasets are now read as bytes objects instead of str by default, for consistency with other kinds of strings. See Strings in HDF5 for more details.`

https://docs.h5py.org/en/latest/whatsnew/3.0.html

PR #1960, opened 2020-10-31
Updates due dependency version changes.

bmccandless:
h5py recently changes and now values once returned as str are now returned as bytes.
This would have caused a much larger change, so instead the version is restricted to <3.0.0.

This caused the bulk of the testing failues.
A few other changes were needed to make a few other tests pass.

 #1959

### DIFF
--- a/server/test/unit/auth/test_oauth.py
+++ b/server/test/unit/auth/test_oauth.py
@@ -78,6 +78,13 @@ class AuthTest(unittest.TestCase):
         cls.mock_oauth_process.start()
 
         # Verify that the mock oauth server is ready (accepting requests) before starting the tests.
+
+        # The following lines are polling until the mock server is ready.
+        # The issue is we are starting a mock oauth server, then we are starting a cellxgene server,
+        # which will start making requests to the mock oauth server.
+        # So there is a race condition because the mock oauth server needs to be ready before it gets requests.
+        # We check to see if it is ready, and if not we wait 1 second, then try again.
+        # If it gets to 5 seconds, which is shouldn't, we assume something has gone wrong and fail the test.
         server_okay = False
         for _ in range(5):
             try:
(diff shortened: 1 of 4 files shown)


server/test/unit/auth/test_oauth.py
```diff
@@ -62,21 +62,34 @@ def jwks():
...
     @classmethod
     def setUpClass(cls):
+        # The port that the mock oauth server will listen on
+        cls.mock_port = random.randint(10000, 12000)
         cls.dataset_dataroot = FIXTURES_ROOT
-        cls.mock_oauth_process = Process(target=launch_mock_oauth)
+        cls.mock_oauth_process = Process(target=launch_mock_oauth, args=(cls.mock_port,))
         cls.mock_oauth_process.start()
+        server_okay = False
+        for _ in range(5):
```
maniarathi, line 80: Can you add a comment explaining why we need to retry in order for this setup to work?

maniarathi: Oops -- sorry I guess my comment wasn't clear. I was more wondering what necessitates the `for _ in range(5):` line (i.e. why do we need to retry). Does it take up to 5 seconds for the mock oauth server to be ready?

bmccandless: All this is doing is polling until the mock server is ready.  The issue is we are starting a mock oauth server, then we are starting a cellxgene server, which will start making requests to the mock oauth server.   So there is a race condition because the mock oauth server needs to be ready before it gets requests.  So, we check to see if it is ready, and if not we wait 1 second, then try again.  If it gets to 5 seconds, which is shouldn't,  we assume something has gone wrong and fail the test.

maniarathi: Ah got it -- this makes sense to me. Mind writing the same explanation as a comment above the for loop? That way, the next reader knows why we need that retry there.

maniarathi (approved):
After adding the comment, LGTM! Thank you!

bmccandless:
> After adding the comment, LGTM! Thank you!

Is it good now, or did you want the longer comment?

maniarathi:
Sorry I meant that I'd like to see your explanation to me about the race condition between the two servers and that the mock Auth0 server shouldn't take more than 5 seconds added to the inline comment in code. Then LGTM.

maniarathi:
Thanks for double checking!

PR #1966, opened 2020-11-04
fix accidental state error

bkmartinjr:
The InfoDrawer was unnecessarily rendering on any prop change scan due to a misuse of the nullish coalescing operator.  This change causes the drawer to only re-render when its props have actually changed.


seve (approved):
Thanks for catching this @bkmartinjr!

PR #1967, opened 2020-11-04
update graph GPU buffers only when necessary

bkmartinjr:
The main graph component updates all three GPU buffers (position, colors, flags) when any one of these buffers changes. Most commonly, the app interaction model changes only one at a time (eg, selection changes only flags, color-by changes only color, etc).  For large datasets, this unnecessary data copying can cause a very significant performance hit.

This change simply adds guards to only perform the buffer update when the data has changed.


seve (approved):
LGTM! Thank you, Bruce!

PR #1968, opened 2020-11-05
fix corpora schema version check 

seve:
Inverted the schema version check so that we check that it IS 1.0.0.

image


tihuan (approved):
LGTM!

PR #1972, opened 2020-11-06
Convert HGNC ids to their symbol

mckinsel:
There are entries in some var indexes like HGNC:18790. We'd like to convert that to its symbol, NSG1.


maniarathi (approved):
LGTM

Issue #1984, opened 2020-11-19
User annotation PUTs return 400s


mckinsel:
We see 400 responses to PUTs to the user annotations endpoint sometimes. You can trigger one this way:

1. Go to https://cellxgene.dev.single-cell.czi.technology/e/Single_cell_drug_screening_mcf7-44-remixed.cxg/
2. Create a new category initialized from "rt_well"

Seems like something with how types are being handled in the TileDB code, see https://sentry.prod.si.czi.technology/sci-sc/hosted-cellxgene/issues/109814/ for example.

There's also a reasonable chance there's a scaling issue as well, but we can rule that in or out better once we've resolved the type issue.

maniarathi:
Fixed went into prod today.

PR #1987, opened 2020-11-20
Convert float annotations if possible.

bmccandless:
The client converts all arrays to floats.
If a category contains integer labels, and that category is copied, it will contains floats (e.g 1.0 instead of 1).
When that category is put back to the server, it fails in the tiledb code, which does not accept floats.
The solution is to convert a float category to integer, if possible.

  #1984


MDunitz (approved):
lgtm

MDunitz (approved):

PR #1988, opened 2020-11-24
Fixing locust scale tests for cellxgene loading apis and adding a Github Actions workflow to run the tests every Sunday.

maniarathi:


mckinsel (approved):
One nit (that I think will break it) but otherwise LGTM.

One thought though:

We will be turning on caching again for a lot of these routes. So we'll have to make sure we don't like see a bad scale result, run scale tests again, see that it got a lot better and decide it must have been a transient thing.

PR #1993, opened 2020-11-30
Pin tiledb version to < 0.7.1

maniarathi:


ihnorton:
Sorry about this, we added an optional dependency for the dataframes functionality, but I made one of the checks an error instead of a warning. Will do a bump release today to fix.

maniarathi:
Thanks @ihnorton !

server/requirements.txt
```diff
@@ -20,7 +20,7 @@ pandas>=1.0,!=1.1  # pandas 1.1 breaks tests, https://github.com/pandas-dev/pand
 PyYAML>=5.3
 scipy>=1.0
 requests>=2.22.0
-tiledb>=0.5.9,>=0.6.2
+tiledb>=0.5.9,>=0.6.2,<=0.7.1
```
bmccandless, line 23: Sounds like the problem will be fixed in the next release (coming soon).  Another option is to 
just exclude the problem version, so we don't have to change again:

tiledb>=0.6.2,!=0.7.2

bmccandless (approved):
One suggestion, then LGTM

maniarathi:
Thanks @bmccandless ! Good suggestion, I've changed it to be so.

PR #1994, opened 2020-11-30
Use sparse=True for from_pandas

ihnorton:
Fixes test failures when using TileDB-Py 0.7.2, due to change in the default array type for `from_pandas`.


maniarathi:
Thank you @ihnorton !

bmccandless (approved):
LGTM

bmccandless:
Actually, that array should have been dense.   I'll rewrite to use a dense schema and fix the test.

bmccandless:
> Actually, that array should have been dense. I'll rewrite to use a dense schema and fix the test.

Oh, I see the dataframe use string typed dimensions, so the schema has to be sparse.
So, this update looks good.

PR #1996, opened 2020-12-02
check if viewport has changed and render accordingly

seve:
Fixes a bug where graph would not update as user resized window

### DIFF
--- a/client/src/components/graph/graph.js
+++ b/client/src/components/graph/graph.js
@@ -731,11 +731,14 @@ class Graph extends React.Component {
   });
 
   updateReglAndRender(asyncProps, prevAsyncProps) {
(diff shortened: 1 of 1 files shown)


tihuan (approved):
LGTM! Do we know why this wasn't an issue when we didn't check for the width and height? So interesting!

seve:
> LGTM! Do we know why this wasn't an issue when we didn't check for the width and height? So interesting!

We added some more controlled re-render conditions in #1967, buffers don't change when the viewport changes, but the actual transformations and projections onto the graph do change when viewport changes. Which is why we need to call `renderCanvas()`

tihuan:
Ooh got it! Thanks for the context, @seve 🎉 🙏

Issue #1981, opened 2020-11-19
Annotation PUT call needs to stop retrying at certain point


tihuan:
Context:
Currently FE keeps retrying a failed annotation PUT request **indefinitely**, but we need to stop retrying at some point, otherwise it will be DDOS on our servers

What:

1. Retry 3 times with exponential backoff (pick a reasonable cool down initial value please! Like 5 seconds or something)

1. Note that after retrying three times, we'll stop retrying. But if the user makes another annotation change, it should send another request + 3 retries too. This means that we'll need a mechanism to manage retry vs. new request. E.g., while retrying, a new annotation change comes in, then we should abort the current retry and start a new request + retry cycle

tihuan:
@maniarathi @signechambers1 PTAL! Not sure when we wanna pull this in, but just wanted to let you know the ticket is here lol Thank you!

seve:
Closing this issue as a "will not fix".

* Because the FE tick only PUTs when user anno updates or the previous PUT fails this isn't a major load, unless the backend is unhealthy in which we make the attempt at every tick.  In the case that the backend **is** unhealthy, this is just an attempt at resolving backend bugs with a frontend fix.

* Adding a constant exp. backoff will lead to FE clients getting in sync via the load balancer which would only compound the issue.

* Holding off PUT's for too long increases the chance of having the unwanted state of sitting on unsaved user edits, increasing the chance that we lose that data

* Our system and user size won't cause these retries to expand into DDOS territory

* The proposed solution in #2001 introduces race conditions where out of date in-flight PUT's may be handled after more recent edits. Corrupting and overriding recent user edits.

PR #2001, opened 2020-12-11
add exponential backoff to annotation put fetches

seve:
This PR creates and uses a fetch utility to help backoff repetitive fetches by collapsing fetches into a single promise which will resolve immediately on a response with an ok status or retry with exponential delay until the max retry amounts is hit, rejecting with an error or resolving non-ok response.

Also surfaced if a save is in progress

---

Hosted QA steps:

* Navigate to dataset with experimental mode enabled
* Make successful annotation change to check that there is no change in default functionality
* Make unsuccessful annotation change and watch network activity in Chrome DevTools tab
  * This can be done by intercepting response manually, with requestly, or just disconnecting from internet

CLI QA Steps:

* Same as above, but you can change the API url in `src/actions/annotation.js:363` instead of doing interception
---
#1981

### DIFF
--- a/client/src/actions/annotation.js
+++ b/client/src/actions/annotation.js
@@ -358,6 +358,7 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
       ? `?annotation-collection-name=${encodeURIComponent(dataCollectionName)}`
       : "";
   const { response, inProgress, error } = exponentialBackoffFetch(
+    annoMatrix,
     `${globals.API.prefix}${globals.API.version}annotations/obs${queryString}`,
     {
       method: "PUT",
--- a/client/src/actions/annotation.js
+++ b/client/src/actions/annotation.js
@@ -325,15 +325,12 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
   const state = getState();
   const { annotations, autosave } = state;
   const { dataCollectionNameIsReadOnly, dataCollectionName } = annotations;
-  const {
-    lastSavedAnnoMatrix,
-    saveInProgress,
-    exponentialBackoffFetch,
-  } = autosave;
+  const { lastSavedAnnoMatrix, exponentialBackoffFetch } = autosave;
 
   const annoMatrix = state.annoMatrix.base();
 
-  if (saveInProgress || annoMatrix === lastSavedAnnoMatrix) return;
+  if (annoMatrix === lastSavedAnnoMatrix) return;
+
   if (!needToSaveObsAnnotations(annoMatrix, lastSavedAnnoMatrix)) {
     dispatch({
       type: "writable obs annotations - save complete",
@@ -357,6 +354,7 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
     !dataCollectionNameIsReadOnly && !!dataCollectionName
       ? `?annotation-collection-name=${encodeURIComponent(dataCollectionName)}`
       : "";
+
   try {
     const response = await exponentialBackoffFetch(
       annoMatrix,
@@ -368,8 +366,13 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
           "Content-Type": "application/octet-stream",
         }),
         credentials: "include",
-      }
+      },
+      // DEBUG
+      // DEBUG
+      // DEBUG
+      Math.floor(Math.random() * 100)
     );
+
     if (response.ok) {
       dispatch({
         type: "writable obs annotations - save complete",
--- a/client/src/actions/annotation.js
+++ b/client/src/actions/annotation.js
@@ -325,7 +325,11 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
   const state = getState();
   const { annotations, autosave } = state;
   const { dataCollectionNameIsReadOnly, dataCollectionName } = annotations;
-  const { lastSavedAnnoMatrix, saveInProgress } = autosave;
+  const {
+    lastSavedAnnoMatrix,
+    saveInProgress,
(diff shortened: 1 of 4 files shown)


client/src/util/exponentialBackoffFetch.js
```diff
@@ -0,0 +1,68 @@
...
+              // Some non 2xx response code
+              promiseExecuter(resolve, reject);
+            }
+          } catch (e) {
+            // Fetch API error
+            error = e;
+            if (retryCount - 1 === retryTimes) reject(error);
+            promiseExecuter(resolve, reject);
+          }
+          // delay amount (3.5, 17.5, 87.5, 437.5 seconds)
```
tihuan, line 50: if we default to retry 3 times, then is it just `0, 3.5, 17.5, and 87.5`?

client/src/util/exponentialBackoffFetch.js
```diff
@@ -0,0 +1,68 @@
...
+        }, 700 * 5 ** (retryCount - 1));
+      }
+    };
+    return new Promise(promiseExecuter);
```
tihuan, line 54: since `async () => {}` already returns a promise, maybe we don't need `new Promise()`?

either `return response` or `throw error` should resolve/reject the async function's promise!

client/src/util/exponentialBackoffFetch.js
seve: Yep! I forget why I originally gave four examples, but I left the last one in since the retry amount is tweakable

client/src/util/exponentialBackoffFetch.js
seve: We need to inspect the response before we hand it back to the user. If response is not OK, we need to fetch again.  In the fetch API a response is a fulfilled promise regardless of the response code, but not to our expBackoffFetch API (non 2xx code only fulfills the promise once we expended all attemps)

client/src/util/exponentialBackoffFetch.js
```diff
@@ -0,0 +1,68 @@
...
+  let response = null;
+  let error = null;
+  let promise = null;
+
+  // This will only be called by itself or when we've seen a new matrix
+  const backoffFetch = (fetchArgs) => {
+    const promiseExecuter = (resolve, reject) => {
+      retryCount += 1;
+      // We don't want to timeout our first request
+      if (retryCount === 1) {
```
tihuan, line 13: is there a way to remove a conditional block for `retryCount === 1` and streamline the exponential backoff?

Something like: 

setTimeout(() => {}, retryCount * 700 * 5 ** retryCount)

So we get:
`retryCount === 0: 0`
`retryCount === 1: 3500`
`retryCount === 2: 35000`
`retryCount === 3: 262500`

Just trying to see if we could streamline the code XD Thank you!

seve: `setTimeout( ___, 0)` will still timeout the function until the call stack empties, didn't want to add a perf dip in immediately successful requests

seve: https://flaviocopes.com/javascript-event-loop/

client/src/util/exponentialBackoffFetch.js
tihuan: Ahh yeah, you're right! We can't avoid `new Promise`, since we need to inspect `response.ok` and `fetch()` happens in `setTimeout()`

PTAL this PR! https://github.com/chanzuckerberg/cellxgene/pull/2003#pullrequestreview-551089371

Thank you!

client/src/util/exponentialBackoffFetch.js
tihuan: That makes sense! Thanks!

client/src/util/exponentialBackoffFetch.js
tihuan: ah I was thinking right after `backoffFetch(fetchArgs)` is synchronously executed, the `setTimeout(() => {}, 0)` callback function will be the next task in the stack, so there shouldn't be any other synchronous calls in the stack?

Screen Shot 2020-12-13 at 11 36 29 PM

seve: Just seeing this now.  I might be wrong but isn't the status of the call stack at the time which the timeout added to the stack pretty indeterminate? We have a few ticks going off in the background like manual GC, WebGL re-renders and other user interactions.  I ended up pulling the timeout cb into its own function and then just using it in multiple places, that way its still performant while keeping it readable and DRY

client/src/actions/annotation.js
```diff
@@ -365,18 +366,23 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
           "Content-Type": "application/octet-stream",
         }),
         credentials: "include",
-      }
+      },
+      // DEBUG
```
bkmartinjr, line 370: cleanup debug code pls.

client/src/util/exponentialBackoffFetch.js
```diff
@@ -0,0 +1,70 @@
...
+
+      function delayFetch() {
+        if (retryCount === 0) {
+          fetchAndHandle();
+          // delay amount (3.5, 17.5, 87.5, 437.5 seconds)
+        } else timeoutId = setTimeout(fetchAndHandle, 700 * 5 ** retryCount);
+      }
+      async function fetchAndHandle() {
+        try {
+          promise.response = await mockFetch(...fetchArgs);
```
bkmartinjr, line 38: debug cleanup needed?

seve:
https://github.com/chanzuckerberg/cellxgene/issues/1981#issuecomment-745534538

Closing as #1981 is closed

PR #2003, opened 2020-12-14
thuang-1981-feedback

tihuan:
Wanted to see if this makes sense 😄 Thank you!

### DIFF
--- a/client/src/actions/annotation.js
+++ b/client/src/actions/annotation.js
@@ -325,15 +325,12 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
   const state = getState();
   const { annotations, autosave } = state;
   const { dataCollectionNameIsReadOnly, dataCollectionName } = annotations;
-  const {
-    lastSavedAnnoMatrix,
-    saveInProgress,
-    exponentialBackoffFetch,
-  } = autosave;
+  const { lastSavedAnnoMatrix, exponentialBackoffFetch } = autosave;
 
   const annoMatrix = state.annoMatrix.base();
 
-  if (saveInProgress || annoMatrix === lastSavedAnnoMatrix) return;
+  if (annoMatrix === lastSavedAnnoMatrix) return;
+
   if (!needToSaveObsAnnotations(annoMatrix, lastSavedAnnoMatrix)) {
     dispatch({
       type: "writable obs annotations - save complete",
@@ -357,6 +354,7 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
     !dataCollectionNameIsReadOnly && !!dataCollectionName
       ? `?annotation-collection-name=${encodeURIComponent(dataCollectionName)}`
       : "";
+
   try {
     const response = await exponentialBackoffFetch(
       annoMatrix,
@@ -370,6 +368,7 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
         credentials: "include",
       }
     );
+
     if (response.ok) {
       dispatch({
         type: "writable obs annotations - save complete",
--- a/client/src/actions/annotation.js
+++ b/client/src/actions/annotation.js
@@ -366,7 +366,11 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
           "Content-Type": "application/octet-stream",
         }),
         credentials: "include",
-      }
+      },
+      // DEBUG
+      // DEBUG
+      // DEBUG
+      Math.floor(Math.random() * 100)
     );
 
     if (response.ok) {
--- a/client/src/components/autosave/index.js
+++ b/client/src/components/autosave/index.js
@@ -38,8 +38,8 @@ class Autosave extends React.Component {
   }
 
   tick = () => {
-    const { dispatch, saveInProgress } = this.props;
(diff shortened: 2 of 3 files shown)


client/src/actions/annotation.js
```diff
@@ -325,15 +325,12 @@ export const saveObsAnnotationsAction = () => async (dispatch, getState) => {
   const state = getState();
   const { annotations, autosave } = state;
   const { dataCollectionNameIsReadOnly, dataCollectionName } = annotations;
-  const {
-    lastSavedAnnoMatrix,
-    saveInProgress,
-    exponentialBackoffFetch,
-  } = autosave;
+  const { lastSavedAnnoMatrix, exponentialBackoffFetch } = autosave;
```
tihuan, line 328: remove `saveInProgress` from `if (saveInProgress || annoMatrix === lastSavedAnnoMatrix) return;`, so `exponentialBackoffFetch()` gets called synchronously to allow either cancel timeout if `hash` has changed, or just return the resolved/rejected promise

client/src/components/autosave/index.js
```diff
@@ -38,8 +38,8 @@ class Autosave extends React.Component {
   }
 
   tick = () => {
-    const { dispatch, saveInProgress } = this.props;
-    if (this.needToSave() && !saveInProgress) {
```
tihuan, line 42: limit the knowledge and responsibility of `tick()` to only checking `this.needToSave()`, since `exponentialBackoffFetch()` already knows to return inflight promise

Additionally, we rely on `tick()` to invoke `exponentialBackoffFetch()` to cancel timeout if `hash` changes

client/src/components/autosave/index.js
```diff
@@ -52,10 +52,13 @@ class Autosave extends React.Component {
 
   statusMessage() {
     const { error, saveInProgress } = this.props;
+
+    if (saveInProgress) return `Saving...`;
```
tihuan, line 56: show `Saving...` first, so we show the correct saving state in the following condition:

1. After a hash fails to save after 3 retries (which populates error state) 
1. We make a new change 
1. Save status is now `Saving...` again instead of `Autosave error: Foo`

client/src/util/exponentialBackoffFetch.js
```diff
@@ -1,68 +1,56 @@
 export default function createBackoffFetch(retryTimes = 3) {
   let prevHash = null;
   let retryCount = 0;
-  let response = null;
```
tihuan, line 4: since we already capture `promise` and `resolve/reject` it accordingly, the `promise` will store its `success/error` state for us!

client/src/util/exponentialBackoffFetch.js
```diff
@@ -1,68 +1,56 @@
...
 
   return (hash, ...fetchArgs) => {
     if (prevHash !== hash) {
       prevHash = hash;
       retryCount = 0;
-      response = null;
-      error = null;
+
+      if (timeoutId) {
+        clearTimeout(timeoutId);
```
tihuan, line 13: clearing timeout when hash changes

client/src/util/exponentialBackoffFetch.js
```diff
@@ -28,29 +28,43 @@ export default function createBackoffFetch(retryTimes = 3) {
...
-          } catch (error) {
-            if (retryCount === retryTimes) {
-              reject(error);
-            }
-
-            retryCount += 1;
+        if (retryCount === 0) {
+          fetchAndHandle();
+          // delay amount (3.5, 17.5, 87.5, 437.5 seconds)
+        } else timeoutId = setTimeout(fetchAndHandle, 700 * 5 ** retryCount);
```
seve, line 34: Not timing out the first call

client/src/util/exponentialBackoffFetch.js
```diff
@@ -28,29 +28,43 @@ export default function createBackoffFetch(retryTimes = 3) {
...
+      }
+      async function fetchAndHandle() {
+        try {
+          promise.response = await mockFetch(...fetchArgs);
```
seve, line 38: Just adding the response as an attribute of the promise itself

client/src/util/exponentialBackoffFetch.js
```diff
@@ -28,29 +28,43 @@ export default function createBackoffFetch(retryTimes = 3) {
...
 
-            delayFetch();
+          if (promise.response.ok || retryCount === retryTimes) {
+            resolve(promise.response);
           }
```
seve: Early exit if this is the last retry

client/src/util/exponentialBackoffFetch.js
```diff
@@ -28,29 +28,43 @@ export default function createBackoffFetch(retryTimes = 3) {
...
-          // delay amount (3.5, 17.5, 87.5, 437.5 seconds)
-        }, retryCount * 700 * 5 ** retryCount);
+          console.log("success", fetchArgs[2]);
+        } catch (error) {
+          console.log("error", fetchArgs[2]);
+          if (retryCount === retryTimes) {
+            reject(error);
+          }
```
seve, line 48: We're only going to serve fetch API error if it's the one that pushes us over the retry amounts.  (Maybe we serve a list of fall the errors instead?)

seve (approved):

PR #2014, opened 2021-01-04
Refactor to support configurable adaptor classes

bmccandless:
A new class can be configured to handle cxg datasets or h5ad datasets.
A type called CZI_cxg_adaptor is implemented to handle specializations needed
for interactions between explorer and portal.
For example: remapping path names to dataset locations, updated corpora_props, and genesets.

So far the CZI_cxg_adaptor is just a stub.
This patch is meant to do the refactoring needed to support specialized adaptor classes.

#### Reviewers
**Functional:** @mckinsel 

**Readability:** @MDunitz 

---

## Changes
- add
- remove
- modify


server/test/unit/compute/test_diffexp_cxg.py
```diff
@@ -19,8 +19,8 @@ class DiffExpTest(unittest.TestCase):
 
     def load_dataset(self, path, extra_server_config={}, extra_dataset_config={}):
         config = app_config(path, extra_server_config=extra_server_config, extra_dataset_config=extra_dataset_config)
-        loader = MatrixDataLoader(path)
-        adaptor = loader.open(config)
+        loader = MatrixDataLoader(path, config)
```
MDunitz, line 22: It might be useful to add a test where the dataset_config and matrix_data_type are passed in, and even one with an incorrect data type

server/test/performance/run_diffexp.py
```diff
@@ -36,8 +36,8 @@ def main():
     app_config.update_server_config(app__verbose=True)
     app_config.complete_config()
 
-    loader = MatrixDataLoader(args.dataset)
-    adaptor = loader.open(app_config)
+    loader = MatrixDataLoader(args.dataset, app_config)
```
MDunitz, line 39: you could also pass the dataset_config in here, depending on which path you are trying to test

server/test/unit/compute/test_diffexp_cxg.py
bmccandless: Yes, I can write a test case for that.  I'll put it in test/unit/data_common/test_matrix_loader.py

server/test/performance/run_diffexp.py
bmccandless: Yes, I could pass in the dataset_config, but in this isn't so much of a test case as a script to test performance.  Since there is only the default dataset_config in this script then we can let the MatrixDataLoader get it from the app_config.

server/data_common/__init__.py
```diff
@@ -0,0 +1,10 @@
+# import the data adaptor types so they can be registered
+
+import server.data_cxg.cxg_adaptor  # noqa: F401
+import server.data_anndata.anndata_adaptor  # noqa: F401
+
+# FIXME: may want to make this one a plugin so that hosted CZI specific
```
MDunitz, line 6: nit can you make this a TODO instead?

bmccandless:
We decided to go a different way.

Issue #1888, opened 2020-09-28
P value and Log fold change not showing up for all DE results


signechambers1:
Steps to reproduce:

1. Go to https://cellxgene.cziscience.com/e/Single_cell_drug_screening_a549-42-remixed.cxg/
2. Select cluster -> 1 as group 1
3. Select cluster -> 7 as group 2
4. Run DE

Expected results: All genes that show up as a result show the log fold change and p-value.

Actual result: None of the genes show the log fold change or p-value.

Note: if you select cluster 12 as group 2, 6 of the genes have the p-value and log fold change.

signechambers1:
![results from clusters 1 and 7.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/1543f161-1837-4aa4-8887-d54d119fc47a)![results from clusters 1 and 12.png](https://images.zenhubusercontent.com/5f47df0cc52f8855d9f0c5d5/db5cb9b2-fb74-4226-a096-1cedd036ddbc)

bmccandless:
This should be an easy fix:
in src/components/brushableHistogram/index.js
it does this check before outputting the log fold change:
```


        {logFoldChange && pvalAdj ? (
          <div
            style={{
              display: "flex",
              justifyContent: "center",
              alignItems: "baseline",
            }}
          >
            <span>
              <strong>log fold change:</strong>
              {` ${logFoldChange.toPrecision(4)}`}
            </span>
      ...

```

So if pvalAdj is 0 then we don't show the logfold change.
In the example given in this ticket (cluster=1 vs cluster=7), the pval is indeed 0:

```
[[309, -2.6433210372924805, 0.0, 0.0], [48378, -2.9396235942840576, 0.0, 0.0], [7289, -2.4838926792144775, 0.0, 0.0], [7986, -2.5104591846466064, 0.0, 0.0], [36549, 1.0494585037231445, 0.0, 0.0], [10023, -2.550673723220825, 0.0, 0.0], [2581, -1.9949843883514404, 0.0, 0.0], [8406, -3.1643800735473633, 0.0, 0.0], [14030, -2.3969335556030273, 0.0, 0.0], [15737, -2.5326099395751953, 0.0, 0.0]]
```

But for cluster=1 vs cluster=12 we see some (very small) but non zero pvals:

```
[...]

PR #2016, opened 2021-01-07
P value and Log fold change not showing up for all DE results

bmccandless:
The HistogramFooter needs to distinguish between an undefined
value and a value of 0.  If the pvalAdj was 0, then the logFolChange
was previously not showing up.

 #1888

#### Reviewers
**Functional:** @seve 

**Readability:**


seve (approved):
LGTM! Thanks, @bmccandless

Issue #2019, opened 2021-01-12
cellxgene scale tests are currently not running


maniarathi:
Due to some import errors, the weekly cellxgene scale tests are not running (example [run](https://github.com/chanzuckerberg/cellxgene/runs/1675256660?check_suite_focus=true)). This needs to be fixed.

PR #2020, opened 2021-01-12
Fix locust test.

bmccandless:
The locust test deployment hook has been failing.
The problem was that locust is meant to just test our API and therefore requires a
limited set of modules.

However it imports a few things from server, and the modules it traverses and
the files it import, then import other pieces of the code base, that in turn bring in more than it should.

There were two main issues:
1.  the decode_matrix_FBS helper function in the server/test/unit directory has imported by locustfile.py
which pulled in many things from server/test/__init__.py.  There were a few ways to address this issue.
One was to move everything out of __init__.py to somewhere else.  This would have required a large number
of changes in the test code.  The second option was to avoid importing anything from server/test directory, which
is what I chose.  I moved the decode_matrix_FBS functionality into data_common/fbs.  This is an okay
choice, in my opinion, because there is already existing code there that decodes an FBS into a
pandas dataframe, instead of a dict.

2.  The server/__init__.py checks for plugins.  The code for this was written in the server/common/utils/utils.py
module with several other functions.  I split out the plugins handling part to its own module to reduce
dependencies.

 #2019

#### Reviewers
**Functional:** @MDunitz 

**Readability:** @mckinsel


server/test/unit/decode_fbs.py
```diff
@@ -1,31 +0,0 @@
-"""
-Code to decode, for testing purposes, the flatbuffer encoded blobs.
-
-This code will need to be updated if fbs/matrix.fbs changes. For more information, see fbs/matrix.fbs and
-server/data_common/fbs/
-"""
-
-import server.data_common.fbs.NetEncoding.Matrix as Matrix
-from server.data_common.fbs.matrix import deserialize_typed_array
```
MDunitz, line 9: it looks like this function was added to `server/data_common/fbs/fbs_core.py` does it need to be removed from ` server.data_common.fbs.matrix`? or could it be reused?

bmccandless: Thanks, good catch.  I had meant to move it, not duplicate it.

MDunitz (approved):
lgtm

mckinsel:
Is it just the import in locustfile.py that causes trouble? If so you could do a one-off `importlib` thing there and leave everything else alone.

bmccandless:
> Is it just the import in locustfile.py that causes trouble? If so you could do a one-off `importlib` thing there and leave everything else alone.

@mckinsel, yes in locustfile.py it is the import of "server.test.unit.decode_fbs" that is the problem.  I'm not sure how to use importlib to avoid executing the __init__.py files and partially importing parts of other modules.  Let me know if you know how.
Another solution that is  simpler is to change the requirements-locust.py to include a lot more.  We could just  use server/requirements.txt instead.  That would definitely work, but it's a bit ugly in a way since the test doesn't  need most of that stuff.

PR #2021, opened 2021-01-13
update umap version

bmccandless:
There is an interface change in 0.5.0 which is not compatible with our
version of scanpy.

#### Reviewers
**Functional:**   @MDunitz


MDunitz (approved):
lgtm

PR #2023, opened 2021-01-17
Split backends WIP

mckinsel:


mckinsel:
Per conversation with the team, I'm going to close the PR and open a new one with a slightly different strategy on git history. This PR attempts a couple tricks to keep the history intact for each version of each file between local and hosted versions. But, this kind of screws up the overall history. Instead, I'm going to do a big `cp -R` and commit it. This will create a bunch of brand new files as far as git is concerned, but I'll leave a note in the commit message so people know where history/blame picks up again.

PR #2026, opened 2021-01-20
simple solution to the locust test problem

bmccandless:
In this solution, all the server requirements are installed.
This is a slightly overkill, but it avoid having to restructure
any of the server or test code to avoid unnecessary imports.

 #2019

#### Reviewers
**Functional:**  @MDunitz 

**Readability:**  @mckinsel


mckinsel (approved):
LGTM

MDunitz (approved):
lgtm

PR #2035, opened 2021-02-03
Fix deprecated np.unicode type

mckinsel:
Until numpy version 1.20.0, `numpy.unicode` was an alias for `str` in python3. In 1.20.0, it's fully deprecated and is an `int`. This is bad and breaks things. This commit drops the `np.unicode` alias and just uses `str`, as is advised here:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

#### Reviewers
@MDunitz


MDunitz (approved):
lgtm

PR #2052, opened 2021-02-05
Split out the local backend

mckinsel:
## Changes
This splits the backend into two parts: the local backend for desktop cellxgene and the AWS backend for hosted cellxgene. The local backend is in `local_server` while the hosted remains in `server`. The general idea is to copy everything from `server` to `local_server`, pull unneeded stuff out of `local_server`, and keep `server` as-is for this PR. Not touching `server` means all the infra and deployment code will continue working just as it did before so we can make those changes incrementally.

---

#### Reviewers
This is pretty expansive, so I think everybody is going to take a look.


local_server/app/app.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class HealthAPI(Resource):
+    def get(self):
+        config = current_app.app_config
+        return health_check(config)
+
+
+class SchemaAPI(Resource):
+    # TODO @mdunitz separate dataset schema and user schema
```
maniarathi, line 80: @mckinsel Sorry, I have a rather annoying request.. for each of the TODOs in the comments in this PR (I count 9 I think), would you mind creating a ticket for it and tagging it with the "Dev-ish" label? That way we don't lose the todos in the code forever.

Makefile
```diff
@@ -79,12 +79,18 @@ fmt-py:
...
-	flake8 server --per-file-ignores='server/test/fixtures/dataset_config_outline.py:F821 server/test/fixtures/server_config_outline.py:F821 server/test/performance/scale_test_annotations.py:E501'
+lint-server: lint-local-server lint-hosted-server
 
+.PHONY: lint-local-server
+lint-local-server: fmt-py
+	flake8 local_server --per-file-ignores='local_server/test/fixtures/dataset_config_outline.py:F821 local_server/test/fixtures/server_config_outline.py:F821 local_server/test/performance/scale_test_annotations.py:E501'
+
+.PHONY: lint-hosted-server
+lint-hosted-server: fmt-py
+	flake8 hosted_server --per-file-ignores='hosted_server/test/fixtures/dataset_config_outline.py:F821 hosted_server/test/fixtures/server_config_outline.py:F821 hosted_server/test/performance/scale_test_annotations.py:E501'
```
bkmartinjr, line 93: Is `hosted_server` correct?  Or should it be `server`?  Ditto for other targets, eg, clean

Makefile
```diff
@@ -46,11 +63,17 @@ copy-client-assets:
...
 test: unit-test smoke-test
 
+.PHONY: test-local
+test-local: unit-test-local smoke-test
+
+.PHONY: unit-test-local
+unit-test-local: unit-test-local-server unit-test-client
+
 .PHONY: unit-test
 unit-test: unit-test-server unit-test-client
```
bkmartinjr, line 73: `make unit-test` no longer works.

bkmartinjr:
@mckinsel - it looks like there is a bunch of lint in `local_server`.  May just need to run `make fmt`.

```
$ flake8 local_server --per-file-ignores='local_server/test/fixtures/dataset_config_outline.py:F821 local_server/test/fixtures/server_config_outline.py:F821 local_server/test/performance/scale_test_annotations.py:E501'
local_server/converters/schema/remix.py:219:1: E302 expected 2 blank lines, found 1
local_server/converters/schema/remix.py:225:1: E302 expected 2 blank lines, found 1
local_server/test/unit/converters/schema/test_validate.py:203:5: E303 too many blank lines (2)
local_server/test/unit/converters/schema/test_validate.py:354:5: E303 too many blank lines (2)
local_server/test/unit/converters/schema/test_remix.py:54:1: E302 expected 2 blank lines, found 1
local_server/test/unit/converters/schema/test_remix.py:193:121: E501 line too long (123 > 120 characters)
local_server/test/unit/converters/schema/test_remix.py:195:121: E501 line too long (124 > 120 characters)
local_server/test/unit/common/test_colors.py:4:121: E501 line too long (122 > 120 characters)
```

Makefile
bkmartinjr: ditto for `make test`

mckinsel:
@bkmartinjr those are all in `server` too, which hasn't changed from this PR. I think the issue is we run `black` and _then_ `flake8`, which always passes.

local_server/app/app.py
mckinsel: Yes 😐

local_server/auth/auth_test.py
```diff
@@ -0,0 +1,73 @@
+from local_server.auth.auth import AuthTypeClientBase, AuthTypeFactory
+from flask import session, request, redirect
+
+
+class AuthTypeTest(AuthTypeClientBase):
```
bkmartinjr, line 5: Why is this code in the main source tree, and enabled for use via the config (rather than being part of the test framework only)?  It seems prone to possible issues if the user accidentally configures the test auth adaptor.

bkmartinjr: Alternatively, the registration of the test adaptor could be predicated on some other signal that it is actually being used in a test environment.  I'd just like to see it completely disabled in PROD environments for safety.

local_server/common/aws_secret_utils.py
```diff
@@ -0,0 +1,23 @@
+import logging
```
bkmartinjr, line 1: Why does the code base need the AWS secret manager?  (same question for the config code that uses it - eg, `aws_secrets_manager` config in `common/config/external_config.py`)?

local_server/common/config/client_config.py
```diff
@@ -49,8 +49,8 @@ def get_client_config(app_config, data_adaptor):
         "annotations_cell_ontology_terms": None,
         "custom_colors": dataset_config.presentation__custom_colors,
         "diffexp-may-be-slow": False,
-        "about_legal_tos": dataset_config.app__about_legal_tos,
-        "about_legal_privacy": dataset_config.app__about_legal_privacy,
+        "about_legal_tos": None,
+        "about_legal_privacy": None,
```
MDunitz, line 53: Because the client uses optional chaining when accessing config params I think you can just remove this. But it would be good to check with @tihuan or @colinmegill

tihuan: Thanks for tagging!

Since DP is now rendering ToS and Privacy, I think we can remove anything related to ToS and Privacy pages in `cellxgene`!

local_server/auth/auth_test.py
mckinsel: So it might be tough to detect if we're in a prod environment at that point in the code, but I threw together a [commit](https://github.com/chanzuckerberg/cellxgene/pull/2052/commits/f1aa655adcd0bca4684583b347b85338c0ce8b97) with an "insecure_test_environment" config setting that must be set to true or it will refuse to use AuthTypeTest. Does that help?

local_server/common/aws_secret_utils.py
mckinsel: We wanted to continue supporting S3 access from the local server. I thought there could be use cases where you set some secrets via secretsmanager and didn't want to break that. Happy to take it out if that seems far-fetched though.

local_server/common/config/client_config.py
mckinsel: Sounds good, removed it.

tihuan: Awesome thanks a lot! 🤩

maniarathi (approved):
LGTM

PR #2065, opened 2021-02-16
Fix supported schema encoding check to allow version 0.2.0

pablo-gar:
## Changes
Fixes `server/common/corpora.py` to support [corpora encoding version 0.2.0](https://github.com/chanzuckerberg/corpora-data-portal/blob/main/backend/schema/corpora_schema_h5ad_implementation.md) which was implemented in the schema version 1.1.0

### DIFF
--- a/server/common/corpora.py
+++ b/server/common/corpora.py
@@ -41,7 +41,8 @@ def corpora_is_version_supported(corpora_schema_version, corpora_encoding_versio
         corpora_schema_version
         and corpora_encoding_version
         and corpora_schema_version.startswith("1.")
-        and corpora_encoding_version.startswith("0.2.")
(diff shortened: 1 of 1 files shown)


tihuan:
Looking at the test output, it seems like the test is expecting a few fields to be present!

Screen Shot 2021-02-17 at 9 37 45 AM

mckinsel (changes requested):
Hey this will break cellxgene for everything with 0.1.0 right? That's why the tests are breaking.

pablo-gar:
Yes I should've added both

pablo-gar:
After a discussion a better solution is to drop the `corpora encoding version` slot and any validators as well as tests associated with it.

PR #2078, opened 2021-02-24
fix local_server unit-test target

bkmartinjr:
Removed obsolete local_server Makefile targets related to the annotations test DB.


MDunitz (approved):
lgtm

PR #2087, opened 2021-03-02
Dunitz/2072 shared modules

MDunitz:
#### Reviewers
**Functional:** 
- @bkmartinjr 

**Readability:** 
- @maniarathi 

---

## Changes
- add - shared_utils directory and move code shared between the local server and the hosted server into it. 
Updated tests/imports
Update Make recipes to include shared_utils in the build

## Question
- There were a few shared files/functions that only differ because they are pulling in different versions of the same error -- but it felt like an anti pattern to have errors be shared as well. Any thoughts? or objections to moving errors into shared_utils?


shared_utils/corpora.py
```diff
@@ -7,10 +7,30 @@
 """
 import collections
 import json
+import re
 
-from server.cli.upgrade import validate_version_str
-from server.common.utils.corpora_constants import CorporaConstants
+from shared_utils.utils.corpora_constants import CorporaConstants
 
+# Official SemVer regex: https://semver.org/
```
maniarathi, line 14: I think this chunk of code is specifically for the desktop version only since we use it to bump the PyPi release so I don't think it makes sense to move it to shared.

maniarathi (changes requested):
Couple of high level things:
- I would keep the stuff related to anndata and CXG where they are because ultimately we are going to delete all anndata support from `server` and the CXG stuff has already been deleted from `local_server`. This mean leaving `diffexp_generic.py` in both (since it will later be deleted from `server`), `scanpy.py` in both (will be deleted from `server` since we don't support re-embeddings on hosted). 
- Another thing is that we'd like to consolidate testing for the shared utilities in a folder. You could opt to do that here or write another issue to de-dupe unit tests for the shared utilities and ensure that the unit tests for shared utilities runs as part of the unit test make commands.

bkmartinjr:
Given the current plan to put all back-end related code in a single directory, do you want to start that by moving "shared_utils" to that new location?  (eg, renamed shared_utils to "backend/shared_utils"

Also, I suggest using the name "common" rather than "shared_utils", as that is what we use elsewhere in the code base as a name for "shared" stuff.

bkmartinjr:
> objections to moving errors into shared_utils?

I don't have any objection to unifying errors/exceptions as long as they are common to both code bases.  if truly unique to one or the other, we should probably separate (ie, allow the back-ends to add their own specialized exceptions in addition to the common code).

My expectation is that the functions you found that differ only by exception thrown could be unified.  Did you see any cases where that wasn't true?

local_server/common/config/dataset_config.py
```diff
@@ -4,7 +4,7 @@
 from local_server.common.annotations.local_file_csv import AnnotationsLocalFile
 from local_server.common.config.base_config import BaseConfig
 from local_server.common.errors import ConfigurationError, OntologyLoadFailure, AnnotationsError
-from local_server.compute.scanpy import get_scanpy_module
+from shared_utils.compute.scanpy import get_scanpy_module
```
bkmartinjr, line 7: I am fairly certain the scanpy-dependent features (re-embedding) are only supported in the local/pip-installed back-end, and are not destined for the hosted (or at least not with the same scanpy-derived implementation).  So `scanpy.py` is probably not shared/common code.

bkmartinjr:
One other thought: are there other modules or base classes common?  Candidates with my viewpoint:

* auth - seems like there might be some value in keeping this code common IF the 'local' version requires identity.  I haven't yet internalized the implementation details enough to know if this is a forced fit, or really a win.
* config - the base framework could be common, but the bulk of the code is in the config-variable-specific handling.  So probably not worth sharing IMHO.
* converters - seems like a lot of this is likely shared, IF the local client continues to support commands like `cellxgene schema`.  @maniarathi - is the plan to leave CZI-specific schema support in both?

PR #2088, opened 2021-03-02
rename "geneset" to "gene set" in CLI

bkmartinjr:
#### Reviewers
**Functional:** @signechambers1 

**Readability:** @MDunitz 

---

In the CLI, we want to use "gene sets", not "genesets".  This PR changes:
* the CLI params are changed
* the CSV file header is changed
* the config file names are changed
* and a bunch of misc code in the local_server backend are changed to match

Note that I did not change _all_ instances of "genesets", in particular those where only API/internals are affected.


MDunitz (approved):
lgtm -- does this spelling also apply in corpora?

bkmartinjr:
> lgtm -- does this spelling also apply in corpora?

No idea - that is a question for @signechambers1

signechambers1:
@MDunitz yes, I'd recommend we use "gene set" instead of "geneset" everywhere across the applications for consistency. fyi @brianraymor

brianraymor:
Portal uses "gene set" in all our UI and that's also what I presented to the curators earlier in the week.

Issue #2059, opened 2021-02-09
Gene sets doubles the effort for a user to color by an individual gene


signechambers1:
“Add gene” or similar search bar in the right side bar. A user can add a single gene and color by, plot on x or y, expand and brush on.

The single gene stays in the upper right side bar until another gene is added, the gene is cleared, or the session ends (ie the single gene is not persisted). 

Future: Gene can be added to a gene set w/o retyping the gene name, we discussed dragging or add button. I do not think this is something we need to incorporate for the desktop release (they can retype the gene name to add to a gene set if they want. We can add this and optimize the flow for hosted later.

final figma: https://www.figma.com/file/OmZwf8p1qa4nG32sbfOfc5/Add-Gene?node-id=58%3A27

Design Due Date: 6/11

hthomas-czi:
@signechambers1 I can't update the issue itself but here's the [up-to-date Figma](https://www.figma.com/file/OmZwf8p1qa4nG32sbfOfc5/Add-Gene?node-id=58%3A27).

signechambers1:
Thans @hthomas-czi I updated the description with the new figma link.

PR #2097, opened 2021-03-08
Gene sets UI, right sidebar refactor

colinmegill:
This PR completely rewrites the user interface for the right sidebar. The new functionality includes:

* Adding a gene set
* Deleting a gene set
* Adding genes to gene set
* Deleting genes from gene set
* Collapsing and expanding genes within gene set for easy skimming / brush selection

There is UI infrastructure in place for coloring by the aggregate of genes, this will be implemented in a subsequent PR.


client/src/actions/index.js
```diff
@@ -161,7 +161,7 @@ const dispatchDiffExpErrors = (dispatch, response) => {
   }
 };
 
-const requestDifferentialExpression = (set1, set2, num_genes = 10) => async (
+const requestDifferentialExpression = (set1, set2, num_genes = 50) => async (
```
bkmartinjr, line 164: This is a big UI change (I think -- assuming it now displays top 50 genes, rather than top 10).  

@signechambers1 - you good with this?

@colinmegill - in the future, could we break unrelated stuff like this into its own PR?

client/src/components/brushableHistogram/clamp.js
```diff
@@ -0,0 +1,3 @@
+export default function clamp(val, rng) {
+  return Math.max(Math.min(val, rng[1]), rng[0]);
+}
```
bkmartinjr, line 3: given how generic this is, the file should be moved into the `src/util/` directory

package.json
```diff
@@ -2,5 +2,8 @@
   "name": "cellxgene",
   "scripts": {
     "postinstall": "npm install --prefix client && npm run build --prefix client && make copy-client-assets"
+  },
```
bkmartinjr, line 5: why is there a package.json at both top-level and in client/ directory?

client/src/reducers/index.js
```diff
@@ -16,6 +16,7 @@ import layoutChoice from "./layoutChoice";
 import controls from "./controls";
 import annotations from "./annotations";
 import genesets from "./genesets";
+import genesetsUI from "./genesetsUI";
```
bkmartinjr, line 19: I noticed you did not add genesetsUI to the undoable state list.  Is that intentional?  

I suggest confirming that undo/redo works correctly with genesets, as I suspect it is broken without this.

colinmegill: You cannot click undo and redo while the modals are popped, and this reducer only controls the `show` and `cancel/hide` for the modals

bkmartinjr: gotcha - so as long as undo/redo is confirmed to otherwise work, it sounds like all is good....

package.json
colinmegill: Removed.

bkmartinjr: oh, weird - there is also a package.json at the top level.   And they are also in main (you didn't add them).

This came from a PR @MDunitz added a few months ago.  #1427     @MDunitz - any idea why we need two package.jsons?

colinmegill: It's all good — this may have been a misfire — I have done that. If you run npm start in the wrong folder, it will blow up. If you run `npm install foo`, it will start implicitly creating things and add a lock file. That can cause some confusion and may have been what happened?

client/src/reducers/genesetsUI.js
```diff
@@ -0,0 +1,54 @@
+/*
```
bkmartinjr, line 1: this reducer needs a unit test in `client/__tests__/reducer/`

colinmegill: Yup, incoming

client/src/components/brushableHistogram/index.js
```diff
@@ -664,16 +266,20 @@ class HistogramBrush extends React.PureComponent {
 
   // eslint-disable-next-line class-methods-use-this -- instance method allows for memoization per annotation
   calcHistogramCache(col, margin, width, height) {
+    /* make this more structured and doing a forEach */
```
maniarathi, line 269: Is this still a relevant comment or..?

client/src/components/brushableHistogram/maybeScientific.js
```diff
@@ -0,0 +1,17 @@
+import significantDigits from "../../util/significantDigits";
+
+export default function maybeScientific(x) {
```
maniarathi, line 3: Should this file also belong in the `/util` directory?

client/src/reducers/differential.js
```diff
@@ -47,14 +47,6 @@ const Differential = (
         celllist1: null,
         celllist2: null,
       };
-    case "reset subset":
```
bkmartinjr, line 50: Can you explain the goal and side-effects of this change?   My guess is that the intent is to not delete the gene list created by the diffexp request) when doing a subset/reset.

However, this change has at least one other side effect -- upon a "subset to selection", the cell lists will now potentially contain cells not visible.  I'm not sure what side effects that will  have, but it violates our past assumptions about the cell lists always being a subset of the currently viewable cells (aka "world").  I am suspicious it will lead to future bugs.

Why not leave the previous celllist clear code, but remove the diffExp clear?  That also has potential for confusion, in that the statistics (eg, p-val, ...) will no longer represent the diffexp result of the current cell sets, but at least it won't break the world/universe constraints.

TL;DR - this change needs some thought, as it violates some constraints I thought we considered important.

client/src/components/categorical/value/index.js
```diff
@@ -677,8 +677,8 @@ class CategoryValue extends React.Component {
               display={isColorBy && categoryValueIndices ? "auto" : "none"}
               style={{
                 marginLeft: 5,
-                width: VALUE_HEIGHT,
-                height: VALUE_HEIGHT,
+                width: 15,
```
maniarathi, line 680: Can we set the `VALUE_HEIGHT` variable directly instead of hardcoding?

client/src/components/geneExpression/index.js
```diff
@@ -1,59 +1,83 @@
...
+  };
+
+  handleActivateCreateGenesetMode = () => {
+    const { dispatch } = this.props;
+    dispatch({ type: "geneset: activate add new geneset mode" });
+  };
+
   render() {
-    const { userDefinedGenes, differential } = this.props;
+    const geneSetsFeatureEnabledTODO = true;
```
maniarathi, line 61: This const is never false? So do we need it?

client/src/reducers/colors.js
```diff
@@ -66,6 +66,7 @@ const ColorsReducer = (
       };
     }
 
+    case "color by gene set":
```
bkmartinjr, line 69: is this implemented?  Is color-by-gene-set any different than color-by-continuous?

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,199 @@
...
+        /* Unlike bulk add... with gene sets this test needs to be within the set... */
+        // if (upperUserDefinedGenes.get(upperGene) !== undefined) {
+        //   return keepAroundErrorToast(`${upperGene} already exists`);
+        // }
+
+        const indexOfGene = upperWorldGenes.get(upperGene);
+
+        if (indexOfGene === undefined) {
+          console.log("found a gene that doesn't appear to be a valid name");
+          // return;
```
bkmartinjr, line 67: extraneous comments need cleanup

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,199 @@
...
+
+    const upperGenes = this._genesToUpper(setGenes);
+    const upperWorldGenes = this._memoGenesToUpper(worldGenes);
+
+    dispatch({ type: "bulk user defined gene start" });
+
+    Promise.all(
+      [...upperGenes.keys()].map((upperGene) => {
+        /* Unlike bulk add... with gene sets this test needs to be within the set... */
+        // if (upperUserDefinedGenes.get(upperGene) !== undefined) {
```
bkmartinjr, line 59: cleanup

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,199 @@
+// jshint esversion: 6
```
bkmartinjr, line 1: Unless I'm mistaken, you don't need to add these jshints anymore.  They are ignored.

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,199 @@
...
+    const upperWorldGenes = this._memoGenesToUpper(worldGenes);
+
+    dispatch({ type: "bulk user defined gene start" });
+
+    Promise.all(
+      [...upperGenes.keys()].map((upperGene) => {
+        /* Unlike bulk add... with gene sets this test needs to be within the set... */
+        // if (upperUserDefinedGenes.get(upperGene) !== undefined) {
+        //   return keepAroundErrorToast(`${upperGene} already exists`);
+        // }
```
maniarathi, line 61: Delete?

client/src/components/geneExpression/menus/createGenesetDialogue.js
```diff
@@ -0,0 +1,146 @@
+import _ from "lodash";
+import React from "react";
+import { connect } from "react-redux";
+import AnnoDialog from "../../categorical/annoDialog";
+import LabelInput from "../../categorical/labelInput";
+// import LabelInput from "../labelInput";
+// import { genesetPrompt, isGenesetErroneous } from "../genesetUtil";
```
maniarathi, line 7: Delete

client/src/components/geneExpression/menus/createGenesetDialogue.js
```diff
@@ -0,0 +1,146 @@
...
+    });
+    this.setState({
+      genesetName: "",
+      genesToPopulateGeneset: "",
+    });
+    e.preventDefault();
+  };
+
+  genesetNameError = () => {
+    /* todo genesets validation */
```
maniarathi, line 76: Is this going to be done in a later PR? Is the issue tracked somewhere? If so, mind adding the issue number here in the comments so that we have a paper trail that this will be implemented in a subsequent PR?

client/src/components/geneExpression/menus/createGenesetDialogue.js
```diff
@@ -0,0 +1,146 @@
...
+    this.setState({ genesetName: e });
+  };
+
+  handleGenesetInputChange = (e) => {
+    this.setState({ genesToPopulateGeneset: e });
+  };
+
+  instruction = () => {
+    return "New, unique geneset";
+    /* todo genesets */
```
maniarathi, line 90: I think there's a bit of cleanup here along with line 109 below. The TODOs have similar content.

client/src/components/geneExpression/menus/editGenesetNameDialogue.js
```diff
@@ -0,0 +1,104 @@
+import React from "react";
+import { connect } from "react-redux";
+import AnnoDialog from "../../categorical/annoDialog";
+import LabelInput from "../../categorical/labelInput";
+// import LabelInput from "../labelInput";
+// import { genesetPrompt, isGenesetErroneous } from "../genesetUtil";
```
maniarathi, line 6: Delete

client/src/components/geneExpression/menus/createGenesetDialogue.js
colinmegill: cc @signechambers1

client/src/reducers/colors.js
colinmegill: The UI is there, but the gene `aggregate` functionality is still outstanding and won't be resolved in this PR.

maniarathi (commented):
Couple general comments:
- Can you do a quick sweep over the PR and delete any commented out lines of code please?
- For TODOs, I'm having a hard time figuring out whether there are bits of missing implementation in this PR or whether you actually mean to do them later in a separate PR. Can you sweep through your TODOs and mark the ones that actually will be implemented later with a Github issue number for tracking purposes? Thank you!

client/src/reducers/differential.js
colinmegill: Agreed, this should be discussed fully in an issue such as `reconsider diffexp now that we have gene sets and the results of diffexp are displayed in a gene set`. I'm not tied to diffexp functionality in this PR, such as changing the number of genes returned, but @signechambers1 and I should work together I think to capture where we're headed.

client/src/reducers/index.js
colinmegill: I believe you are correct @bkmartinjr, it doesn't break it completely (I didn't notice before) but if you undo/redo past the point of popping the modal, there is an error in console. Do you repro?

client/src/reducers/colors.js
bkmartinjr: Unless I'm mistaken, triggering this action will cause an error.  I would remove this until it is implemented.

client/src/reducers/differential.js
bkmartinjr: OK, seems reasonable. So are you going to revert all the diffexp stuff in this PR, or do we need to resolve the design issues before landing it?

colinmegill: I'd rather revert. cc @signechambers1

bkmartinjr: fully concur (ditto for the number of results). We can turn this around in another PR quickly, if we know what behavior we want.

client/src/components/geneExpression/menus/createGenesetDialogue.js
bkmartinjr: Validation of names should be super trivial now that we decided to allow (more or less) any unicode character.   The backend has not yet accommodated this, but I think validation should be in the front-end.  Primarily it is simple stuff like:
* no duplicate genes in a gene set
* gene set name is unique
* gene set name does not include running or leading/trailing spaces (at least I think this is the new requirement)
So, TL;DR - very easy to add to this PR.

bkmartinjr:
@colinmegill - the description says:
> There is UI infrastructure in place for coloring by the aggregate of genes, this will be implemented in a subsequent PR.

I suspect that any aggregation operations are deferred as well - eg, select by the mean expression, displaying the histo of the mean expression, etc.

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+    const upperWorldGenes = this._memoGenesToUpper(worldGenes);
+
+    dispatch({ type: "bulk user defined gene start" });
+
+    Promise.all(
+      [...upperGenes.keys()].map((upperGene) => {
+        const indexOfGene = upperWorldGenes.get(upperGene);
+
+        if (indexOfGene === undefined) {
+          console.log("found a gene that doesn't appear to be a valid name");
```
bkmartinjr, line 60: Can you clean up all of the debugging/logging code?   If this should have a UI, it needs to do more than log to console....

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+            justifyContent: "space-between",
+            alignItems: "center",
+          }}
+        >
+          <span
+            role="menuitem"
+            tabIndex="0"
+            data-testclass="gene-expand"
+            data-testid={`${gene}:gene-expand`}
+            onKeyPress={/* todo_genesets */ () => {}}
```
bkmartinjr, line 93: is this TODO captured in an issue, or should it be in this PR?

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+            justifyContent: "space-between",
+            alignItems: "baseline",
+          }}
+        >
+          <span
+            role="menuitem"
+            tabIndex="0"
+            data-testclass="geneset-expand"
+            data-testid={`${setName}:geneset-expand`}
+            onKeyPress={/* todo_genesets */ () => {}}
```
bkmartinjr, line 109: is this TODO captured in an issue, or should it be in this PR?

client/src/components/geneExpression/menus/editGenesetNameDialogue.js
```diff
@@ -0,0 +1,102 @@
...
+      newName: genesetName,
+    });
+    dispatch({
+      type: "geneset: disable rename geneset mode",
+    });
+    e.preventDefault();
+  };
+
+  genesetNameError = () => {
+    /* todo genesets validation */
```
bkmartinjr, line 48: is this TODO captured in an issue, or should it be in this PR?

client/src/components/geneExpression/menus/editGenesetNameDialogue.js
```diff
@@ -0,0 +1,102 @@
...
+    return false;
+  };
+
+  handleChange = (e) => {
+    this.setState({ genesetName: e });
+  };
+
+  instruction = () => {
+    return "New, unique geneset name";
+    /* todo genesets */
```
bkmartinjr, line 58: is this TODO captured in an issue, or should it be in this PR?

client/src/components/geneExpression/menus/geneMenus.js
```diff
@@ -0,0 +1,90 @@
...
+
+@connect(() => ({}))
+class GenesetMenus extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  handleDeleteGeneFromGeneset = () => {
+    // const { dispatch, geneset } = this.props;
```
bkmartinjr, line 21: this (delete action) appears to be unimplemented.  Intentional?

client/src/components/geneExpression/gene.js
colinmegill: It's for accessibility so that users can tab. Here's the jslint error if it is removed. I will capture as an issue.

```
 104:11  error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events

✖ 1 problem (1 error, 0 warnings)
```

colinmegill: https://github.com/chanzuckerberg/cellxgene/issues/2101

client/src/components/geneExpression/geneSet.js
colinmegill: https://github.com/chanzuckerberg/cellxgene/issues/2101

client/src/reducers/colors.js
colinmegill: Removed.

client/src/components/geneExpression/menus/geneMenus.js
colinmegill: Dead code, removed file

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+              <AnchorButton
+                minimal
+                small
+                data-testid={`delete-from-geneset-${gene}`}
+                onClick={this.handleDeleteGeneFromSet}
+                intent="none"
+                style={{ fontWeight: 700, marginRight: 2 }}
+                icon={<Icon icon="trash" iconSize={10} />}
+              />
+            ) : null}
```
seve, line 133:
```suggestion
            {!isDiffexp && (
              <AnchorButton
                minimal
                small
                data-testid={`delete-from-geneset-${gene}`}
                onClick={this.handleDeleteGeneFromSet}
                intent="none"
                style={{ fontWeight: 700, marginRight: 2 }}
                icon={<Icon icon="trash" iconSize={10} />}
              />
            )}
```

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+                    310 /* todo_genesets this magic number determines how much of a long geneset name we see, and will be tweaked as we build */,
+                }}
+                data-testid={`${gene}:gene-label`}
+              >
+                {gene}
+              </span>
+            </Truncate>
+            {!geneIsExpanded ? (
+              <HistogramBrush isUserDefined field={gene} mini />
+            ) : null}
```
seve, line 120:
```suggestion
            {!geneIsExpanded && (
              <HistogramBrush isUserDefined field={gene} mini />
            )}
```

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+              onClick={this.onColorChangeClick}
+              active={isColorAccessor}
+              intent={isColorAccessor ? "primary" : "none"}
+              icon={<Icon icon="tint" iconSize={12} />}
+            />
+          </span>
+          {/* <TestMiniHisto /> */}
+        </div>
+        {geneIsExpanded ? <HistogramBrush isUserDefined field={gene} /> : null}
+        {/* <GeneMenus genesetsEditable gene={gene} /> */}
```
seve: Are these comments meant to hit main?

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+            />
+            <AnchorButton
+              minimal
+              small
+              data-testclass="colorby"
+              data-testid={`colorby-${gene}`}
+              onClick={this.onColorChangeClick}
+              active={isColorAccessor}
+              intent={isColorAccessor ? "primary" : "none"}
+              icon={<Icon icon="tint" iconSize={12} />}
```
seve, line 175: IconName constant

client/src/components/geneExpression/gene.js
```diff
@@ -0,0 +1,191 @@
...
+            </AnchorButton>
+            <AnchorButton
+              minimal
+              small
+              data-testclass="maximize"
+              data-testid={`maximize-${gene}`}
+              onClick={this.handleGeneExpandClick}
+              active={geneIsExpanded}
+              intent="none"
+              icon={<Icon icon="maximize" iconSize={10} />}
```
seve, line 164: IconName constant

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+
+  _genesToUpper = (listGenes) => {
+    // Has to be a Map to preserve index
+    const upperGenes = new Map();
+    for (let i = 0, { length } = listGenes; i < length; i += 1) {
+      upperGenes.set(listGenes[i].toUpperCase(), i);
+    }
+
+    return upperGenes;
+  };
```
seve, line 40: Why define this per class and not in a larger scope?

client/src/components/geneExpression/geneSet.js
seve: Same as `gene.js`

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+              userSelect: "none",
+            }}
+            onClick={this.onGenesetMenuClick}
+          >
+            <Truncate>
+              <span
+                style={{
+                  maxWidth:
+                    globals.leftSidebarWidth -
+                    120 /* todo_genesets this magic number determines how much of a long geneset name we see, and will be tweaked as we build */,
```
seve, line 120: ''

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+                style={{ fontStyle: "italic" }}
+                checked={toggleSummaryHisto}
+                label="Show mean expression across entire set"
+                alignIndicator="left"
+                innerLabel="off"
+                innerLabelChecked="on"
+                onChange={this.toggleSummaryHisto}
+              />
+            </Tooltip>
+          ) : null}
```
seve, line 166:
```suggestion
          {isOpen && (
            <Tooltip
              content="Aggregate all genes in this geneset, and allow coloring by and selecting on the entire set."
              position={Position.BOTTOM_RIGHT}
              usePortal
              hoverOpenDelay={globals.tooltipHoverOpenDelay}
              modifiers={{
                preventOverflow: { enabled: false },
                hide: { enabled: false },
              }}
            >
              <Switch
                style={{ fontStyle: "italic" }}
                checked={toggleSummaryHisto}
                label="Show mean expression across entire set"
                alignIndicator="left"
                innerLabel="off"
                innerLabelChecked="on"
                onChange={this.toggleSummaryHisto}
              />
            </Tooltip>
          )}
```

client/src/components/geneExpression/geneSet.js
```diff
@@ -0,0 +1,187 @@
...
+              return (
+                <Gene
+                  key={gene}
+                  gene={gene}
+                  geneset={setName}
+                  isDiffexp={isDiffexp}
+                />
+              );
+            })
+          : null}
```
seve, line 180:
```suggestion
        {isOpen && !toggleSummaryHisto &&
           _.map(setGenes, (gene) => {
              return (
                <Gene
                  key={gene}
                  gene={gene}
                  geneset={setName}
                  isDiffexp={isDiffexp}
                />
              );
            })}
```
(probably needs linting)

client/src/components/geneExpression/index.js
```diff
@@ -1,59 +1,80 @@
...
+    }
+
+    return differential.diffExp ? (
+      <GeneSet
+        key="Temp DiffExp Set"
+        setGenes={setGenes}
+        isDiffexp
+        setName="Temp DiffExp Set"
+      />
+    ) : null;
```
seve, line 50: We're essentially doing the same if statement twice in succession. Maybe exit early instead

client/src/components/rightSidebar/index.js
```diff
@@ -1,7 +1,6 @@
 // jshint esversion: 6
```
seve, line 1:
```suggestion
```

client/src/components/categorical/annoDialog.js
```diff
@@ -18,6 +18,8 @@ class AnnoDialog extends React.PureComponent {
       validationError,
       annoSelect,
       annoInput,
+      secondaryInstructions,
```
bkmartinjr, line 21: This component is now used by `.../components/categorical/` and `.../components/geneExpression/`.  Seems like it nees to be moved up the tree into `.../components`, so geneExpression has no dependency on internal categorical sub-components.

seve (commented):
First half of the review

client/src/components/categorical/index.js
```diff
@@ -182,6 +182,35 @@ class Categories extends React.Component {
           }
         />
 
+        {writableCategoriesEnabled ? (
```
bkmartinjr, line 185: this change seems unrelated to gene sets, and just moves the Create button for categorical annotation to the top.  The layout for the tools is also now quite odd (see snapshot below).   So I'm wondering:
a) why change it at all in this PR?
b) or if we are going to change it, due to the new Create Gene Sets button, why not clean up the layout so it is more visually pleasing?

image

colinmegill: Great question. Yes, I want to leave create category at the top, since there are now two primary actions that might be buried by a large list of categories and/or genesets. The area above `create new gene set` is still in flight. This is captured in #2059

image

client/src/components/geneExpression/geneSet.js
colinmegill: @seve I disagree that this is more clear. @maniarathi, tiebreaker?

seve: I originally did it the same way, but Timmy pointed out this version to me. With a ternary its expected that there's a second half (if/else).  While the `&&` more clearly denotes that it's `if...then`

bkmartinjr: I agree with @seve.   Ternary is confusing in this case.

colinmegill: Ok, search shows `src/client/components` use this pattern upwards of 60 times. I'll fix here and create a ticket for tech debt.

signechambers1:
Hey Colin,

Thanks for the PR. Here are my notes from my review:

What doesn’t work:
* Renaming gene set name does not work
* When you click rename gene set, existing name should be prepopulated in text box (similar to renaming a category)
* Adding an duplicate gene to gene set (bruce caught too in slack)
* Adding a gene set that has the same name as an existing gene set (bruce caught too in slack)
* Saving seems a bit buggy - If I create a new gene set the bottom left hand corner of the main viz says “Unsaved” but if I open a new tab or refresh it says All Saved and I did not lose any work
    * At a different point in time I got Autosave error: HTTP error 400 - BAD REQUEST but I’m not sure what caused it and can’t reproduce

Polish - for later (I can make a zenhub ticket)
* histogram is too close to gene name
* Carrot for expanding gene set should have more space between the name of the gene set and the carrot (see left side bar)
* Need to do a copy review of everything, language can be tighter and more consistent with annotations (Edit gene set vs rename in modal, create new gene set modal)

Questions - for later (I can make a zen hub ticket)
* If someone creates an empty gene set or deletes all genes in a gene set, what should the behavior be? My idea would be to either disable expand or enable expand plus a massage “Add genes to gene set in order to view gene expression”

Other things I tested that worked:
* Expansion of gene set
* Expanding gene
* Brushing on histogram
* Plotting a gene on x and y
* Deleting a gene set works
* Adding a gene that is in the dataset
* Adding a gene that is not in the dataset
* Bulk adding genes to gene set
* Deleting a gene from a gene set
* Cellxgene regular functionality (color by, diffexp, annotations, subset, reembedding, panning, clipping, lassoing)

client/src/components/geneExpression/gene.js
colinmegill: I like this a lot as a convention

bkmartinjr:
@colinmegill - autosave is broken.  If I add a new gene set, wait a few seconds, I then see the following in the footer:

image

The error is a 404 "TID is stale".  This may be my bug, not yours - I have not yet investigated. Why don't you evaluate and let me know if you need help.

client/src/reducers/index.js
bkmartinjr: I don't reproduce, but I also am unclear on the sequence of steps you performed to see the error. Can you give me a step-by-step example?

In principle, leaving `genesetsUI` off the undoable state just means that it will be untouched by undo/redo.  So if the reducer _only_ captures state while the modal is active, AND the modal and undo/redo UI are exclusive, I don't think you need to list it in the undo config.   That said, I may be misunderstanding the reducer state. If you send me the test case, I can dig future.

Last thought - also include the error that you see, on the off chance that it has nothing to do with undo/redo.

client/src/components/categorical/index.js
```diff
@@ -182,6 +182,35 @@ class Categories extends React.Component {
...
+          <div style={{ marginBottom: 10 }}>
+            <Tooltip
+              content={
+                userInfo.is_authenticated
+                  ? "Create a new category"
+                  : "You must be logged in to create new categorical fields"
```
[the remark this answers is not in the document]
colinmegill: @bkmartinjr can check me, but I think all local development is non session and anno enabled

client/src/components/geneExpression/menus/editGenesetNameDialogue.js
colinmegill: Capturing in issue.

client/src/components/geneExpression/gene.js
colinmegill: Most of our keyboard commands do work. I'm going to keep this issue open to figure this out.

client/src/components/geneExpression/menus/createGenesetDialogue.js
colinmegill: @signechambers1 is creating an issue for validation

client/src/components/geneExpression/gene.js
colinmegill: I'll make an issue, we do this everywhere

client/src/components/geneExpression/gene.js
colinmegill: #2106

client/src/components/categorical/value/index.js
colinmegill: Renaming it. It's actually stacked bar height, and it relates to value because of Roboto Condensed at 11 pt's height. The engineering work to make that pixel perfect would be fragile.

image

client/src/components/geneExpression/geneSet.js
colinmegill: #2107

client/src/components/categorical/index.js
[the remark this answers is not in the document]
bkmartinjr: @colinmegill - I'm not sure you are answering @seve 's core question (ie, how would writable annotations be enabled in the absence of authenticated identity).

For the local server / desktop, `userInfo.is_authenticated` is always true, as (worst case) we will have a randomly generated guest ID that is used to save per-user info.  So in this environment, the answer is "no" - if `writableCategoriesEnabled` is true, `is_authenticated` will always be true.

I believe this code was added to support the original conception of the hosted service, where the "create annotation" button was always visible (if enabled), but would not do anything until the user logged in.

client/src/reducers/index.js
colinmegill: I saw a 404 on put, which has been resolved thanks to @bkmartinjr

signechambers1 (approved):
All my check's passed, thank you Colin! This is so exciting. My checks are documented [here](https://docs.google.com/spreadsheets/d/1H8Pr6ZRLiGA7cFNZJvM5LFgD3os4Ixl4RsXJE6O0aQ8/edit#gid=0) just as an FYI.

bkmartinjr:
@colinmegill - I don't think you have configured the undoable reducer correctly for gene sets.   For example:
* the gene sets UI state is not preserved in the history stack, which means it will get out of sync with the rest of the UI
* the gene set UI actions cause history to be pushed, which seems wrong for a modal.
This is based upon code inspection, not finding a specific UI error/bug (but I'm certain they are present :-).

Several suggestions:
1. add gene sets UI state to the history stack
2. add all of the gene set UI-related actions to the history skip list in udoableConfig.js

Use the annotations UI controller as a hint...

client/src/util/maybeScientific.js
```diff
@@ -0,0 +1,17 @@
+import significantDigits from "./significantDigits";
+
+export default function maybeScientific(x) {
```
bkmartinjr, line 3: Please add a comment that describes the purpose of the function, the parameters and the return value.  For example:
```
/**
  * The function will ...
  * @param x - a d3 object which ....
  * @returns - ...
  */
```

client/src/util/clamp.js
```diff
@@ -0,0 +1,3 @@
+export default function clamp(val, rng) {
```
bkmartinjr, line 1: Please add a comment that describes the purpose of the function, the parameters and the return value.  For example:
```
/**
  * The function will ...
  * @param x - a d3 object which ....
  * @returns - ...
  */
```

bkmartinjr:
@colinmegill - all new exported functions in `util` should have comments detailing their purpose, behavior, params and return value.

client/src/reducers/undoableConfig.js
```diff
@@ -30,6 +31,9 @@ const skipOnActions = new Set([
   "writable obs annotations - save complete",
   "writable obs annotations - save started",
   "writable obs annotations - save error",
+  "autosave: genesets started",
+  "autosave: genesets error",
+  "autosave: genesets complete",
```
bkmartinjr, line 37: It appears that the gene set UI-related actions are missing.  I believe they should *not* generate a new history stack element, so you should put them in the skip list.

client/src/reducers/genesetsUI.js
```diff
@@ -0,0 +1,55 @@
...
+*/
+const GeneSetsUI = (
+  state = {
+    createGenesetModeActive: false,
+    isEditingGenesetName: false,
+    isAddingGenesToGeneset: false,
+  },
+  action
+) => {
+  switch (action.type) {
```
bkmartinjr, line 12: please document the interface of each action, at least briefly (ie, if the action name is clear, at least add the parameters and their type).

bkmartinjr:
@colinmegill - it looks like this PR partially migrates us from blueprintjs/popover to .../popover2. That seems like a good idea, but we also need to get the rest of the code migrated.  Can you ensure there is an issue filed, to do the cleanup sometime soon (so we don't have both in the code base).

client/src/components/rightSidebar/index.js
```diff
@@ -22,10 +20,24 @@ class RightSidebar extends React.Component {
           overflowY: "inherit",
           height: "inherit",
           width: "inherit",
+          padding: globals.leftSidebarSectionPadding,
         }}
       >
+        {/* <div
```
bkmartinjr, line 26: Dead code.

client/src/components/util/truncate.js
```diff
@@ -86,18 +86,12 @@ export default (props) => {
     })
   );
   return (
-    <Tooltip
+    <Tooltip2
```
bkmartinjr, line 89: I noticed that the old user annotation code is using this module, and it uses the old Popover component.  Is popover compatible with Tooltip2?

Put another way, have we confirmed that this change does not break category user annotations?

client/src/components/brushableHistogram/error.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import * as globals from "../../globals";
+
+const ErrorLoading = ({ displayName, error, zebra }) => {
+  console.log(error); // log to console as this is unexpected
```
bkmartinjr, line 5: If this is intentionally trying to log an error to the console (as opposed to informational logging or debugging), then you should use `console.error()` rather than log().

bkmartinjr: Also, we do not perform this action for any other type of error.  We should have a consistent policy and implement it everywhere.  Not just here.

client/src/components/categorical/value/index.js
```diff
@@ -739,7 +739,7 @@ class CategoryValue extends React.Component {
                       ) : null}
                       {displayString !== globals.unassignedCategoryLabel ? (
                         <MenuItem
-                          icon="delete"
+                          icon="trash"
                           intent="danger"
```
bkmartinjr, line 743: should this be `Intent.DANGER`?  (ditto for "trash")

client/src/components/categorical/value/index.js
bkmartinjr: How about adding a new global then?  Maybe called `STACKED_BAR_HEIGHT`?

client/src/components/brushableHistogram/index.js
```diff
@@ -4,439 +4,20 @@ https://bl.ocks.org/mbostock/34f08d5e11952a80609169b7917d4172
 https://bl.ocks.org/SpaceActuary/2f004899ea1b2bd78d6f1dbb2febf771
```
bkmartinjr, line 4: What is the significance of these comments?  They seem like notes from the component development phase?

client/src/components/geneExpression/index.js
```diff
@@ -1,59 +1,78 @@
...
-                    pvalAdj={value[3]}
-                  />
-                );
-              })
-            : null}
+          <div style={{ marginBottom: 10, position: "relative", top: -2 }}>
+            <Button
+              data-testid="open-create-geneset-dialog"
+              onClick={this.handleActivateCreateGenesetMode}
+              intent="primary"
```
bkmartinjr, line 66: Intent.PRIMARY?

bkmartinjr:
There are still some bugs in validation of gene set names.  Example:  if I "edit" a gene set name, enter the _same_ (identical) name, and hit rename button, I get an error:

```
genesets.js:169 Uncaught Error: geneset: update -- update specified existing name.
    at GeneSets (genesets.js:169)
    at eval (cascade.js:33)
    at skip (undoable.js:152)
    at eval (undoable.js:279)
    at dispatch (redux.js:221)
    at eval (middleware.js:19)
    at eval (index.js:12)
    at RenameGeneset.renameGeneset (editGenesetNameDialogue.js:48)
    at HTMLUnknownElement.callCallback (react-dom.development.js:70)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:90)
```

client/src/components/geneExpression/menus/addGeneToGenesetDialogue.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+  };
+
+  handleAddGeneToGeneSet = (e) => {
+    const { geneset, dispatch } = this.props;
+    const { genesToAdd } = this.state;
+
+    const genesTmpHardcodedFormat = [];
+
+    const genesArrayFromString = _.pull(_.uniq(genesToAdd.split(/[ ,]+/)), "");
```
bkmartinjr, line 31: this little bit of magic is repeated in `createGenesetDialogue`.  Suggest pulling it out into a helper utility function given its generic behavior (ie, split string by comma or space, return unique non-empty strings)

signechambers1:
@colinmegill tested undo redo and it is not working as expected as @bkmartinjr mentioned, here are the tests I ran and behavior I witnessed:

Undo / Redo for create new gene set
* FAIL, undoing might revert two actions? the first time i tested it removed my gene set plus my color by of a categorical variable; the second time it removed my gene set plus added back in a prior gene set I created. Redo seems to work as expected
Undo / Redo for add new gene
* FAIL, undoing removes the entire gene set, not just the individual gene i added. Redo shows the gene but the gene set is not expanded.
Undo / Redo for rename a gene set
* FAIL, undoing removes the entire gene set, not the rename. Redo does not work as expected
Undo / Redo for delete gene 
* FAIL, undo does nothing. Redo does nothing
Undo / Redo for delete gene set
* FAIL, undo deletes the gene set but redo does not restore the gene set

client/src/components/util/truncate.js
colinmegill: image

client/src/components/categorical/value/index.js
colinmegill: https://github.com/chanzuckerberg/cellxgene/issues/2120

bkmartinjr (approved):

maniarathi (approved):

PR #2099, opened 2021-03-10
gene sets summary route

bkmartinjr:
Back end route implementing gene set summary calculation.  Also includes:
* refactored the error checking logic for gene sets so that anything returned by read_genesets() will be valid, and anything provided to write_genesets() will be validated prior to write (this removes the need for users of the IO tier to separately coordinate error checking)
* test cases for the new route
* slightly improved test cases for the GET /genesets route

Fixes #2099 

#### Reviewers
**Functional:** @MDunitz 

**Readability:** @maniarathi

### DIFF
--- a/local_server/common/rest.py
+++ b/local_server/common/rest.py
@@ -19,6 +19,7 @@ from local_server.common.errors import (
     ColorFormatException,
     AnnotationsError,
     ObsoleteRequest,
+    UnsupportedSummaryMethod,
 )
 
 import json
@@ -336,7 +337,7 @@ def genesets_get(request, data_adaptor):
 
     try:
         annotations = data_adaptor.dataset_config.user_annotations
-        (genesets, tid) = data_adaptor.check_new_gene_sets(annotations.read_gene_sets(data_adaptor))
+        (genesets, tid) = annotations.read_gene_sets(data_adaptor)
 
         if preferred_mimetype == "text/csv":
             return make_response(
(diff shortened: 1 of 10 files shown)


local_server/common/rest.py
```diff
@@ -373,10 +374,39 @@ def genesets_put(request, data_adaptor):
...
+
+    geneset_name = request.args.get("geneset_name", default=None)
+    summary_method = request.args.get("method", default="mean")
+    request_tid = request.args.get("tid", default=None)
+
+    try:
+        annotations = data_adaptor.dataset_config.user_annotations
+        (genesets, tid) = annotations.read_gene_sets(data_adaptor)
+
+        if request_tid is not None and int(request_tid) != tid:
```
maniarathi, line 398: Quick question about the case where the tid is None -- is this actually a valid scenario where a gene set summary can be fetched? The initial gene sets need to be read before a summary can be fetched (I think?) so the TID is guaranteed to be 0 based on a line in the critciial section of your `read_gene_sets` function where `tid = self.last_geneset_tid`.

local_server/test/unit/common/test_api.py
```diff
@@ -677,8 +691,86 @@ def test_case(test, expected_code, original_data):
...
+            endpoint = f"geneset_summary?geneset_name={gs}"
+            result = self.session.get(f"{self.URL_BASE}{endpoint}", headers={"Accept": "application/octet-stream"})
+            self.assertEqual(result.status_code, HTTPStatus.OK)
+            self.assertEqual(result.headers["Content-Type"], "application/octet-stream")
+            df = decode_fbs.decode_matrix_FBS(result.content)
+            self.assertEqual(df["n_rows"], 2638)
+            self.assertEqual(df["n_cols"], 1)
+            self.assertEqual(df["col_idx"], [gs])
+
+    def test_get_geneset_error_handling(self):
```
maniarathi, line 762: I think missing a test case for `UnsupportedSummaryMethod`

bkmartinjr: It is the last test in the list, where it asks for `method=NO_SUCH_METHOD`

local_server/data_common/data_adaptor.py
```diff
@@ -283,17 +284,16 @@ def check_new_gene_sets(self, args, context=None):
...
         """
-        (genesets, tid) = args
         messagefn = context["messagefn"] if context else (lambda x: None)
 
         # accept genesets args as either the internal (dict) or REST (list) format,
         # as they are identical except for the dict being keyed by geneset_name.
         if type(genesets) not in (dict, list):
             raise ValueError("Genesets must be either dict or list.")
-        genesets = genesets if type(genesets) == list else genesets.values()
+        genesets_iterable = genesets if type(genesets) == list else genesets.values()
```
maniarathi, line 293: This is very much a nit but I believe the more Pythonic way to check for type is `isinstance(genesets, list)`. Noticed it in a few places, but up to you whether to switch or not. Just a nit.

maniarathi (approved):
Just a few comments, nothing major. LGTM.

local_server/common/rest.py
maniarathi: Ahh I see. I did not grok that if the `request_tid` is None that the client does not want synchronization. Thanks for the clarification.

local_server/test/unit/common/test_api.py
maniarathi: 🤦 I see it now. Thanks. My squirrel brain just skimmed right over it.

local_server/data_common/data_adaptor.py
```diff
@@ -262,15 +262,16 @@ def check_new_labels(self, labels_df):
...
         May also modify the gene set for conditions that should be resolved,
         but which do not warrant a hard error.
 
-        Argument 'args' must be a tuple containing (genesets, tid).  Genesets
-        may be either the REST OTA format (list of dicts) or the internal format
-        (dict of dicts, keyed by the geneset name).
+        Argument genesets may be either the REST OTA format (list of dicts) or the internal
+        format (dict of dicts, keyed by the geneset name).
+
+        Will return a modified genesets (eg, remove dups).
```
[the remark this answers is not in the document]
bkmartinjr: No, returns it as the original type.  dict->dict, list->list.

Will clarify in the comment

local_server/data_common/data_adaptor.py
bkmartinjr: will fix.  I assume I should do this for all of the other type(x) is Y tests? (there are many :-)

maniarathi: I'm happy just not _adding_ additional `type(x) == type` lines in this PR :) Don't have to switch everything right now.

bkmartinjr: oops - I just did a bunch. Will commit them and stop!

maniarathi (approved):
LGTM thanks for making those changes.

Issue #2096, opened 2021-03-08
User can create gene sets with ASCII characters 32-126


signechambers1:
User can use ASCII characters 32-126. Allowing all these characters will allow for more flexibility for users and still allow for parsing. This will be consistent with annotations once #2095 is complete.

bkmartinjr:
requires both f/e and b/e changes.

signechambers1:
Closing in favor of https://app.zenhub.com/workspaces/cellxgene-explorer-600f112266e8830017f47e12/issues/chanzuckerberg/cellxgene/2160

PR #2117, opened 2021-03-22
update gene set name validation to match latest requirements

bkmartinjr:
gene set name validation has been updated to be more permissive.  This updates the back-end checks to match.

TL;DR on validation rules for gene set names:
* no leading or trailing white space
* no newline, return, tab or vertical tab
* no multi-space runs

This is a partial fix to #2096.


maniarathi (approved):
LGTM

MDunitz (approved):
lgtm

signechambers1 (approved):
LGTM! loaded a gene sets file with gene set names with special characters []$^ that were previously not allowed and it loaded without error. Thanks Bruce!

PR #2118, opened 2021-03-22
revert removal of cache control headers

### LINKED ISSUE #2077: add cache control headers back to local_server
They were accidentally removed during the back-end split.  They are needed for various self-hosting use cases.

bkmartinjr:
For local server, revert removal of cache control headers and the config which enables/disables them.

Fixes #2077 

Code is identical to that which was previously removed.


MDunitz (approved):
lgtm

PR #2135, opened 2021-03-31
propagate diffexp state through component tree

### LINKED ISSUE #2133: After running DiffExp, log fold change and p-value should show up in histogram when you expand a gene in a gene set
When you run Diffexp, the current working version of gene sets (on main) produces a gene set with 50 genes. when you expand a gene, a histogram shows up. The log fold change and pvalue should also show up underneath the histogram; these values are currently missing. Please add them back.

bkmartinjr:
The differential expression state was not correctly being propagated through the component tree, which manifest as a missing "p-val" and "lfc" on histograms related to diffexp. This PR makes the state available to the components that actually need it.

Fixes #2133

### DIFF
--- a/client/src/components/brushableHistogram/index.js
+++ b/client/src/components/brushableHistogram/index.js
@@ -452,6 +452,7 @@ class HistogramBrush extends React.PureComponent {
                 {!mini ? (
                   <HistogramFooter
                     isObs={isObs}
+                    isDiffExp={isDiffExp}
                     displayName={field}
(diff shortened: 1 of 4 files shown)


signechambers1 (approved):
Looks good to me, loaded PBMC3K and ran diffexp a couple times and checked that both p-value and log fold change appear for each gene.

Thanks Bruce!

colinmegill (approved):

Issue #2040, opened 2021-02-03
Add front end support for gene set description


signechambers1:
Being able to see the description for a gene set

Issue #2120, opened 2021-03-22
Use blueprint IconNames everywhere


colinmegill:
Tech debt issue `icon={IconNames.TRASH}` rather than `icon={<Icon icon="more" iconSize={10} />}`

PR #2139, opened 2021-04-02
Create / edit geneset description

### LINKED ISSUE #2050: User can add and edit a gene set description
User can add a description when they create a gene set and edit that description.

colinmegill:
Fixes #2050 and #2040


signechambers1:
Hi Colin,

A few findings from my review:

* Undo / Redo does not work with editing a description
    * Undo edit description brings up dialog box, inconsistent with annotations which just changes the name back
    * If you edit a description, hit Undo / Redo, and then try to edit again the text boxes with name and description are no longer populated
* Creating a new category causes an autosave error (create new category from left side bar, enter a name, click create new category)
* When there is no description, I think the hover over should only state the name of the gene set instead of gene set name: <no description>. I think this is more consistent with the rest of the hover overs (for example in annotations)
* Rename gene set modal - wording is inconsistent since you can edit the name and description. Feel free to edit but here are a couple suggestions:
    * “Edit gene set name and description” for the title and blue button
    * “Rename <gene set name>” for the title of the name box
    * “Edit description” for the title of the description box

Let me know if you would like to discuss!
Signe

bkmartinjr:
@signechambers1 - I suspect that this:

> Creating a new category causes an autosave error (create new category from left side bar, enter a name, click create new category)

Is due to inconsistent validation checks on the front-end and back-end.  If you recall, we had deferred reconciling back-end validation until the rules were finalized.  I can explore resolving this if they are now finalized (please point me at the final rules).

signechambers1 (approved):
LGTM.

Just a note for my future self, if you create a gene set with description and genes, hitting undo removes the genes but the gene set still exists. If you hit undo again, it removes the gene set. Under the hood, this is two actions (creating a gene set and adding a gene set). This is an acceptable user experience for launch.

client/src/components/geneExpression/menus/createGenesetDialogue.js
```diff
@@ -97,47 +108,94 @@ class CreateGenesetDialogue extends React.PureComponent {
...
+            }}
+          >
+            <div className={Classes.DIALOG_BODY}>
+              <div style={{ marginBottom: 20 }}>
+                <p>{this.instruction(genesetName, genesets)}</p>
+                <LabelInput
+                  onChange={this.handleChange}
+                  inputProps={{
+                    "data-testid": "create-geneset-modal",
+                    leftIcon: "manually-entered-data",
```
seve, line 129: IconName constants when you can

client/src/components/geneExpression/menus/createGenesetDialogue.js
```diff
@@ -97,47 +108,94 @@ class CreateGenesetDialogue extends React.PureComponent {
...
+                    intent: "none",
```
seve, line 130: BP3 also exposes Intent constants we can use

seve (approved):
small nits related to #2120

PR #2155, opened 2021-04-16
hosted gene sets routes, plus a few bug fixes

### LINKED ISSUE #2152: Replace "genesets" with "gene set" for all CLI error messages
Change gene set to be two words instead of one:

Expected behavior: Error message should read "Gene set CSV missing required gene set or gene name on line 8"
Actual behavior: Error message reads "Geneset CSV missing required geneset or gene name on line 8" 

Expected behavior: Error: 'user-generated-data-dir' may not be used with annotations-file' or 'gene-sets-file'.
Actual behavior: Error: 'user-generated-data-dir' may not be used with annotations-file' or 'genesets-file'.

These are two examples, Bruce found a bunch, this ticket should cover all changes.

### LINKED ISSUE #2154: CSV parser should not strip leading spaces
[Test Case # 5](https://docs.google.com/spreadsheets/d/1H8Pr6ZRLiGA7cFNZJvM5LFgD3os4Ixl4RsXJE6O0aQ8/edit#gid=551030192) Leading spaces in gene_set_name

When loading a file with leading spaces in the gene_set_name:

Expected behavior: Error, does not load 
Actual behavior: App loads and strips leading spaces

As discussed, we want to keep the description field raw and consistent behavior for gene_set_name with leading and trailing spaces, so the CLI should error when a user uploads a file with leading spaces in any gene_set_name.

bkmartinjr:
Changes in this PR:
* front-end:  refined logic which triggers the annotation collection dialog on the front-end.  It was asking the user for a collection name in situations where it was not needed (eg, gene sets file specified and --disable-annotations).
* czi_hosted backend: routes for gene sets support
* all backends: change "geneset" to "gene set" in user visible messages/errors
* all backends: gene set tidy csv parser no longer strips leading spaces
* all backends: gene set name validation modified to match requirements 

Fixes #2152 
Fixes #2154


signechambers1:
@bkmartinjr there are two typos in the following error message (missing quote and genesets should be gene-sets):

Actual error message: Error: 'user-generated-data-dir' may not be used with annotations-file' or 'genesets-file'.
Expected error message" Error: 'user-generated-data-dir' may not be used with 'annotations-file' or 'gene-sets-file'.

Otherwise looks good. I only tested the bug fixes, if theres a way for me to test the hosted back end routes let me know.
Thanks!

bkmartinjr:
> typos in the following error message

@signechambers1 - these are fixed. Please review when you have a chance.

signechambers1 (approved):
LGTM, confirmed the typos are fixed in the error message. Thanks Bruce!

backend/test/test_czi_hosted/unit/common/test_api.py
```diff
@@ -315,6 +316,95 @@ def test_static(self):
...
+
+    def test_post_summaryvar(self):
+        index_col_name = self.schema["schema"]["annotations"]["var"]["index"]
+        endpoint = "summarize/var"
+        headers = {"Content-Type": "application/x-www-form-urlencoded", "Accept": "application/octet-stream"}
+
+        # single column
+        filter = f"var:{index_col_name}=F5"
+        query = f"method=mean&{filter}"
+        query_hash = hashlib.sha1(query.encode()).hexdigest()
```
MDunitz, line 382: nit -- maybe use `summarizeQueryHash` here to avoid issues if it is changed in the future

backend/server/common/annotations/local_file_csv.py
```diff
@@ -259,6 +260,7 @@ def update_parameters(self, parameters, data_adaptor):
         params = {}
         params["annotations"] = self.user_annotations_enabled()
         params["annotations_genesets_readonly"] = not self.gene_sets_save_enabled()
+        params["annotations_genesets_name_is_read_only"] = self.gene_sets_output_file is not None
```
MDunitz, line 263: I might be misunderstanding how this is being used, but I think it should be         `params["annotations_genesets_name_is_read_only"] = self.gene_sets_output_file is None`
my interpretation is if there is no output file set it should be read only (param set to True) but if there is a value for the output file the param `annotations_genesets_name_is_read_only` should be false.

backend/czi_hosted/common/config/dataset_config.py
```diff
@@ -54,7 +55,7 @@ def __init__(self, tag, app_config, default_config):
             raise ConfigurationError(f"Unexpected config: {str(e)}")
 
         # The annotation object is created during complete_config and stored here.
-        self.user_annotations = None
+        self.user_annotations = Annotations()
```
MDunitz, line 58: Why is this being initialized here and in complete_config? Should the comment above it be updated?

backend/test/test_czi_hosted/unit/common/test_api.py
bkmartinjr: The reason I did this is the specific hash is part of the protocol (the front-end also needs to be able to generate the hash from the query).  So the protocol mandates a given hash - in this case SHA1, with a hex digest string.  I figured we should have our tests confirm that this hasn't changed. If it does change, then the front-end and back-end both must modify their code (so it didn't seem like a stretch to have the test code change too).

Does that seem reasonable?  if not, we need to add some other validation that the backend is really using SHA1/hex.

backend/czi_hosted/common/config/dataset_config.py
bkmartinjr: The "old" Annotations class was an ABC, as it had a hard requirement on configuration.  The new class is _not_ abstract, as it can provide read of gene sets without config (it can derive their location from the dataset location).   So the default setup, barring additional config, is to use the default behavior.

The comment is a bit misleading, as the `complete_config` may create a different Annotation instance, and replace the default one.  This happens, for example, if the user has enabled the "local csv" annotation class.

I'll update the comment.

backend/server/common/annotations/local_file_csv.py
bkmartinjr: This param indicates whether the annotation collection name is read-only (fixed by the backend) or read-write (the f/e will prompt the user for a name).  The current and only use for the param is to tell the f/e whether or not it should ask the user for a name.

When the backend already knows the file name, the name is read-only.  When the backend does not know a name, it sets this to True and the f/e prompts.  See:
* client/src/reducers/annotations.js: line 37
* client/src/components/autosave/filenameDialog.js: line 110

This is probably a case of a badly named parameter, predating this PR.

backend/test/test_czi_hosted/unit/common/test_api.py
MDunitz: That makes sense to me

backend/server/common/annotations/local_file_csv.py
MDunitz: ahh I figured I was missing something. Thanks!

colinmegill (approved):

MDunitz:
Bug description: 
- Temp DiffExp gene set list  disappears from the right side bar
- See message `Failure loading UMAP`

To reproduce:
- do differential expression on two populations
- get the mean expression
- color by the mean expression
- brush on a random subset of the histogram
- click the subset button

bkmartinjr:
> Bug description:
> 
> * Temp DiffExp gene set list  disappears from the right side bar
> * See message `Failure loading UMAP`
> 

I was able to reproduce, but this is a bug in some new code Colin added (unrelated to this PR).  Filed as #2176

MDunitz (approved):
lgtm

PR #2181, opened 2021-05-03
Geneset tooltip description copy

### LINKED ISSUE #2144: Remove colon from gene hover over when there is no gene description
Remove colon from gene hover over when there is no gene description

colinmegill:
Fixes #2144


signechambers1 (approved):
LGTM!

PR #2216, opened 2021-05-20
feat(geneset): save diff exp result as geneset

seve:
#### Reviewers

**Functional:**  @mkmartinjr

**Readability:** @colinmegill 

######  Note: It's probably best to review this PR one commit at a time
---

## Changes
- add
  - gene set reducer/rendering now handles diffexp result store and display 93daf9a 
- remove
  - Removes old diffexp handling and data structures c86c4ba
- modify
  - Update package-lock to v2 ba8ed9e
  - found a small bug that broke diff exp on falsey values 17be78a
  
---

## Note

I utilized the gene description field to store the diffexp data to display during rendering.

Example column:
```csv
 "DiffExp Set 5/20/2021, 11:28:36 AM",,AC145205.1-1,"0,1,1" 
 ```
 
 Where `0,1,1` => `logfoldchange, pval, pval_adj`

---

Related Issues:
 * #729
 * #2145 
 * #2176

### CODE REVIEW

File: client/src/components/brushableHistogram/footer.js [hunk 1]
```diff
@@ -61,7 +61,7 @@ const HistogramFooter = React.memo(
           </span>
         </div>
 
-        {logFoldChange && pvalAdj ? (
+        {logFoldChange !== undefined && pvalAdj !== undefined ? (
```
colinmegill, line 64: This appears equivalent, but wouldn't catch `null` — is there a specific motivation?

client/src/components/geneExpression/index.js
```diff
@@ -17,11 +17,14 @@ class GeneExpression extends React.Component {
     const { genesets } = this.props;
 
     for (const [name, geneset] of genesets) {
+      const isDiffExp = geneset.genesetName.includes("DiffExp Set");
```
colinmegill, line 20: Is it the case here that `DiffExp Set` is a magic word, and does the for loop have to know about it? Could it be extracted to globals as a constant?

client/src/components/brushableHistogram/footer.js
colinmegill: ✅ good catch

client/src/components/geneExpression/index.js
seve: Yes to magic word, the loops uses this to know if the geneset is a diffexp geneset. I'll extract it to a global const

seve: On second thought, this might not be the best idea.  If a user renames this geneset they'll lose diffexp functionality which would be unexpected functionality.

I'll most likely shift over to a geneset description instead, any suggestions for what the insert into the geneset description? 
cc: @bkmartinjr , @colinmegill

client/src/components/brushableHistogram/footer.js
seve:
```suggestion
        {logFoldChange != null && pvalAdj != null ? (
```

client/src/components/geneExpression/index.js
seve: Spoke with colin offline, we'll be moving forward with description

client/__tests__/util/stateManager/controlsHelpers.test.js
```diff
@@ -1,38 +0,0 @@
-/*
```
bkmartinjr, line 1: Why is this file being deleted (or put another way, where are the tests for exported functions from stateManager/controlsHelpers)?

seve: The only function being tested in this file was `subsetAndResetGeneLists()` which was deleted in this PR.  Should I just empty the file contents?

client/src/components/categorical/category/index.js
```diff
@@ -144,15 +144,14 @@ class Category extends React.PureComponent {
     */
     const { schema } = annoMatrix;
     const { colorAccessor, colorMode } = colors;
-    const { genesets, differential } = this.props;
```
bkmartinjr, line 147: if you delete the use of `differential`, please also delete the setting of the property (line 48, above)

client/src/components/continuousLegend/index.js
```diff
@@ -111,7 +111,7 @@ const continuous = (selectorId, colorScale, colorAccessor) => {
 }))
 class ContinuousLegend extends React.Component {
   async componentDidUpdate(prevProps) {
-    const { annoMatrix, colors, genesets, differential } = this.props;
+    const { annoMatrix, colors, genesets } = this.props;
```
bkmartinjr, line 114: need to delete line 110, where the (now removed) differential prop is defined.

colinmegill (approved):

client/src/components/geneExpression/geneSet.js
```diff
@@ -84,14 +84,17 @@ class GeneSet extends React.Component {
       setGenes,
       setGenesWithDescriptions,
       isDiffExp,
```
bkmartinjr, line 86: The existence of this flag breaks the model -- ie, "diff exp" results are not "just a gene set", but are now special in some way.  And overloading the semantics of the description field to be _both_ a description (user editable) and a flag will lead to fragility and likely future bugs (ie, user edits the field, and it no longer acts like a diffexp result).

I think the only solution is to pick one of:
* diff exp just generates a _generic_ gene set.  Get rid of the `isDiffExp` flag, including the presentation of p-val and lfc.
* Or, diffexp are "magically" special, and we need to add the flag and p-val/lfc to the gene set model, persisting them.

the latter seems untenable.   The former seems correct, and aligned with a related discussion @ambrosejcarr has been having around the (non)value of lfc/pval.

@signechambers1 - you also should comment.  The model as implemented is not consistent and will break if user edits the description.

client/src/components/geneExpression/index.js
```diff
@@ -17,11 +18,15 @@ class GeneExpression extends React.Component {
     const { genesets } = this.props;
 
     for (const [name, geneset] of genesets) {
+      const isDiffExp =
```
bkmartinjr, line 21: super fragile as noted in `geneSet.js`

client/src/components/graph/graph.js
```diff
@@ -785,17 +785,11 @@ class Graph extends React.Component {
   }
 
   createColorByQuery(colors) {
-    const { annoMatrix, genesets, differential } = this.props;
+    const { annoMatrix, genesets } = this.props;
```
bkmartinjr, line 788: Need to delete the (now unused) `differential` property from line 80

client/src/components/graph/overlays/centroidLabels.js
```diff
@@ -76,16 +76,10 @@ class CentroidLabels extends PureComponent {
   };
 
   colorByQuery() {
-    const { annoMatrix, colors, genesets, differential } = this.props;
+    const { annoMatrix, colors, genesets } = this.props;
```
bkmartinjr, line 79: Need to delete the (now unused) `differential` property from line 18

client/src/components/geneExpression/geneSet.js
colinmegill: I support moving to `just a generic geneset` if @ambrosejcarr supports nonvalue of pval

bkmartinjr (changes requested):
See comment on `geneExpression/geneSet.js` -- the UX model is attempting to treat the gene set created by differential expression as "special", without properly supporting this new feature through the entire gene set data model.

As implemented, it is going to break in many cases:
* user saves the diffexp-created gene set (shows pval), restarts cellxgene, and reloads - the pval no longer displays
* user edits the description, and it changes behavior
* etc

I think we should simple make diffexp results a _generic_ gene set.  Ie, remove pval/lfc and any other special behavior.

Alternatively, we need to add this special behavior to the end-to-end gene set model, including persisted formats, reducers, etc.

@signechambers1 --^

client/__tests__/util/stateManager/controlsHelpers.test.js
bkmartinjr: Best case would be to improve code coverage by adding a few simple tests.   But if not viable, leaving an empty file and a "TODO" would be reasonable.  Totally understand the gap already existed (and was probably my fault!)

client/src/components/geneExpression/geneSet.js
signechambers1: Talked with @seve and @MaximilianLombardo. To move forward, we are going to remove the pvalues and log fold change from the diff exp output to allow for diffexp to be a generic gene set in this PR. 

Ambrose and Max have both said these values may have limited value to the user so we may be able to remove these permanently. However, to guard against the situation where we remove these values and then our users are upset about this, @MaximilianLombardo is going to check in with a few people on active threads and get more community feedback before we decide what to do in the medium / longer term.

Issue #2180, opened 2021-05-03
Diffexp returns two gene sets in the front end


signechambers1:
### Need

We want to be able to show two gene sets in the front end, one with top differentially expressed genes up-regulated in group A and one with top differentially expressed genes up-regulated in group B. We can determine these groups by the sign on log fold change, and then sorting by test statistic. 

### Approach

Dependent on #2179 completing, or agreement on API interface. Changes will be required in the reducer to accommodate the new data structure.

bkmartinjr:
we currently sort by the test statistic, not the p-val.  Can you confirm this change?  and do you want the p-val or the adjusted p-val?

Issue #2179, opened 2021-05-03
Update backend routes to support differential expression returning two lists of differentially expressed genes


signechambers1:
### Need

We want to be able to show two gene sets in the front end, one with top differentially expressed genes upregulated in group A and one with top differentially expressed genes upregulated in group B. We can determine these groups by the sign on log fold change, and then sorting by p-value. In other words, return both the 25 genes with the smallest p-values and positive log fold change, and 25 genes with the smallest p-values and negative log fold change.

### Approach

Today, we:
1. compute the t-statistic and other metrics
2. sort genes by the absolute value of that t-stat
3. grab the "top N"
4. put them in a poorly designed JSON response. It is just a flat array, which makes it super hard to debug

Instead, what we should do:
1. same, no change
2. sort genes by the actual value of that t-stat
3. grab the "top N" and "bottom N"
4. return both top N and bottom N in an improved JSON object, which is just easier to understand (ie, isn't a flat array)

Add a temporary mode switch so that the frontend can either choose to receive one or two.
Please also enable this for both hosted and desktop server implementations.

~There will also be a small amount of frontend work required to shape the response into the reducer into a format that Colin's current code understands (i.e. fakes the creation of the merged old-style data structure too).~ This work will be completed in #2180.

maniarathi:
Instructions to implement the backend:

Today, we:
1. compute the t-statistic and other metrics
2. sort genes by the absolute value of that t-stat
3. grab the "top N"
4. put them in a poorly designed JSON response. It is just a flat array, which makes it super hard to debug

Instead, what we should do:
1. same, no change
2. sort genes by the **actual** value of that t-stat
3. grab the "top N" and "bottom N"
4. return both top N and bottom N  in an improved JSON object, which is just easier to understand (ie, isn't a flat array)

There will also be a small amount of frontend work required to shape the response into the reducer into a format that Colin's current code understands (i.e. fakes the creation of the merged old-style data structure too).

maniarathi:
Hey @MDunitz -- I am assigning you this ticket. We'd like for this to complete before June 1st so if we end up finish up the sprint early, would you be able to take this on please? If not, I will help make time for it in the next sprint.

maniarathi:
Add a temporary mode switch so that the frontend can either choose to receive one or two.

PR #2230, opened 2021-05-25
feat: diffexp returns two genesets

seve:
#### Reviewers
**Functional:** @Bento007 + @colinmegill 

**Readability:** @MDunitz 

---

## Changes
- add
   -  Diffexp endpoint can return dual gene lists based on `two_lists` param
   - frontend requests diffexp with `two_lists` enabled
   - frontend persists the lists as two separate gene sets 

## Related Issues
 * #2180 
 * #2179


client/src/actions/index.js
```diff
@@ -197,6 +197,7 @@ const requestDifferentialExpression = (set1, set2, num_genes = 50) => async (
         }),
         body: JSON.stringify({
           mode: "topN",
+          two_lists: true,
```
colinmegill, line 200: Is this used?

client/src/actions/index.js
```diff
@@ -211,10 +212,13 @@ const requestDifferentialExpression = (set1, set2, num_genes = 50) => async (
 
     const response = await res.json();
     const varIndex = await annoMatrix.fetch("var", varIndexName);
-    const data = response.map((v) => [
-      varIndex.at(v[0], varIndexName),
-      ...v.slice(1),
-    ]);
+    const data = { negative: [], positive: [] };
```
colinmegill, line 215: Is there a more descriptive variable name that makes sense here, instead of `data`

colinmegill (changes requested):

client/src/actions/index.js
seve: `data` is cxg's de facto name for action payloads

client/src/actions/index.js
seve: Backend accepts it to toggle on the two list formatted response

client/src/actions/index.js
colinmegill: Ah! Yes, referring to the private variable before assignment, so, perhaps:

```
return dispatch ({
  type: ...,
  data: _descriptivePrivateVariableName
})
```

client/src/actions/index.js
colinmegill: This would seem to encode `diffExp` as a magic word, but a bit more cryptically, is there a reason we aren't using something more explicit like `isDifferentialExpression`?

seve: I'm lost, could quote the lines you're referencing?

colinmegill: To rephrase, line 200 uses the variable name `two_lists`, but does this variable name really mean `isDifferentialExpression`?

seve: No it means, we want the backend to return the diffexp results as an object with two lists, rather than a single list

colinmegill: Are there any other circumstances we want the system to do this other than diffexp?

seve: No, this is the only endpoint that accepts this argument

backend/test/test_server/unit/compute/test_diffexp_cxg.py
```diff
@@ -2,6 +2,8 @@
 
 import numpy as np
 
+
```
bkmartinjr, line 5: the comments, filename and code in this file are misleading.  They refer to "CXG", but that data format is no longer supported by the `server` backend.   The code loads an H5AD, but claims it is testing the CXG adaptor.  Please rename the file, correct the comments, etc, to refer to H5AD.

backend/czi_hosted/compute/diffexp_cxg.py
```diff
@@ -35,7 +35,7 @@ def get_thread_executor():
     return diffexp_thread_executor
 
 
-def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01):
+def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01, two_lists=False):
```
bkmartinjr, line 38: why retain the old "single list" mode?  Is it in use?  If not, it seems that code and test simplicity should be preferable to adding another modality....

backend/server/compute/diffexp_generic.py
```diff
@@ -65,16 +67,61 @@ def diffexp_ttest_from_mean_var(meanA, varA, nA, meanB, varB, nB, top_n, diffexp
...
     logfoldchanges = np.log2(np.abs((meanA + 1e-9) / (meanB + 1e-9)))
+    stats_to_sort = np.abs(tscores)
 
     # find all with lfc > cutoff
-    lfc_above_cutoff_idx = np.nonzero(np.abs(logfoldchanges) > diffexp_lfc_cutoff)[0]
-    stats_to_sort = np.abs(tscores)
 
-    # derive sort order
-    if lfc_above_cutoff_idx.shape[0] > top_n:
+    if two_lists:
```
bkmartinjr, line 74: This code seems problematic:
* partitions twice, which is inefficient
* lots of redundancy

Why not just replace it with something which leverages a single (multi-pivot) partition, sorting by natural order (not abs), and just cuts from top & bottom of sorted list?

(caveat: typed this in without testing, so treat as psuedo-code)
(caveat: this is only the two_list variant, and it inlines derive_sort_order())

```
stats_to_sort = tscores
lfc_above_cutoff_idx = np.nonzero(np.abs(logfoldchanges) > diffexp_lfc_cutoff)[0]
rel_t_partition = np.argpartition(stats_to_sort[lfc], (top_n, -top_n))[-top_n:][:top_n]
t_partition = lfc[rel_t_partition]
rel_sort_order = np.argsort(stats_to_sort[t_partition])[::-1]
sort_order = t_partition[rel_sort_order]

logfoldchanges_top_n = logfoldchanges[sort_order]
pvals_top_n = pvals[sort_order]
pvals_adj_top_n = pvals_adj[sort_order]

result = {
    "positive": [[sort_order[i], logfoldchanges_top_n[i], pvals_top_n[i], pvals_adj_top_n[i]] for i in range(top_n)],
    "negative": [
        [sort_order[i], logfoldchanges_top_n[i], pvals_top_n[i], pvals_adj_top_n[i]] for i in range(-1, -1 - top_n, -1)
    ],
}
```

backend/server/compute/diffexp_generic.py
```diff
@@ -2,7 +2,7 @@
 from scipy import sparse, stats
 
 
-def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01):
+def diffexp_ttest(adaptor, maskA, maskB, top_n=8, diffexp_lfc_cutoff=0.01, two_lists=False):
```
bkmartinjr, line 5: As far as I can tell, this code is identical between czi_hosted/compute and server/compute/

Why is it not consolidated into backend/common?

backend/czi_hosted/common/rest.py
```diff
@@ -260,7 +260,9 @@ def diffexp_obs_post(request, data_adaptor):
     try:
         # TODO: implement varfilter mode
         mode = DiffExpMode(args["mode"])
-
+        two_lists = False
```
bkmartinjr, line 263: why do we need a mode switch?  Why not just remove the old code, and support only the "topN/bottomN" implementation?

this would simplify tests & code.

bkmartinjr (changes requested):
Several change requests:
1. remove the two_lists flag, and just have that behavior by default.  This will simplify the implementation and test cases.
2. the core of the diffexp seems identical between backends.  Can we move it into backend/common?
3. the "two_lists" changes to diffexp_ttest() could be made faster and simplified if you just sort/partition once, and take the topN/bottomN from the sorted list.  See comments left inline.

backend/server/compute/diffexp_generic.py
MDunitz: moved to backend/common

client/src/components/brushableHistogram/histogram.js
```diff
@@ -26,7 +26,13 @@ const Histogram = ({
     /*
       Create the d3 histogram
       */
-    const { marginLeft, marginRight, marginBottom, marginTop } = margin;
+    //  This is just a constant that's flipped by an `isMini` boolean
```
bkmartinjr, line 29: `isMini`?  No such variable.  Maybe `mini`?

bkmartinjr (approved):
one very tiny nit noted (obsolete comment).  Otherwise LGTM.

colinmegill (approved):

PR #2253, opened 2021-06-23
Added hotkeys & Invert current selection

atarashansky:
#### Reviewers
**Functional:** 
@colinmegill 
**Readability:** 
@Bento007 
---

## Changes
- add
Added inverse selection of cells, cell set setting, differential expression, and select all cells via hotkeys.
Added blueprint hotkeys (hold SHIFT+? to see available hotkeys)

- modify: 1) connected the app component to the differential state and added the hotkey react component to it. I’m passing both the dispatch function and differential as properties to the hotkey component (this is the bit I’m not sure is proper). 2) wrapped the application by the hotkey provider

The pattern for finding differentially expressed genes (one vs all) would be to lasso/select the cells of interest of interest, and do: SHIFT+1+I+2+D.

Disclaimer, I’m not super experienced with React so if I’m doing anything unorthodox, I really appreciate the feedback

### DIFF
--- a/client/src/actions/index.js
+++ b/client/src/actions/index.js
@@ -232,6 +232,23 @@ const requestDifferentialExpression = (set1, set2, num_genes = 50) => async (
   }
 };
 
+const selectAll = () => async (dispatch, getState) => {
+  dispatch({ type: "select all observations" });
+  try {
+    const { obsCrossfilter: prevObsCrossfilter } = getState();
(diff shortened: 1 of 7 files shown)


atarashansky:
Hotkey for differential expression (SHIFT+D) currently not working as it's passing null for `differential.celllist1/2`. But manually clicking the button works, so all upstream hotkeys are fine.

PR #2256, opened 2021-06-27
Added Hotkeys & Invert current selection action

atarashansky:
(I remade this pull request to be not from the main branch of my fork)

#### Reviewers
**Functional:** 
@colinmegill 
**Readability:** 
@Bento007 
---

## Changes
- add
Added inverse selection of cells, cell set setting, differential expression, and select all cells via hotkeys.
Added blueprint hotkeys (hold SHIFT+? to see available hotkeys)

The pattern for finding differentially expressed genes (one vs all) would be to lasso/select the cells of interest of interest, and do: SHIFT+1+I+2+D (or SHIFT+M+D).

### DIFF
--- a/client/src/actions/index.js
+++ b/client/src/actions/index.js
@@ -232,6 +232,23 @@ const requestDifferentialExpression = (set1, set2, num_genes = 50) => async (
   }
 };
 
+const selectAll = () => async (dispatch, getState) => {
+  dispatch({ type: "select all observations" });
+  try {
+    const { obsCrossfilter: prevObsCrossfilter } = getState();
+    const obsCrossfilter = await prevObsCrossfilter.selectAll();
+    return dispatch({
+      type: "selected all observations",
+      obsCrossfilter,
+    });
+  } catch (error) {
+    return dispatch({
+      type: "error selecting all observations",
+      error,
+    });
+  }
+};
+
 function fetchJson(pathAndQuery) {
   return doJsonRequest(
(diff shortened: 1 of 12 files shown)


signechambers1:
Thanks for the PR @atarashansky ! We're focused on prepping for a feature release right now but we'll get back to you with an update on our bandwidth to review your PR in 2 weeks. Thank you!

atarashansky:
After discussion with @colinmegill, I decided to remove the invert selection hotkey as it breaks the selection state of the world.

Now, SHIFT+I does the following: 
1) Set current selection to cell set 1
2) Inverts selection
3) Set current selection to cell set 2
4) Inverts selection again

I'm still inverting the selection because `setCellSetFromSelection` uses the selected cells. Alternatively, I could make a sister action that uses an input array of cells instead: `setCellSetFromInputArray`. That might be preferable for extremely large datasets as inverting the selection twice in the same command might cause some visual artifacts due to lag.

Also, @colinmegill told me that @bkmartinjr is suffering from having to manually delete the millions of genesets he is making, so I added a hotkey (`SHIFT+Q`) that removes the most recent geneset.

As a reminder, you can see a tooltip of all active hotkeys by pressing `SHIFT+?`.

atarashansky:
As mentioned in the last post, I added a function `setCellSetFromInputArray` so that I don't need to explicitly invert the selection anymore to set cell sets 1 and 2 (`ad5729f`).

PR #2278, opened 2021-07-06
docs: release 0.17.0

seve:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.17.0-rc.0
+current_version = 0.17.0
 commit = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:-(?P<prerel>rc)\.(?P<prerelversion>\d+))?
 serialize = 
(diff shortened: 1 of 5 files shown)


seve:
@MDunitz Do we also want to be keeping track of the version # in the hosted_czi server?

.bumpversion.cfg
```diff
@@ -1,9 +1,6 @@
 [bumpversion]
-current_version = 0.16.7
+current_version = 0.17.0
 commit = True
-# The below regex details an acceptable version number by naming the groups (major, minor, patch, prerel, and
```
MDunitz, line 4: why was this removed?

MDunitz:
> @MDunitz Do we also want to be keeping track of the version # in the hosted_czi server?

I dont think we need to track the version # in hosted_czi. For hosted the version should be based on the sha of the deployed commit

.bumpversion.cfg
seve: the cli removed it, none of these changes are manual

MDunitz (approved):
Lgtm

PR #2298, opened 2021-07-13
Dunitz/data adaptor metadata api

MDunitz:
#### Reviewers
**Functional:** 
@bkmartinjr 
**Readability:** 
@maniarathi 
---

## Changes

This is setting up hosted cellxgene to use  /datasets/meta?url={} in the portal to retrieve information (collection id, collection visibility, s3 uri, dataset id and tombstone status) about the dataset based on the url) and then pull the dataset based on the location given in the s3_uri.
For further detail see [ticket 2234](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/2234) and [RFC](https://github.com/chanzuckerberg/single-cell/blob/main/design-documents/0006-metadata-api/text.md)

## Approach
Split apart caching and dataset management. 
Add an additional cache to handle dataset location (redirects from explorer location).


backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -152,6 +155,21 @@ def data_adaptor(self, url_dataroot, location, app_config):
                 cache_item = info.cache_item
 
             if data_adaptor is None:
+                if app_config.server_config.data_locator__api_base:
+                    if url_dataroot == "e":  # only pull dataset identification information from the portal for /e/ route
+                        headers = {"Content-Type": "application/json"}
```
bkmartinjr, line 160: should include an Accept header given you are assuming JSON response below.

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -152,6 +155,21 @@ def data_adaptor(self, url_dataroot, location, app_config):
                 cache_item = info.cache_item
 
             if data_adaptor is None:
+                if app_config.server_config.data_locator__api_base:
+                    if url_dataroot == "e":  # only pull dataset identification information from the portal for /e/ route
```
bkmartinjr, line 159: this seems like a fragile assumption (the 'e' value).  Is it really fixed/always true?  Or should it be a config param?

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -152,6 +155,21 @@ def data_adaptor(self, url_dataroot, location, app_config):
...
+                        headers = {"Content-Type": "application/json"}
+                        curr_url = f"{app_config.server_config.get_web_base_url()}/{url_dataroot}/{dataset}"
```
bkmartinjr, line 161: `dataset` can be None, in which case I suspect the request below will not behave correctly.  Missing check?

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -152,6 +155,21 @@ def data_adaptor(self, url_dataroot, location, app_config):
...
+                        response = requests.get(
```
bkmartinjr, line 162: this can throw exceptions in many situations [docs](https://docs.python-requests.org/en/latest/api/#exceptions). I"m not sure that you want it to throw up the stack in the case where the Explorer API gateway can't talk to the Portal API service, unless all requests for the dataset are invalid when the API gateway is unreachable (or maybe you do - would be good to document the error path, as these network failures will definitely occur).

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -133,7 +136,7 @@ def __init__(self, max_cached, timelimit_s=None):
         self.timelimit_s = timelimit_s
 
     @contextmanager
-    def data_adaptor(self, url_dataroot, location, app_config):
+    def data_adaptor(self, url_dataroot, location, app_config, dataset=None):
```
bkmartinjr, line 139: All of the parameters combinations between app.py:get_data_adaptor() and this function are confusing, especially given the global (config) state used in some of the code branches.  The function signature seems to allow cases that aren't handled (eg, dataset==None and Portal API base URI is defined).

If I read the code right, you should never have a case where _both_ location and dataset are used and/or correct.  You might consider changing this function signature to more correctly match the function semantics.

My sense (with partial understanding) is that there are only three legal modes:
  1. single dataset, in which case the `location` is all you need
  2. multi-dataset, with Portal API used -- `dataset` is all you need
  3. multi-dataset, no Portal API -- `location` is all you need

Given this, the semantics are really a "redirect" based on location.  So you could change the function signature to something like the following and handle both:

    data_adaptor(self, url_dataroute, dataroot, dataset)

(and move the path_join from get_data_adaptor into this function)

Or, alternatively, create separate wrapper functions for both, and make the caller invoke the one with intended semantics?

bkmartinjr (changes requested):
Couple of suggestions:
* better error handling, or at least more obvious error handling, around the embedded HTTP request to the Portal API.  Needs to log the underlying error when the API HTTP request fails so it is debugable in production (at a minimum).
* the data_adaptor() context manager is getting complicated, with many branches, and some combinations of paramters not handled (but also not overtly rejected).  Suggest refactoring this for simplicity.  More comments inline.

backend/czi_hosted/data_common/matrix_loader.py
MDunitz: We only want to look it up for the `/e/` route because that is the route for datasets in the data portal. Other dataset paths wont be in the portal/should default to the name based lookup used historically.

backend/czi_hosted/data_common/matrix_loader.py
MDunitz: For case number 2 you still need the location to use as the key for the cache, otherwise the lookup will fail and we'll end up hitting the data portal endpoint for every request

backend/czi_hosted/data_common/matrix_loader.py
bkmartinjr: `/e/` is the current route on the portal, but that could change.  Just confirming that the intent here is to require a code change to the Explorer backend if the portal changes their routes (rather than adding a configuration parameter to Explorer, ie, `PortalDatasetBaseURL: "/e/"`)

I don't feel super strongly, but my point was that hard-coding routes like this is brittle - it forces a release (code change) to Explorer whenever Portal changes it's configuration.

bkmartinjr:
@MDunitz - please write a description.  Both for current reviewers as well as future readers (eg, I'm dealing with bugs due to regressions introduced by old PRs, which lack adequate descriptions, making it difficult to understand what motivated the PR).

backend/czi_hosted/data_common/matrix_loader.py
bkmartinjr: Point taken!   That still doesn't address the issue of how complex this particular function is and the number of independent routes through it (based on config changes).  It also spreads the logic of detering the S3 path across two separate modules (get_data_adaptor() uses `dataset` to make a physical path, and now data_adaptor() does the same thing with indirection via the poral).  Super likely this will be a source of future bugs IMHO.

Here is an alternative approach that (I think) solves both:
* do caching based on the `dataset` value
* add a new function which, given a `dataset`, returns two values:   the explorer path (ie, `join(dataroot, datapath)`, and the S3 dataset location.  Something like:   `getDatasetLocation(dataset:str) -> (explorer_path: str, S3_path: str)`
* All of the config-driven logic and communication with Portal are in that single function.  This function could even have its own cache to make things simpler (ie, don't rely on the adaptor cache)

This new function could be called by either get_data_adaptor or the matrix loader (assuming it has its own cache).  It makes the code in matrix_loader.py::data_adaptor()  more transparent, ie, you could define it as:

> def data_adaptor(self, base_url, explorer_dataset_url, S3_dataset_url)

And all of the config-related branching in data_adaptor() gets removed....

atolopko-czi (commented):
Unable to finish full review this week, so submitting as `Comment` for now in case any of my minor suggestions are of help. I'm still grokking the overall design changes, but generally makes sense! Will pickup on Mon.

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
+import threading
```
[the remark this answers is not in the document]
MDunitz: to make data retrieval more generalizable `acquire_and_open` was replaced with `get`, which accepts a lambda and a dictionary of arguments, but the cache invalidation and lock logic should be the same.

backend/czi_hosted/common/config/dataset_config.py
```diff
@@ -194,10 +199,7 @@ def handle_diffexp(self, context):
 
         server_config = self.app_config.server_config
         if server_config.single_dataset__datapath:
-            with server_config.matrix_data_cache_manager.data_adaptor(
-                self.tag, server_config.single_dataset__datapath, self.app_config
-            ) as data_adaptor:
-                if self.diffexp__enable and data_adaptor.parameters.get("diffexp_may_be_slow", False):
```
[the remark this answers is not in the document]
MDunitz: This should be safe to remove because we arent launching any single dataset environments. The code on line 201  (previously 196) `if server_config.single_dataset__datapath:`  and the code below it will be removed when we do the hard split.

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+
+    This is the intended usage pattern:
+
+           cache_manager = DataCacheManager(max_cached=..., timelimmit_s = ...)
+           with cache_manager.data_adaptor(location, app_config) as data_adaptor:
+               # use the data_adaptor for some operation
+    """
+
+    # FIXME:   If the number of active datasets exceeds the max_cached, then each request could
+    # lead to a dataset being deleted and a new only being opened: the cache will get thrashed.
```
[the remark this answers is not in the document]
MDunitz: I think the number of active datasets(in prod) is less than the max_cached, also I suspect our access may be somewhat bursty with lots of people accessing the same dataset at once (for example after a publication or author tweet). In the long run considering dataset size in addition to number of datasets for the caching of the actual dataset may be worth considering.  Because this is actively blocking the revisions workI think a redesign using the python tools is a little out of scope but would be good to keep in mind for the future

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+        try:
+            assert cache_item
+            if delete_adaptor:
+                delete_adaptor.delete()
+            # TODO @mdunitz can this be removed?
+            if desired_data is None:
+                desired_data = cache_item.get(cache_key)
+
+            yield desired_data
+        except (DatasetAccessError, DatasetNotFoundException):
```
bkmartinjr, line 173: where is DatasetNotFoundException thrown?

backend/czi_hosted/data_common/cache.py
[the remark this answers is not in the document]
bkmartinjr: I think this file needs reformatting to conform to our normal conventions (eg, line length).  At least vscode thinks so :-)

backend/test/test_czi_hosted/unit/common/config/test_server_config.py
```diff
@@ -260,32 +260,35 @@ def test_mulitdatasets_work_e2e(self, mock_render_template):
...
+
         assert data_config["config"]["parameters"]["about_legal_tos"] == "tos_set1.html"
 
         response = session.get("/set2/pbmc3k.cxg/api/v0.2/config")
 
         data_config = json.loads(response.data)
-        assert data_config["config"]["displayNames"]["dataset"] == "pbmc3k"
-        assert data_config["config"]["parameters"]["annotations"] is True
-        assert data_config["config"]["parameters"]["about_legal_tos"] == "tos_set2.html"
+        self.assertEqual(data_config["config"]["displayNames"]["dataset"], "pbmc3k")
```
bkmartinjr, line 269: too bad we don't use pytest!!!

backend/czi_hosted/data_common/dataset_metadata.py
```diff
@@ -0,0 +1,91 @@
+import json
```
bkmartinjr, line 1: this file also needs reformatting to flake8 rules (eg, line length)

backend/czi_hosted/data_common/dataset_metadata.py
```diff
@@ -0,0 +1,91 @@
...
+            return None
+    except Exception:
+        return None
+
+def extrapolate_dataset_location_from_config(server_config: ServerConfig, dataset_explorer_location: str):
+    """
+    Use the dataset_explorer_location and the server config to determine where the dataset is stored
+    """
+    # TODO @mdunitz remove after fork, update config to remove single_dataset option, the multiroot lookup will need to
+    #  remain while we support covid 19 cell atlas
```
bkmartinjr, line 37: I suspect the real requirement is not covid19cellatlas support, but need for stand-alone data collections that the Data Directory (aka Portal) does not manage.  Suggest getting this issue onto PM roadmap questions for a decision.

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+        #  TODO:  This is very simple.  When this class is used to cache the actual datasets
+        #   this can be improved by taking into account how much space is actually taken by each dataset,
+        #   instead of arbitrarily picking a max number to cache.
+        self.max_cached = max_cached
+
+        # items are automatically removed from the cache once this time limit (in seconds) is reached
+        self.timelimit_s = timelimit_s
+
+    @contextmanager
+    def data_adaptor(self, cache_key: str, create_data_lambda: Optional[Callable]=None, create_data_args: object={}):
```
bkmartinjr, line 137: This method name seems wrong, and a hold-over from when the cache only managed data adaptors.  I would generalize it, as it doesn't make a lot of sense from calling code.

Suggested name:   get(), ie,

```
    with cache.get(key, create_fn) as value:
        ....
````

File: backend/czi_hosted/common/config/dataset_config.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (bkmartinjr) [reply backend/czi_hosted/common/config/dataset_config.py:200]: In general, these heuristic warnings are not useful in the server environment (they just go into the logs or /dev/null)

File: backend/czi_hosted/default_config.py [hunk 14]
```diff
@@ -122,6 +122,15 @@
       # If timelimit_s is set to None, then there is no time limit.
       timelimit_s: 30
 
+    metadata_cache:
+      # The maximum number of datasets that may be opened at one time.  The least recently used dataset
+      # is evicted from the cache first.
+      max_datasets: 400
```
bkmartinjr, line 128: why 400?  they are small, right, so why not 10000 (or whatever)

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -217,19 +12,28 @@ class MatrixDataType(Enum):
 
 
 class MatrixDataLoader(object):
-    def __init__(self, location, matrix_data_type=None, app_config=None):
-        """ location can be a string or DataLocator """
+    def __init__(self, location, app_config=None, url_dataroot=None, matrix_data_type=None):
```
bkmartinjr, line 15: this routine would benefit from a bit more error checking, eg, throw on missing dataset_config.  It is cheap, and helps enforce assumptions downstream.

bkmartinjr (changes requested):
Overall, looking good!  Three suggestions:
1. Need a stand-alone unit test that exercises CacheItemInfo/CacheItem/CacheManager (ie, 100% test coverage on it, as it is a very, very critical module)
2. Lots of whitespace lint.  Run & fix flake8
3. Naming - several confusing names (like using `dataadaptor()` for the cache getter) could use improvement.

backend/test/test_czi_hosted/unit/common/config/test_server_config.py
MDunitz: Sorry I'm not sure I understand. I think that self.assertEqual works for regular python unittests. I prefer it because when it fails it tells you what didnt match instead of just raising an assertion error

backend/czi_hosted/default_config.py
MDunitz: I just picked a number larger than current dataset count. But 10000 probably gives us more room for growth

backend/czi_hosted/data_common/dataset_metadata.py
brianraymor: I will review with @ambrosejcarr. 

Thanks for reaching out @MDunitz.

backend/test/test_czi_hosted/unit/common/config/test_server_config.py
bkmartinjr: I was just lamenting that we still use unittest, instead of pytest. pytest nicely handles asserts (printing all of the useful info) and is generally much nicer than unittest.

Sorry - this was commentary.  I'm going to resolve this, as there is no action on your part!

backend/czi_hosted/default_config.py
bkmartinjr: Unless there is a compelling reason/feature for a small number, pick something that we are not likely to regret over the next few years!

backend/czi_hosted/data_common/cache.py
```diff
@@ -0,0 +1,219 @@
...
+            return self.data
+        self.data_lock.r_release()
+        if create_data_lambda is None:
+            return None
+
+        self.data_lock.w_acquire()
+        # the data may have been loaded while waiting on the lock
+        if not self.data:
+            try:
+                # self.data = self.loader.
```
maniarathi, line 38: Debugging line to delete?

backend/czi_hosted/data_common/matrix_loader.py
```diff
@@ -244,6 +48,7 @@ def __init__(self, location, matrix_data_type=None, app_config=None):
 
             self.matrix_type = CxgAdaptor
 
+    # TODO @mdunitz remove when removing conversion code, also remove server_config.multi_dataset__allowed_matrix_types
```
maniarathi, line 51: For any todos in code, make sure you create a new issue in the cellxgene repo and tag the issue number here so that we don't forget to eventually address the todos.

backend/czi_hosted/data_common/cache.py
```diff
@@ -155,7 +157,7 @@ def data_adaptor(self, cache_key: str, create_data_lambda: Optional[Callable]=No
                 cache_item = CacheItem()
                 desired_data = cache_item.get(
                     cache_key=cache_key,
-                    create_data_lambda=create_data_lambda,
+                    create_data_lambda=create_data_function,
```
bkmartinjr, line 160: any reason you use two different names for the same item/value?   (ie, lambda vs function)?

backend/test/test_czi_hosted/unit/common/utils/test_cache.py
```diff
@@ -0,0 +1,248 @@
...
+    def test_cache_manager_deletes_old_datasets(self):
+        cache_manager = CacheManager(max_cached=3, timelimit_s=2)
+        get_data = MagicMock()
+        get_data.return_value = {"cache_key": "cache_key"}
+        with cache_manager.get("key_one", lambda x: {x: x}) as cache_item:
+            self.assertIsNotNone(cache_item)
+        with cache_manager.get("key_one", get_data) as cache_item:
+            self.assertIsNotNone(cache_item)
+            self.assertEqual(get_data.call_count, 0)
+        sleep(3)
```
bkmartinjr, line 190: the '3' is a bit mysterious.  I suggest:
* define a constant called 'cacheTimeLimit_s' and set to 2.
* use that variable in the CacheManager constructor
* use that variable in sleep, eg, `sleep(cacheTimeLimit_s + 1)`
Makes it more obvious what you are doing, and makes it easier to keep the code in sync

bkmartinjr (approved):
Couple of very small items (suggestions really), but otherwise it LGTM.

The tests are dramatically better - really nice job.

backend/czi_hosted/data_common/cache.py
MDunitz: No they should all be the same. I'll update them

MDunitz:
Opened in new explorer repo see https://github.com/chanzuckerberg/single-cell-explorer/pull/3/files

Issue #2284, opened 2021-07-09
Improve performance of desktop DE algorithm


signechambers1:

signechambers1:
Closing

PR #2318, opened 2021-07-21
fix for incorrect stats computation in diff exp t-test

### LINKED ISSUE #2211: Incorrect differential expression with log-transformed data [BUG]
It appears that the differential expression (as log fold-change) calculation in `diffexp_ttest()` and `diffexp_ttest_from_mean_var()` assumes raw count data as input and will produce incorrect values if the expression data is log-transformed. The relevant code (in the latter function) is...
(line 67)   `logfoldchanges = np.log2(np.abs((meanA + 1e-9) / (meanB + 1e-9)))`
but for log-transformed expression data should be simply
`logfoldchanges = meanA - meanB`
For count data, the abs() call would be unnecessary unless counts can be negative. Also note that while the existing code would be appropriate for count-based expression, the accompanying use of t-tests to estimate _p_ values is not (since it assumes a normal distribution).
As calculated, the _sign_ of the differential expression will be correct, but not its magnitude. Also, the result is not monotonic with respect to the correct `meanA-meanB`, which means that rankings by log(FC) will also be incorrect.
Unless I'm mistaken and overlooking something this should be considered a high-priority fix!

### LINKED ISSUE #2282: a CLI param to control heuristic determination of approximate X data distribution
Add a new CLI parameter which allows the user to force cellxgene to assume a particular approximate X data distribution, for the purposes of calculating differential expression statistics.  See also chanzuckerberg/cellxgene#2211

Parameter name:  --X-approx-distribution

> cellxgene launch --help
> ...
> Options:
>   --X-approx-distribution [auto|normal|exponential]

Values:
* Auto - "sniff" the data, applying a heuristic to determine distribution.
* Normal - assume the data is normally distributed (what we commonly call "logged")
* Exponential - assume the data is exponentially distributed (what we commonly call "not logged"), and apply log1p() before calculating statistics.

Example which forces cellxgene to assume the data is not logged (ie, exponentially distributed, and will be logged before computing diffexp):
> cellxgene launch --X-approx-distribution=exponential mydata.h5ad

Example which forces cellxgene to assume the data is normally distributed, and will not be logged before calculating differential expression:
> cellxgene launch --X-approx-distribution=normal mydata.h5ad

The 'auto' value is default and will apply a heuristic to determine the approximate distribution of X values.

bkmartinjr:
The differential expression t-test will now correctly assume normally distributed data.  On the desktop version, it will determine the approximate distribution via a heuristic or user specification, and log the data if it is not normally distributed.

Change details:
* Change the t-test calculation to consistently assume normally distributed data, and to log data when it is not
* On desktop backend, add a heuristic algo to determine normal/exponential distribution, and a CLI param to control/override it when the user knows the distribution
* On the hosted backend, assume normally distributed data
* Update AnnData and H5PY to modern versions
* Fix several places where our use of AnnData slicing was incompatible with latest H5Py
* Update tests to match above and add tests for heuristic

Fixes #2211 
Fixes #2282 

Note: in the case where data is exponentially distributed (not logged), the changes to the t-test calculation will slow down diffexp modestly (due to the addition of the log1p()).  This performance issue will be addressed in later work (#2284).


backend/common/compute/estimate_distribution.py
```diff
@@ -0,0 +1,61 @@
...
+    return min_val, max_val
+
+
+def estimate_approximate_distribution(X) -> XApproxDistribution:
+    """
+    Estimate the distribution (normal, exponential) of the X matrix.
+
+    Currently this is based upon the assumption that scRNA-seq data is
+    exponentially distributed in its raw (count) form, and when logged,
+    any (max-min) range in excess of 24 is extremely unlikely.
```
ambrosejcarr, line 38:
```suggestion
    any (max-min) range in excess of 24 implies tens of millions of 
    observations of a single feature and so is extremely unlikely.
```
Suggest a note to explain explicitly why it is unlikely.

ambrosejcarr (approved):
I made a rookie error, but it's easily rectified. 

The Gaussian distribution is a [member of the exponential family](https://en.wikipedia.org/wiki/Exponential_family), so the nomenclature I suggested is insufficiently descriptive. Replacing `Exponential` with `count` rectifies this issue. See [count data](https://en.wikipedia.org/wiki/Count_data) for evidence. 

Note that instead of `normal` we could use `Guassian`, but I thought the capitalization would be more likely to result in curation errors, so suggest `normal`.)

The calculations made in this PR look good, I approve pending the nomenclature change.

maniarathi (approved):
LGTM. Can you also bump up the anndata/h5py versions here: https://github.com/chanzuckerberg/single-cell-infra/blob/main/terraform/modules/hosted-cellxgene/common/customize/requirements.txt

These are the versions that are used when deploying to the czi hosted platform.

MDunitz (approved):
lgtm

backend/czi_hosted/data_common/data_adaptor.py
```diff
@@ -158,7 +169,7 @@ def _index_filter_to_mask(self, filter, count):
         mask = np.zeros((count,), dtype=np.bool)
         for i in filter:
             if type(i) == list:
-                mask[i[0]: i[1]] = True
+                mask[i[0] : i[1]] = True
```
MaximilianLombardo, line 172: Asking this question out of curiosity - does the change in this line do anything or is it more of a style thing?

MaximilianLombardo (commented):
Lgtm! There was one change that I was curious about (and commented on), but everything else makes sense to me!

backend/czi_hosted/data_common/data_adaptor.py
bkmartinjr: Just lint (style, clarity) - no functional change.  We have a programmatic linter and a collection of rules that drive it.  I just ran the linter while I was working on the file.

MaximilianLombardo: Thank you for the explanation :)

PR #2320, opened 2021-07-21
fix: decrease the topN count explicitly on hosted

seve:
#### Reviewers
**Functional:** @MDunitz 

---

References #1281 as a TODO


MDunitz (approved):
lgtm

PR #2322, opened 2021-07-21
run make lint and make fmt

MDunitz:
#### Reviewers
**Functional:** 
@seve 
**Readability:** 
@bkmartinjr 
---

## Changes
- modify code by running make lint and make fmt and committing all of the changes. The lint test should now pass and future prs should have less extraneous lint changes


bkmartinjr:
@MDunitz - I don't think the makefile lint / format config matches the editor (package) definitions in pyproject.toml.  If these don't match, then we can't use autoformat or other editor hooks.  Can we sync these as part of this PR?

At least in vscode (for me), when I open files in this PR, vscode wants to immediate reformat some of them...

PR #2327, opened 2021-07-23
genesets e2e tests, undo/redo

colinmegill:
Tests undo / redo across genesets tests


backend/test/fixtures/pbmc3k-genesets.csv
```diff
@@ -16,6 +16,6 @@ summary test,,F5,
 summary test,,PIGU,
 geneset_to_delete,,,
 geneset_to_edit,,,
-fill_this_geneset,,RER1,
+fill_this_geneset,,,
```
seve, line 19: Did this change break the python tests?

colinmegill: Yes. @bkmartinjr updated the python spec in the previous e2e PR, but there was a mistake that we need to correct here — this geneset should be empty (so that it can be filled). Help appreciated, when I changed the python test in the previous PR, it did not behave as I anticipated. 

Python error: https://github.com/chanzuckerberg/cellxgene/pull/2327/checks?check_run_id=3139825571#step:7:1124

Line in question: https://github.com/chanzuckerberg/cellxgene/blob/colinmegill/undo-redo-e2e/backend/test/test_server/unit/common/test_api.py#L609

seve: You probably have to update the tests to reflect the changes you're making here.  There are assertions being made against this file in the python test suite, so be cautious when changing this file:

https://github.com/chanzuckerberg/cellxgene/blob/f150401ebff6bee345a3fb32ea19d62cff646464/backend/test/test_server/unit/common/test_api.py#L520-L584

client/src/components/util/truncate.js
```diff
@@ -85,13 +85,30 @@ export default (props) => {
...
+  );
+
   return (
     <Tooltip2
-      content={`${originalString}${tooltipAddendum}`}
+      content={
+        isGenesetDescription
+          ? descriptionContent
+          : `${originalString}${tooltipAddendum}`
+      }
```
seve, line 105: Why not always wrap it in a span and optionally pass a test-id prop to keep it more agnostic

client/src/components/util/truncate.js
```diff
@@ -85,13 +85,30 @@ export default (props) => {
...
       hoverOpenDelay={tooltipHoverOpenDelayQuick}
       targetProps={{ style: children.props.style }}
     >
       {newChildren}
     </Tooltip2>
   );
 };
+
+// data-test-id={`geneset-description-tooltip-${originalString}${tooltipAddendum}`}
```
seve, line 114: I think this was accidentally left in

seve (dismissed):

seve:
Should probably rebase onto main to accept the TS changes, also since we're editing truncate here, let's type it!

client/src/components/util/truncate.js
colinmegill: Snapshots.

client/src/components/util/truncate.js
colinmegill: Resolved in merge from main

client/src/components/util/truncate.tsx
```diff
@@ -89,9 +89,22 @@ export default (props: any) => {
       "aria-label": originalString,
     })
   );
+  // we need an ID to check for this content, since this is the only place the geneset description appears
+  const descriptionContent = (
+    <span
+      test-id={`geneset-description-tooltip-${originalString}${tooltipAddendum}`}
```
bkmartinjr, line 95: why not just put the test-id on the tooltip element (does it not propagate attributes?)

bkmartinjr (approved):
One minor question about possibly simplifying the addition of the test-id to the Tooltip2 element.  Otherwise LGTM.

client/src/components/util/truncate.tsx
colinmegill: My tests of that showed they are not propegated

client/src/components/util/truncate.js
seve: But it would still show in the snapshot, can you write an issue to not couple the truncate component to geneset descriptions.

PR #2332, opened 2021-07-24
clean up type inferencing

### LINKED ISSUE #2291: Selection on gene or gene set histograms not working in 0.17.0
**The performance issue**
On the Azimuth dataset, which has approximately 1 million cells and 77,000 genes, the following two use cases do not work: 

1. I can brush over the mean expression of a gene set to select a range of expression values
2. I can brush over an individual gene to select a range of expression values

Brushing over the quantitative metadata, which have been moved to the left sidebar, **does** work. 

**To Reproduce**
1. Download the [azimuth dataset](https://cellxgene.cziscience.com/collections/2f75d249-1bec-459b-bf2b-b86221097ced)
2. Download the [hallmark.csv](https://czi-sci.slack.com/archives/C02450V6L59/p1625709669042500) file containing hallmark gene sets in cellxgene format. 
3. `cellxgene launch local.h5ad --gene-sets-file hallmark.csv`
4. Expand `HALLMARK_IL6_JAK_STAT3_SIGNALING` 
5. Attempt to color by the average
6. Expand a gene within the gene set
7. Attempt to color by the gene

**Version (please complete the following information):**
 - Desktop
 - Version 0.17.0

### LINKED ISSUE #2309: [BUG] support float64 .X for cell selection from gene expression histogram.
Copying the comment I made [here](https://github.com/chanzuckerberg/cellxgene/issues/2303#issuecomment-883095375) as a fresh issue.

Using the demo data at https://cellxgene-example-data.czi.technology/pbmc3k.h5ad if I run
```
import scanpy as sc
import numpy as np
import scipy
adata = sc.read('pbmc3k.h5ad')
adata.X = scipy.sparse.csr_matrix(adata.X, dtype=np.float32)
adata.write('pbmc3k_32.h5ad')

adata = sc.read('pbmc3k.h5ad')
adata.X = scipy.sparse.csr_matrix(adata.X, dtype=np.float64)
adata.write('pbmc3k_64.h5ad')
```

Then the highlighting and selection as a group of cells by selecting a region of the gene expression histogram only works for pbmc3k_32.h5ad in cellxgene v0.17.0

This is a problem because h5ad files produced by [SeuratDisk](https://mojaveazure.github.io/seurat-disk/) seem to use float.64

Which is probably also the cause of https://github.com/chanzuckerberg/cellxgene/issues/2291 since the Azimuth data h5ad file was likely produced with SeuratDisk and this file looks like:
```
adata = sc.read('local.h5ad')
adata.X
<982538x77521 sparse matrix of type '<class 'numpy.float64'>'
	with 1748651019 stored elements in Compressed Sparse Row format>
```

**Screenshots**
Screen Shot 2021-07-20 at 23 06 49

**Expected behavior**
cellxgene supports this function for float64 .X matrix

### LINKED ISSUE #2333: H5AD/CXG representation of booleans is inconsistent
If a dataset contains a boolean column in obs/var:
* when viewed using the desktop (direct from H5AD), it will show in the UI as a categorical with labels "True" and "False"
* when viewed using hosted, after conversion to CXG, it will show in the UI as a categorical with labels "1" and "0"

Indirectly, this is caused by an attempt to work around TileDB's lack of a native bool type, and the front-end giving precedence to the OTA FBS type (rather than the REST API schema typing).

bkmartinjr:
This PR cleans up the type inferencing between AnnData and the REST FBS encoding and REST schema.  There are also some small tweaks to the front-end to normalize handling of boolean arrays for CXG.

Fixes #2309 
Fixes #2291
Fixes #2333 

Changes / punch list:
* [x] Documented the specific rules for converting between the typing systems.  Documented in the header of the type_conversions_util file which provides the API.
* [x] Rewrote the type conversion utilities to implement categorical conversion correctly.  There were many issues with handling Pandas categorical optimization where (pure) numeric categorical data is FBS-encoded in a binary format.  Changes:
        * the schema hint for categorical numeric types no longer specifies categories (front-end will correctly handle this case).  See #1779 for one reason why this is an improvement (rounding errors after cast).
        * where an int numeric category contains missing (NA), it will be converted to float and NA represented as NaN (this is in keeping with previous intent and Pandas conventions)
* [x] CXG and H5AD support for bool columns was inconsistent, resulting in different presentation on the front-end (CXG would display bool as "1" or "0", rather than "true" or "false".  This was likely a side-effect of tiledb not having a native "bool" type, and an cast to uint8, which the front-end did not correctly handle.  Front-end now takes lead from the schema rather than the transport data type.
* [x] Modified all schema and FBS generation to use this single type inteference API (previously they used a hodge-podge of code paths and inconsistent inferencing).
* [x] Added tests to cover (hopefully all of) the expected and error cases
* [x] Added 64 bit test fixture (the pbmc3k dataset with several items converted to 64 bit) for anndata adaptor
* [x] Update czi_hosted backend tests
* [x] Add tests to ensure that FBS and Schema types are always consistent 
* [x] Update CXG generation and tests specific to that code path (chanzuckerberg/corpora-data-portal#1290)


backend/common/utils/type_conversion_utils.py
```diff
@@ -17,129 +51,132 @@ def get_dtypes_and_schemas_of_dataframe(dataframe: pd.DataFrame):
...
 
-    if can_cast_to_int32(dtype, array_values):
-        return np.int32
-    if can_cast_to_float32(dtype, array_values):
-        return np.float32
-    if not can_cast_to_float32(dtype, array_values):
-        return np.float64
+    if dtype.kind == "f":
+        if np.can_cast(dtype, np.float32):
+            return (np.float32, {"type": "float32"})
```
maniarathi, line 92: These lines seem to be pretty much identical to 134-138. I wonder if there's a way to consolidate these functions a bit more?

backend/common/utils/type_conversion_utils.py
```diff
@@ -17,129 +51,132 @@ def get_dtypes_and_schemas_of_dataframe(dataframe: pd.DataFrame):
...
 
-    raise TypeError(f"Annotations of type {dtype} are unsupported.")
+    return None
 
 
-def get_schema_type_hint_from_dtype(dtype, array_values=None):
+def _get_type_info(array: Union[np.ndarray, pd.Series]) -> Tuple[np.dtype, dict]:
```
maniarathi, line 97:
```suggestion
def _get_type_info(array: Union[np.ndarray, pd.Series, pd.Index]) -> Tuple[np.dtype, dict]:
```

Based on the type check on line 103.

backend/common/utils/type_conversion_utils.py
```diff
@@ -17,129 +51,132 @@ def get_dtypes_and_schemas_of_dataframe(dataframe: pd.DataFrame):
...
     """
-
-    if dtype.kind == "f":
-        if not np.can_cast(dtype, np.float32):
-            logging.warning(f"Type {dtype.name} will be converted to 32 bit float and may lose precision.")
-
-        return True
-
-    if dtype.kind == "O" and array_values.hasnans:
+    # Accept any float as long as magnitude can still be represented.
```
maniarathi, line 148: This comment is not true no?

maniarathi: Oh wait, NVM. Magnitude.

backend/common/utils/type_conversion_utils.py
maniarathi: Ditto for other the other `can_cast` functions.

backend/common/utils/type_conversion_utils.py
```diff
@@ -17,129 +51,132 @@ def get_dtypes_and_schemas_of_dataframe(dataframe: pd.DataFrame):
...
+    if array.dtype.kind == "f":
+        if not np.can_cast(array.dtype, np.float32):
+            logging.warning(f"Type {array.dtype.name} will be converted to 32 bit float and may lose precision.")
         return True
 
     return False
 
 
-def can_cast_to_int32(dtype, array_values=None):
+def can_cast_to_int(array: Union[np.ndarray, pd.Series], target: np.dtype) -> bool:
```
maniarathi, line 157: In every use of this function, target is `int32`. Is there a reason you parametrized it?

backend/test/test_common/unit/utils/test_type_conversion_utils.py
```diff
@@ -213,5 +217,272 @@ def test__convert_pandas_series_to_numpy__int64_to_int32_with_nans_throws_error(
             convert_pandas_series_to_numpy(int_series, np.int32)
 
             self.assertIn(
-                "Cannot convert a pandas Series object to an integer dtype if it contains NaNs", logger.output[0]
+                "Cannot convert a pandas Series object to an integer dtype if it contains NA values.", logger.output[0]
             )
+
+
```
maniarathi, line 223: Seems like all the test cases above this line are redundant given the parametrized test cases below. Any reason to keep them around?

backend/test/test_common/unit/utils/test_type_conversion_utils.py
```diff
@@ -213,5 +217,272 @@ def test__convert_pandas_series_to_numpy__int64_to_int32_with_nans_throws_error(
...
+    }
+    for data in [
+        pd.Series(["a", "b", "c"], dtype="category"),
+        pd.Series(["a", "b", "c", 0, 1, 2], dtype="category"),
+        pd.Series(["a", "b", "c"], dtype="category").cat.remove_categories(["b"]),
+        pd.Series(["a", "b", "c", 0, 1, 2], dtype="category").cat.remove_categories(["b", 0]),
+    ]
+]
+
+category_numeric_OK_cases = [
```
maniarathi, line 357: There was a comment in the original `can_cast_to_int32` function that said roughly:

"If the array is categorical, then we need to order the array values so that functions min and max that occur later, can function. They do not function on unordered categories."

Can we add a test case to verify that the code still works even though those lines were removed?

backend/test/test_czi_hosted/unit/data_common/fbs/test_matrix.py
```diff
@@ -46,7 +46,7 @@ def test_encode_DataFrame(self):
                 "d": pd.Series(["x", "y", "z", "x", "y", "z", "a", "x", "y", "z"], dtype="category"),
             }
         )
-        expected_types = ((np.ndarray, np.float32), (np.ndarray, np.int32), (np.ndarray, np.uint32), (list, None))
+        expected_types = ((np.ndarray, np.float32), (np.ndarray, np.int32), (np.ndarray, np.int32), (list, None))
```
maniarathi, line 49: You have the same tuple listed twice (np.ndarray, np.int32)?

backend/test/test_common/unit/utils/test_type_conversion_utils.py
bkmartinjr: they are not fully redundant.  But I take your point and will clean it up.

backend/common/utils/type_conversion_utils.py
bkmartinjr: a bit, but not without propagating some assumptions downward. The core issue is that both are public API.  Fortunately, the can_cast_* does not need to be public, and is unused.  So I will remove those, and consolidate as you suggest.

backend/test/test_common/unit/utils/test_type_conversion_utils.py
bkmartinjr: I believe this was a comment added by your earlier PR, and I was unable to find any instance where the stated constraint was true.  Given that reordering changes the underlying type, it seemed unwise to do so without a clear need and API contract.  Looking at the bug report that triggered the original PR, I could not see a reason this change was required.

Do you recall why you added this?  I definitely recommend against it unless there is a corner case I am unaware of.

backend/test/test_czi_hosted/unit/data_common/fbs/test_matrix.py
bkmartinjr: true, but that was intentional as there are two columns which the test expects to return the int32 dtype ('b' and 'c')

backend/test/test_common/unit/utils/test_type_conversion_utils.py
maniarathi: Huh. I can't remember why I added it unfortunately.

client/src/annoMatrix/loader.ts
```diff
@@ -281,18 +281,62 @@ export default class AnnoMatrixLoader extends AnnoMatrix {
...
+
+  The OTA format does not follow precisely the same conventions. This routine implements
+  the front-end conventions given the schema and an OTA data column.
+  */
+  if (field === "obs" || field === "var") {
+    response = _booleanCast(field, query, schema, response);
+  }
+  if (field === "obs") {
+    /* 
+    Cough, cough - see comment on the function called.
```
MDunitz, line 305: I dont think I understand this comment at all

backend/common/utils/type_conversion_utils.py
```diff
@@ -17,131 +51,130 @@ def get_dtypes_and_schemas_of_dataframe(dataframe: pd.DataFrame):
...
+    if np.can_cast(array.dtype, np.int32):
+        return True
 
+    if array.size == 0:
+        return True
 
-def convert_pandas_series_to_numpy(series_to_convert: pd.Series, dtype):
-    if series_to_convert.hasnans and dtype == np.int32:
-        logging.error("Cannot convert a pandas Series object to an integer dtype if it contains NaNs.")
+    ii = np.iinfo(np.int32)
```
MDunitz, line 173: Could you give this a more informative name? Maybe machine_limits?

client/src/annoMatrix/loader.ts
bkmartinjr: lolz, yes.  Old comment from when the code was first created in 2020 (perhaps expressing some discomfort with the approach).

It is a pointer to the comment at the head of the _normalizeCategoricalSchema() function, which explains what it does.

Here is the comment (from schema.js):
```
/*
  Ensure all enum schema types have a categories array, that
  the categories array contains all unique values in the data
  array, AND that the array is sorted.
  Note that the back-end will not always set this hint, so we
  must assume it may be incorrect and/or missing.
  */
```

MDunitz (approved):
2 small comments but otherwise lgtm

client/src/annoMatrix/loader.ts
bkmartinjr: I cleaned up the comment.

backend/test/test_czi_hosted/unit/data_common/fbs/test_matrix.py
bkmartinjr: Assuming the response clarified - if not, please unresolve and add further info. TY!

Issue #2236, opened 2021-06-01
Add autosuggest to "add gene" and "create gene set" modal


signechambers1:
Note, figma is not finalized (text is unclear that it is a toggle): https://www.figma.com/file/D90Y9bGvlbaEr4IGb0a8bj/%231861-add-gene?node-id=0%3A1

signechambers1:
Closing as a duplicate of https://github.com/chanzuckerberg/cellxgene/issues/2219

PR #2335, opened 2021-07-26
update compat workflow to match latest deps

### LINKED ISSUE #2331: Update github compatibility workflows to match new requirements
We do not need to test for compatibility with anndata/scanpy versions we do not support.

from the current compatibility_tests.yml:
```
    strategy:
      matrix:
        python-version: [3.6, 3.7] # As of Oct 2020 Anndata is not compatible with 3.8
        anndata-version: [0.7.0, 0.7.1, 0.7.2, 0.7.3, 0.7.4, 0.7.5]
        test-suite: [smoke-test, smoke-test-annotations]
```

I recommend:
* Python 3.6, 3.7, 3.8 (3.9 is not quite ready, but there is another bug for that #2299)
* AnnData 0.7.6

bkmartinjr:
This PR modifies the compatibility test github workflow to test AnnData & Python versions that we currently support.

NOTE: this doesn't fix the broken tests (see #2236), just the compatibility test matrix.

Fixes #2331


MDunitz (approved):
lgtm

PR #2342, opened 2021-07-27
Correctly handle non-finite numbers in heuristic determination of X distribution

### LINKED ISSUE #2329: non-finite handling in X approx distribution heuristic
There is a bug in the heuristic determination of the X distribution -- it fails to account for NaN and +/- Inf.  While they are rare, they can theoretically occur in that data structure.

After discussion with Ambrose, we propose that the heuristic ignore these values for the purposes of finding the "range" of the data, and therefore guessing the distribution.

bkmartinjr:
The existing heuristic did not account for non-finites.  This PR:
* Adds explicit support to the algorithm for non-finites, plus a few other edge conditions (eg, sparse array in an  unsupported format)
* adds test cases for the above

Fixes #2329

### DIFF
--- a/backend/common/compute/estimate_distribution.py
+++ b/backend/common/compute/estimate_distribution.py
@@ -8,10 +8,6 @@ from backend.common.constants import XApproximateDistribution
 @numba.njit(error_model="numpy", nogil=True)
 def min_max(arr: np.ndarray):
     """Return (min, max) values for the ndarray."""
-    n = arr.size
-    odd = n % 2
-    if odd:
-        n -= 1
 
     # initialize to first finite value in array.  Normally,
     # this will exit on the first value.
@@ -21,8 +17,10 @@ def min_max(arr: np.ndarray):
             break
 
     # now find min/max, unrolled by two
+    odd = arr.size % 2
+    unrolled_loop_limit = arr.size - 1 if odd else arr.size
     i = 0
-    while i < n:
+    while i < unrolled_loop_limit:
         x = arr[i]
         y = arr[i + 1]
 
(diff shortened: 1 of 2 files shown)


ambrosejcarr (commented):
This looks good to me - one edge case I don't see explicitly tested is the empty array. Did I miss that? If no, does it make sense to add a specific error in that case?

bkmartinjr:
> This looks good to me - one edge case I don't see explicitly tested is the empty array. Did I miss that? If no, does it make sense to add a specific error in that case?

Empty array is tested on line 29 of test_est_dist.py.

```
# empty
        self.assertEqual(estimate_approximate_distribution(np.zeros((0,))), XApproximateDistribution.NORMAL)
```

bkmartinjr:
> This looks good to me 

Also, please explicitly approve the PR when you are happy w/ it.

ambrosejcarr (approved):

MDunitz (approved):
lgtm

backend/common/compute/estimate_distribution.py
```diff
@@ -5,27 +5,46 @@
...
     odd = n % 2
-    if not odd:
+    if odd:
         n -= 1
-    max_val = min_val = arr[0]
-    i = 1
+
+    # initialize to first finite value in array.  Normally,
+    # this will exit on the first value.
+    for i in range(arr.size):
```
MDunitz, line 18: can you use n instead of arr.size here? 
also could you rename n to array_length?

bkmartinjr: you can't use `n` here, as it may have been modified already and not be the actual array length (`n` is the loop unrolling limit).

I will improve variable naming.

PR #2355, opened 2021-07-30
disable formatting rules for eslint and add prettier in lint-staged

tihuan:
#### Reviewers
**Functional:** 
@bkmartinjr 

**Readability:** 
@seve @MillenniumFalconMechanic  @colinmegill 
---

## Changes
- modify
1. This PR removes `eslint-plugin-prettier`, since prettier doesn't recommend it ([source](https://prettier.io/docs/en/integrating-with-linters.html))
2. But we add `prettier` in `extend`, so it will turn off ALL formatting rules in eslint and let prettier play the sole formatter role
3. Formatted the code base via `npx prettier --write .` 
4. Add prettier in `lint-staged` hook
5. Updated package-lock to get latest eslint/prettier/etc.
6. New packages require two additional changes:
    * Set `@babel/plugin-proposal-private-methods` to `loose`
    * Update test snapshots to match latest blueprint/popover2 elements


bkmartinjr (changes requested):
This seems good, but we should also take this optty to bump the versions of the eslint/prettier and other lint/format packages for the front-end.  @MillenniumFalconMechanic is looking at a (likely) related issue where lint breaks when you update these packages, so would be good to fix in one pass.

bkmartinjr (approved):

seve (approved):

tihuan (commented):
LGTM!! Thanks so much for taking care of this!

MillenniumFalconMechanic:
Thanks Bruce, LGTM!

PR #2356, opened 2021-07-30
Thuang update gha unit test and add make lint client

tihuan:
#### Reviewers
**Functional:** 
@bkmartinjr 

**Readability:** 
@seve @MDunitz 
---

## Changes
1.  Change unit test GHA to run `make unit-test` instead of `make unit-test-server`, so we run `unit-test-server unit-test-client unit-test-common` as expected
2.  Add back `make lint` in client `Makefile`, so root `make lint` will trigger lint for both server and client!
3. Ran `make lint` and it formatted some python files 👌

### DIFF
--- a/backend/common/fbs/NetEncoding/Column.py
+++ b/backend/common/fbs/NetEncoding/Column.py
@@ -4,8 +4,9 @@
 
 import flatbuffers
 
+
 class Column(object):
-    __slots__ = ['_tab']
+    __slots__ = ["_tab"]
 
     @classmethod
     def GetRootAsColumn(cls, buf, offset):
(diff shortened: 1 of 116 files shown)


backend/common/fbs/NetEncoding/Column.py
```diff
@@ -4,8 +4,9 @@
 
 import flatbuffers
```
bkmartinjr, line 5: Do not reformat the contents of the `fbs/NetEncoding` directory - they are machine generated.

bkmartinjr:
This PR should not reformat any Python, as it is a f/e lint change.

bkmartinjr (changes requested):
Several requests:
1. we need to cleanly separate the `lint` and `fmt` commands. They should use the same configuration, but IMHO, `lint` is a read-only error check (pass/fail, no code mods), while `fmt` will auto-format the code
2. Please do _not_ reformat Python in this PR.  This is a front-end only PR and should have no impact on the Python code.

tihuan:
Will open a new one after the repo split! Thank you!

PR #2359, opened 2021-07-30
remove unused packages from client

bkmartinjr:
This PR removes a number of unused package dependencies from package.json.  Some of these are completely unused, some are used by sub-dependencies and should not be in the top-level package.json:
* react-easy-emoji
* react-popper
* connect-history-api-fallback
* style-loader
* sw-precache-webpack-plugin
* url-loader


tihuan (approved):
Woohoo! So much cleaner 🎉 Thanks so much, Bruce!

Issue #2366, opened 2021-08-03
Type src/util/dataFrame, src/util/typedCrossfilter, src/util/statemachine


tihuan:
Likely completion by **early-to-mid-Sept** given other commits & vaca.

PR #2374, opened 2021-08-11
undoable TS typing

bkmartinjr:
Partial solution to #2366.  Adds typing to undoable-related reducer, configuration and state machinery.

Note that there are still two eslint disables in this PR, which await typing of the selection reducer state.


client/src/reducers/undoableConfig.ts
```diff
@@ -185,62 +205,63 @@ Basic approach:
...
-  }
-  if (saveOnActions.has(actionType)) {
-    return { [actionKey]: "save", [stateKey]: filterState };
-  }
+const actionFilter =
+  (debug: boolean) =>
+  (
+    state: UndoableState<MyFilterState>,
+    action: AnyAction,
+    prevFilterState: MyFilterState | undefined
```
tihuan, line 213: nit:
```suggestion
    prevFilterState?: MyFilterState
```

client/src/reducers/undoableConfig.ts
```diff
@@ -185,62 +205,63 @@ Basic approach:
...
 
 /*
 return true if objA and objB are ===, OR if:
   - are both objects and not null
   - have same own properties
   - all values are strict equal (===)
 */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-function shallowObjectEq(objA: any, objB: any) {
+function shallowObjectEq(
```
tihuan, line 261: nit: I think `import { shallowEqual } from 'react-redux` does both `shallowObjectEq` and `shallowArrayEq` 🙆‍♂️ 

But out of scope of this PR!

```ts
// client/node_modules/react-redux/src/utils/shallowEqual.js
function is(x, y) {
  if (x === y) {
    return x !== 0 || y !== 0 || 1 / x === 1 / y
  } else {
    return x !== x && y !== y
  }
}

export default function shallowEqual(objA, objB) {
  if (is(objA, objB)) return true

  if (
    typeof objA !== 'object' ||
    objA === null ||
    typeof objB !== 'object' ||
    objB === null
  ) {
    return false
  }

  const keysA = Object.keys(objA)
  const keysB = Object.keys(objB)

  if (keysA.length !== keysB.length) return false

  for (let i = 0; i < keysA.length; i++) {
    if (
      !Object.prototype.hasOwnProperty.call(objB, keysA[i]) ||
      !is(objA[keysA[i]], objB[keysA[i]])
    ) {
      return false
    }
  }

  return true
}
```

tihuan (approved):
LGTM!! Coming together so well 🤩 !! Thanks again, Bruce!!

client/src/reducers/undoableConfig.ts
bkmartinjr: These are not equivalent.  The first says it is defined, but with a value === undefined.  The second says it may not be defined at all (ie, the Object.keys() will not return the `prevFilterState`).

Example: 
```
> Object.entries({'a': undefined})
[ [ 'a', undefined ] ]
```

or
```
> x = { 'a': undefined }
{ a: undefined }
> x.a === undefined
true
```

I specifically want the key defined, and the value to be === undefined.

seve (approved):

PR #2375, opened 2021-08-12
chore: extract schema types

tihuan:
PTAL thank you! 🙏


client/src/common/types/schema.ts
```diff
@@ -0,0 +1,58 @@
+type Category = number | string | boolean;
+
+export interface AnnotationColumnSchema {
+  categories?: Category[];
+  name: string;
+  type: "string" | "float32" | "int32" | "categorical" | "boolean";
+  writable: boolean;
+}
+
+interface DataFrameSchema {
```
bkmartinjr, line 10: This is more properly called something like `XMatrixSchema`

the `dataframe` key in the JSON was an unfortunate (legacy) misunderstanding by the author (me :-) that we never fixed!!!

client/src/common/types/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+  nObs: number;
+  nVar: number;
+  // TODO(thuang): Not sure what other types are available
+  type: "float32";
+}
+
+export interface LayoutColumnSchema {
```
bkmartinjr, line 17: Similarly, this would more appropriately be called `EmbeddingSchema`, as it does not describe a single dataframe column, and the word `layout` was another historically and very incorrect name.   Today we universally call these embeddings, not layouts.

client/src/common/types/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+  dataframe: DataFrameSchema;
+  layout: RawLayoutSchema;
+}
+
+interface AnnotationsSchema extends RawAnnotationsSchema {
+  obsByName: { [name: string]: AnnotationColumnSchema };
+  varByName: { [name: string]: AnnotationColumnSchema };
+}
+
+interface LayoutSchema extends RawLayoutSchema {
```
bkmartinjr, line 50: AFAIK, there is no rewriting of the layout schema (aka embedding schema), so you don't need this type.   The `Schema` and `RawSchema` should have the same type for the `layout` field.

Only `annotations` are rewritten to include `obsByName` and `varByName`

bkmartinjr (approved):
I left a couple of suggestions where our old JSON naming has been superceded by more conventional names that would be good to start using in the code.  That said, these are optional at this point.  The PR adds clarity as is, so I'm fine either way.

tihuan:
Fantastic!! Thanks so much for the feedback, Bruce 👍 Updated PR accordingly!

tihuan:
Just enabled automerge, since I think the update is low risk 🙆‍♂️ Thank you!

PR #2376, opened 2021-08-12
foundation (common) typing for array handling

bkmartinjr:
Utility classes and type guards for various arrays we commonly use.

As part of this, I removed old JS type-guard-like functions from utils/typeHelpers.ts, as they are now obsolete.


tihuan (approved):
Wow this offers so much clarity 😭 So beautiful!! LGTM to the max! Thanks so much, Bruce!!

PR #2377, opened 2021-08-12
fix: disable FE auth testing on compatibility tests

seve:
Disable the frontend auth testing for compatibility tests as they are breaking, and we don't run auth on explorer anymore.


MDunitz (approved):
lgtm

maniarathi (approved):

PR #2381, opened 2021-08-13
chore: type general utils

seve:
#### Reviewers
**Functional:** @bkmartinjr 

**Readability:** @tihuan @MillenniumFalconMechanic 

---
 All general utils other than centroids


client/src/components/framework/toasters.ts
```diff
@@ -26,8 +26,10 @@ export const keepAroundErrorToast = (message: any) =>
 /*
 a hard network error
 */
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-export const postNetworkErrorToast = (message: any, key = undefined) =>
+export const postNetworkErrorToast = (
+  message: string,
+  key: string | undefined = undefined
```
seve, line 31: I think this can just be:
```suggestion
  key?: string
```

client/src/util/camera.ts
```diff
@@ -11,30 +12,24 @@ const panBound = 0.8;
 const scratch0 = new Float32Array(16);
 const scratch1 = new Float32Array(16);
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-function clamp(val: any, rng: any) {
-  return Math.max(Math.min(val, rng[1]), rng[0]);
-}
```
seve, line 17: This was a duplicate of an existing util

client/src/util/camera.ts
```diff
@@ -53,14 +48,13 @@ class Camera {
     return this.viewMatrix[0];
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  pan(dx: any, dy: any) {
+  pan(dx: number, dy: number) {
```
seve, line 51: Looks like a missed a few return types in this class

client/src/util/finiteExtent.ts
```diff
@@ -21,13 +24,14 @@ function finiteExtent(tarr: any) {
       break;
     }
   }
-  for (; i < tarr.length; i += 1) {
-    const val = tarr[i];
-    if (Number.isFinite(val)) {
-      if (min > val) min = val;
-      if (max < val) max = val;
+  if (min && max)
```
seve, line 27: TS compiler couldn't see that for loop would be skipped in the case that min and max weren't defined

client/src/util/promiseLimit.ts
```diff
@@ -68,35 +86,25 @@ export default class PromiseLimit {
   Private below
   **/
 
-  // @ts-expect-error ts-migrate(7006) FIXME: Parameter 'priority' implicitly has an 'any' type.
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
-  _push(priority, fn, args) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-    const order = (this as any).insertCount;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-    (this as any).insertCount += 1;
```
seve, line 77: insertCount didn't exist, changed to insertCounter

client/src/util/quantile.ts
```diff
@@ -12,17 +12,19 @@ Arguments:
...
-export default function quantile(quantArr, tarr, sorted = false) {
+export default function quantile(
+  quantArr: Array<number>,
+  tarr: TypedArray,
+  sorted = false
+): Array<number> {
   /*
 	start with the naive (sort) implementation.  Later, use a faster partition
 	*/
-  const arr = sorted ? tarr : sortArray(new tarr.constructor(tarr)); // copy
```
seve, line 23: Might be wrong, but new keyword wasn't needed

bkmartinjr: it is wrong - doesn't do the same thing (won't make a copy)

You want:
```
const arr = sorted ? tarr : sortArray(new (tarr.constructor as TypedArrayConstructor)(tarr));
```

client/src/actions/index.ts
```diff
@@ -188,87 +189,89 @@ const dispatchDiffExpErrors = (dispatch: any, response: any) => {
...
+  ) =>
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  set2: any,
-  num_genes = 50
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-) => async (dispatch: any, getState: any) => {
-  dispatch({ type: "request differential expression started" });
-  try {
-    /*
+  async (dispatch: any, getState: any) => {
```
bkmartinjr, line 202: shouldn't the parameters be typed, eg, `(dispatch: Dispatch, getState: () => RootState)`?

client/src/common/types/entities.ts
```diff
@@ -1,3 +1,73 @@
...
 export {};
+
+export interface AnnotationColumn {
+  categories?: Category[];
+  name: string;
+  type: "string" | "float32" | "int32" | "categorical" | "boolean";
+  writable: boolean;
+}
+
+interface DataFrame {
```
bkmartinjr, line 12: I thought @tihuan was going to fix this name -- it is super confusing as it has nothing to do with a Dataframe.  It is the schema for the X matrix (ie, count matrix).

client/src/common/types/entities.ts
```diff
@@ -1,3 +1,73 @@
 // If a globally shared type or interface doesn't have a clear owner, put it here
 
 export {};
+
+export interface AnnotationColumn {
```
bkmartinjr, line 5: this would be better made into a union, so that it can reflect that `categories` is mandatory for some types, optional for others.

client/src/common/types/entities.ts
```diff
@@ -1,3 +1,73 @@
 // If a globally shared type or interface doesn't have a clear owner, put it here
 
 export {};
```
bkmartinjr, line 3: is this a typo?  If not, what purpose does it serve?

client/src/util/renderThrottle.ts
```diff
@@ -1,17 +1,16 @@
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-export default function renderThrottle(callback: any) {
+export default function renderThrottle(
```
bkmartinjr, line 1: I don't think this is right, as it doesn't preserve the type of `this`.  I believe you want something like:

```
export default function renderThrottle<T>(
  callback: (this: T) => void
): (this: T) => void {
  /*
  This wraps a call to requestAnimationFrame(), enforcing a single
  render callback at any given time (ie, you can call this any number
  of times, and it will coalesce multiple inter-frame calls into a
  single render).
  */
  let rafCurrentlyInProgress: number | null = null;
  return function f(this: T) {
    if (rafCurrentlyInProgress) return;
    rafCurrentlyInProgress = window.requestAnimationFrame(() => {
      callback.call<T, never, void>(this);
      rafCurrentlyInProgress = null;
    });
  };
}
```

You can test this with something like the following:
```
function bar(this: number) {
  console.log(this);
}
const foo = renderThrottle(bar);
foo.call("hi");
foo.call(1);
```

(the first `foo.call` should get a type error).

Two changes here: 
1. use a lambda which preserves `this` (so you don't need to alias it)
2. explicitly declare `this` type.

client/src/common/types/entities.ts
```diff
@@ -1,3 +1,73 @@
...
+  | Uint8ClampedArray
+  | Uint16Array
+  | Uint32Array
+  | Int8Array
+  | Int16Array
+  | Int32Array
+  | Float32Array
+  | Float64Array;
+
+export type NumericArray = TypedArray | Array<number>;
```
bkmartinjr, line 73: these two array-related items should be in common/types/arraytypes.ts.   `TypedArray` is already defined there.

Also, there is already a `NumberArray` defined in arraytypes.ts that is identical to your `NumericArray`.  Lets consolidate to one, and put it in arraytypes.ts

client/src/util/range.ts
```diff
@@ -22,38 +22,48 @@ rangeFill(array, start, step) -> array
 
 */
 
-// @ts-expect-error ts-migrate(7006) FIXME: Parameter 'arr' implicitly has an 'any' type.
-function _doFill(arr, start, step, count) {
+import type { NumericArray, TypedArray } from "../common/types/entities";
```
bkmartinjr, line 25: all array-related decls should be in common/types/arraytypes.ts

client/src/util/range.ts
```diff
@@ -22,38 +22,48 @@ rangeFill(array, start, step) -> array
...
+
+function _doFill<T extends TypedArray | Array<number>>(
```
bkmartinjr, line 27: Why not define the type signature as `T extends NumberArray`?

client/src/actions/index.ts
```diff
@@ -188,87 +189,89 @@ const dispatchDiffExpErrors = (dispatch: any, response: any) => {
...
-  } catch (error) {
-    return dispatch({
-      type: "request differential expression error",
-      error,
-    });
-  }
-};
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-function fetchJson(pathAndQuery: any) {
```
bkmartinjr, line 268: it would be best to make this a generic, so the caller can specify the expected return type.

eg, 
```
function fetchJson<ReturnType>(pathAndQuery: string): Promise<ReturnType> {
    ...
}
```

Ditto for doJsonRequest()

client/src/util/actionHelpers.ts
```diff
@@ -74,8 +76,10 @@ export const doFetch = async (url: any, init = {}) => {
 /*
 Wrapper to perform an async fetch and JSON decode response.
 */
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-export const doJsonRequest = async (url: any, init = {}) => {
+export const doJsonRequest = async (
```
bkmartinjr, line 79: should be a generic so caller can specify expected return type of the json object.

Ie, caller can do something like:
```
  const schema = doJsonRequest<Schema>("/path/to/schema");
```

client/src/util/actionHelpers.ts
```diff
@@ -110,13 +116,11 @@ Parameters:
...
 */
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
 export const rangeEncodeIndices = (
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  indices: any,
+  indices: Array<number>,
   minRangeLength = 3,
   sorted = false
-) => {
+): Array<number | Array<number>> => {
```
bkmartinjr, line 123: I believe this is the wrong signature (see the comment on the function).

I believe the return type should be `Array<number | [number, number]>`

client/src/util/catLabelSort.ts
```diff
@@ -11,24 +11,23 @@ TL;DR: sort order is:
...
+  isUserAnno: boolean,
+  values: Array<string>
+): Array<string> => {
   /* this sort could be memoized for perf */
 
-  const strings: string[] = [];
-  const ints: number[] = [];
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-  const unassignedOrNaN: any = [];
+  const strings = new Array<string>();
```
bkmartinjr, line 26: what is the benefit of this syntactic change?   Or is there some change in behavior implied?

client/src/util/finiteExtent.ts
```diff
@@ -6,8 +6,11 @@ If undefined or empty array, or array contains only non-finite numbers,
 will return [undefined, undefined]
 */
 
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-function finiteExtent(tarr: any) {
+import type { TypedArray } from "../common/types/entities";
+
+function finiteExtent(
+  tarr: TypedArray
+): [number | undefined, number | undefined] {
```
bkmartinjr, line 13: This signature is wrong as I read the code.  Return type should be:  `[number, number] | [undefined, undefined]`

I can't see a code path that lets you have one defined and the other undefined.

client/src/util/finiteExtent.ts
bkmartinjr: This introduces a serious bug -- if min/max are === 0, it won't do the second loop.

you want

`if (min === undefined & max === undefined) ...`

client/src/util/promiseLimit.ts
bkmartinjr: good catch!!!

bkmartinjr (changes requested):
Found a couple of "real bugs" that were introduced - those need to be fixed. The rest are suggestions for more accurate typing, better type style, etc.

client/src/util/catLabelSort.ts
seve: More readable of a line, is there more of a systematic change then the identifier lookup?

client/src/common/types/entities.ts
seve: I'm not sure how this file is getting committed by me, I'll take a look

client/src/util/range.ts
seve:
```
Argument of type 'NumericArray' is not assignable to parameter of type 'NumberArray'.
  Type 'Uint8ClampedArray' is not assignable to type 'NumberArray'.
    Type 'Uint8ClampedArray' is not assignable to type 'Float32Array'.
      Types of property '[Symbol.toStringTag]' are incompatible.
        Type '"Uint8ClampedArray"' is not assignable to type '"Float32Array"'.

File: client/src/util/catLabelSort.ts (hunk 18 above)
Comment (tihuan) [reply client/src/util/catLabelSort.ts:26]: Ah I think Airbnb style guide recommends using `[]` over `new Array()`!

https://github.com/airbnb/javascript#arrays

File: client/src/util/renderThrottle.ts (hunk 11 above)
Comment (tihuan) [reply client/src/util/renderThrottle.ts]: Ah actually should we just use `lodash.throttle` since by default it uses `rAF` too? 

Also not sure if it matters, but their implementation will invoke the callback with its arguments in addition to passing `this`

https://github.com/lodash/lodash/pull/3567/files#diff-277bea254c6f5a290a109a6cb6c796e89ef069c7964c3ac8095808d779630682R98

File: client/src/common/types/entities.ts (hunk 8 above)
Comment (bkmartinjr) [reply client/src/common/types/entities.ts]: bad merge.  He moved this stuff to schema.ts IIRC

File: client/src/util/renderThrottle.ts (hunk 11 above)
Comment (bkmartinjr) [reply client/src/util/renderThrottle.ts]: My view: try to minimize code changes unless you are committing to writing tests, absorbing the bugs, etc.  We don't need the args in our use cases (we pass state by `this`).

Also, historically we are on a long-term mission to remove lodash because of bloat.  Before introducing new dependencies, might want to discuss that.

KIS!

File: client/src/util/range.ts (hunk 14 above)
Comment (bkmartinjr) [reply client/src/util/range.ts:27]: Right - his just means that Uint8ClampedArray is missing from the TypedArray definition inside arraytypes.ts.  Please add it, as that was an oversight on my part...

File: client/src/util/catLabelSort.ts (hunk 18 above)
Comment (bkmartinjr) [reply client/src/util/catLabelSort.ts:26]: Yes, fully agree with airbnb - it is MUCH better IMO to use the JS-style and skip the generic.

File: client/src/util/renderThrottle.ts (hunk 11 above)
Comment (tihuan) [reply client/src/util/renderThrottle.ts]: pretty funny that lodash gave in and just typed args as any lol 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3kit/index.d.ts#L156

```ts
function throttle(fn: (...args: any[]) => void, delay: number): (...args: any[]) => void;
```

client/src/util/renderThrottle.ts
```diff
@@ -1,17 +1,16 @@
...
+  callback: () => void
+): (this: unknown) => void {
```
tihuan: I think the return should be a function that accepts arguments for the callback instead of `this`, since `this` is the execution context, not the arguments!

```ts
// lodash example
function throttle(fn: (...args: any[]) => void, delay: number): (...args: any[]) => void;
```

I just found that TS now has `Parameters` util! https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype

So we can do something like:

```ts
function throttle<T extends (...args: never[]) => void>(
  callback: T,
  timeoutMS: number
): (...args: Parameters<T>) => void { ... }
```

Inspiration: https://gist.github.com/ca0v/73a31f57b397606c9813472f7493a940

bkmartinjr: why add all this complexity and change if the current code suffices?  Using `this` to pass context is a very old tradition, and is equally good in this use case.

No issues with using explicit args, but it is change in behavior introducing more risk & code motion to this "typing" PR.

client/src/util/actionHelpers.ts
seve: I left doJsonRequest as an unknown return type and instead used the generic in `fetchJson`

client/src/actions/index.ts
seve: This function isn't edited in this PR

client/src/util/range.ts
seve: Done 👍

client/src/util/actionHelpers.ts
seve: (this was because `doJsonRequest` is only called by `fetchJson`)

client/src/actions/index.ts
```diff
@@ -36,29 +40,27 @@ async function userColorsFetchAndLoad(dispatch: any) {
...
-  return fetchJson("schema");
+async function schemaFetch(): Promise<{ schema: Schema }> {
+  return fetchJson<{ schema: Schema }>("schema");
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-async function configFetch(dispatch: any) {
-  return fetchJson("config").then((response) => {
-    const config = { ...globals.configDefaults, ...response.config };
+async function configFetch(dispatch: Dispatch) {
```
bkmartinjr, line 47: why not type the return as `Promise<Config>`?

client/src/actions/index.ts
```diff
@@ -36,29 +40,27 @@ async function userColorsFetchAndLoad(dispatch: any) {
...
+    config,
   });
+  return config;
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
 async function userInfoFetch(dispatch: any) {
   return fetchJson("userinfo").then((response) => {
-    const { userinfo: userInfo } = response || {};
+    const { userinfo: userInfo } = (response || {}) as { userinfo: unknown };
```
bkmartinjr, line 63: while this shuts up tsc/eslint, it doesn't really do much for typing (it is no better than suppressing the lint error).

why not define the full interface for user info, and declare that as a return type?

client/src/actions/index.ts
bkmartinjr: Sorry, I assumed you were cleaning up the entire file.  As long as the comment remains, all is fine.

client/src/util/actionHelpers.ts
bkmartinjr: not sure I understand the rationale for leaving doJsonRequest untyped.  You could have more flexibility, and exactly the same default typing if you declare it as:
```
export const doJsonRequest = async<T = unknown> (...) : Promise<T> {...}
```

this lets you remove the cast from fetchJson.  (sometimes casts are a hint that you may be able to improve things....)

bkmartinjr (approved):
My only suggestion is around fetchJson/doJsonRequest typing (don't cheat with an unknown).  Other than that, it LGTM.  Thanks for all the work here!!!

client/src/actions/index.ts
seve: Fixed!

client/src/actions/index.ts
seve: Fixed!

client/src/actions/index.ts
```diff
@@ -104,17 +108,17 @@ function prefetchEmbeddings(annoMatrix: any) {
...
+) => void) =>
+  catchErrorsWrap(async (dispatch: Dispatch) => {
     dispatch({ type: "initial data load start" });
 
     try {
       const [config, schema] = await Promise.all([
         configFetch(dispatch),
-        // @ts-expect-error ts-migrate(2554) FIXME: Expected 0 arguments, but got 1.
-        schemaFetch(dispatch),
+        schemaFetch(),
```
tihuan, line 121: NOTE: this is runtime code change, but should be safe since `schemaFetch` doesn't seem to accept arguments 👌

client/src/actions/index.ts
```diff
@@ -137,10 +141,9 @@ const doInitialDataLoad = () =>
       const layoutSchema = schema?.schema?.layout?.obs ?? [];
       if (
         defaultEmbedding &&
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-        layoutSchema.some((s: any) => s.name === defaultEmbedding)
+        layoutSchema.some((s: EmbeddingSchema) => s.name === defaultEmbedding)
       ) {
-        dispatch(embActions.layoutChoiceAction(defaultEmbedding));
+        embActions.layoutChoiceAction(defaultEmbedding);
```
tihuan, line 146: @seve I think we need to dispatch `embActions.layoutChoiceAction(defaultEmbedding);`, since it's a thunk?

tihuan (commented):
still reviewing but wanted to provide early feedback!

client/src/util/renderThrottle.ts
bkmartinjr: In general, we just haven't found lodash to be all that useful in most situations (eg, look at how they typed memoize).  In many cases, our code doesn't want the lodash generality, OR we prefer native (eg, don't use lodash map/filter/....)

I'm not making a strong argument, other than "why change the implementation when you are just adding types"?

client/src/components/framework/toasters.ts
tihuan: @seve let's change it to `key?: string` 🎉 !

client/src/util/actionHelpers.ts
```diff
@@ -39,22 +39,24 @@ export function catchErrorsWrap(fn: any, dispatchToUser = false) {
...
+    const acceptType = (init.headers as Headers)?.get("Accept");
     const res = await fetch(url, init);
     if (
-      res.ok &&
-      // @ts-expect-error ts-migrate(2531) FIXME: Object is possibly 'null'.
-      (!acceptType || res.headers.get("Content-Type").includes(acceptType))
+      (res.ok &&
+        (!acceptType ||
+          res.headers?.get("Content-Type")?.includes(acceptType))) ??
+      false
```
tihuan, line 59: @seve do you know why `?? false` was added? Thanks!

client/src/util/actionHelpers.ts
tihuan: Adding generic SGTM!

```suggestion
export const doJsonRequest = async <T = unknown>(
```

client/src/util/actionHelpers.ts
```diff
@@ -74,8 +76,10 @@ export const doFetch = async (url: any, init = {}) => {
 /*
 Wrapper to perform an async fetch and JSON decode response.
 */
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-export const doJsonRequest = async (url: any, init = {}) => {
+export const doJsonRequest = async (
+  url: string,
+  init?: RequestInit
+): Promise<unknown> => {
```
tihuan, line 82:
```suggestion
): Promise<T> => {
```

client/src/util/catLabelSort.ts
```diff
@@ -11,24 +11,23 @@ TL;DR: sort order is:
...
   const textA = String(a).toUpperCase();
   const textB = String(b).toUpperCase();
   return textA < textB ? -1 : textA > textB ? 1 : 0;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-const catLabelSort = (isUserAnno: boolean, values: any[]): any[] => {
+const catLabelSort = (
+  isUserAnno: boolean,
+  values: Array<string>
```
tihuan, line 22: For consistency:

```suggestion
  values: string[]
```

client/src/util/catLabelSort.ts
```diff
@@ -11,24 +11,23 @@ TL;DR: sort order is:
...
+): Array<string> => {
```
tihuan, line 23:
```suggestion
): string[] => {
```

client/src/util/finiteExtent.ts
```diff
@@ -21,14 +24,17 @@ function finiteExtent(tarr: any) {
       break;
     }
   }
-  for (; i < tarr.length; i += 1) {
-    const val = tarr[i];
-    if (Number.isFinite(val)) {
-      if (min > val) min = val;
-      if (max < val) max = val;
+  if (min !== undefined && max !== undefined) {
```
tihuan, line 27: @bkmartinjr I see this is runtime code modification, does this look good to you 😆 ? 
Is there a case that we want to enter the second loop even if only either `min` or `max` is `undefined`?

client/src/util/renderThrottle.ts
```diff
@@ -1,20 +1,17 @@
...
-  return function f(this: any) {
-    if (rafCurrentlyInProgress) return;
-    // eslint-disable-next-line @typescript-eslint/no-this-alias --- FIXME: disabled temporarily on migrate to TS.
-    const context = this;
+  let rafCurrentlyInProgress: number | null = null;
+  return function f(this: T) {
+    if (rafCurrentlyInProgress) return; // eslint-disable-next-line @typescript-eslint/no-this-alias --- required for functionality
     rafCurrentlyInProgress = window.requestAnimationFrame(() => {
-      callback.apply(context);
+      callback.call<T, never, void>(this);
```
tihuan, line 14: Keeping `apply()` would do, since we're not passing arguments, thus no difference between using `apply()` vs. `call()`?

```suggestion
      callback.apply<T, void>(this);
```

tihuan (commented):
Just went through everything! Thanks so much for the massive PR, @seve @colinmegill 🔥 Y'all amazing 🤩 👏 👏 👏

client/src/actions/index.ts
seve: Ah I thought I found a bug, didn't see returning function because of lack of `return`

client/src/util/finiteExtent.ts
bkmartinjr: the test is equivalent either way (&& or ||) as there is no way for only one to be undefined.  Either _both_ are undefined, or neither are undefined.

@seve - if you want to micro-optimize (perhaps even femto-optimize :-), you could change it to an ||.  But the logic is the same in any case.

tihuan (approved):
LGTMMM! Thanks so much again, @seve !!

@bkmartinjr can you take a last look at this please, given that it's runtime code change lol?

https://github.com/chanzuckerberg/cellxgene/pull/2381#discussion_r690738691

Thank you!!

bkmartinjr:
@seve - as far as I'm concerned, once you fix the lint and make this pass tests, it is good by me.

PR #2382, opened 2021-08-15
TS typing for Dataframe

bkmartinjr:
This PR includes:
* TS typing for Dataframe
* a few minor bug fixes turned up by static type analysis
* Partial upstream typing of Dataframe users.  (the PR was getting big, so I will finish the rest of the Dataframe-consumer typing in another PR).

Notable API changes:
* summarize() and histogram() are no longer polymorphic and have separate functions for categorical, continuous and by-category (eg, summarizeContinuous(), summarizeCategorical(), etc).
* LabelIndex.getOffset() and getOffsets() returns -1 instead of undefined if the label is unknown. 

This is another partial fix for #2366

### DIFF
--- a/client/src/util/dataframe/dataframe.ts
+++ b/client/src/util/dataframe/dataframe.ts
@@ -1,25 +1,41 @@
-import { IdentityInt32Index, isLabelIndex } from "./labelIndex";
-// weird cross-dependency that we should clean up someday...
 import { callOnceLazy, memoize, __getMemoId } from "./util";
-import { isTypedArray, isAnyArray } from "../../common/types/arraytypes";
 import {
-  summarizeContinuous,
+  isTypedArray,
+  isAnyArray,
+  AnyArray,
+  GenericArrayConstructor,
+} from "../../common/types/arraytypes";
+import { IdentityInt32Index, LabelIndex, isLabelIndex } from "./labelIndex";
+import {
+  summarizeContinuous as _summarizeContinuous,
   summarizeCategorical as _summarizeCategorical,
 } from "./summarize";
 import {
   histogramCategorical as _histogramCategorical,
+  histogramCategoricalBy as _histogramCategoricalBy,
   hashCategorical,
-  histogramContinuous,
+  histogramContinuous as _histogramContinuous,
+  histogramContinuousBy as _histogramContinuousBy,
   hashContinuous,
 } from "./histogram";
+import {
+  DataframeValue,
+  DataframeColumnarArray,
+  DataframeColumn,
+  OffsetType,
+  OffsetArray,
+  LabelType,
+  ColLabelArray,
+  RowLabelArray,
+} from "./types";
 
 /*
-Dataframe is an immutable 2D matrix similiar to Python Pandas Dataframe,
+Dataframe is an immutable 2D matrix similar to Python Pandas Dataframe,
 but (currently) without all of the surrounding support functions.
 Data is stored in column-major layout, and each column is monomorphic.
 
 It supports:
-* Relatively efficient creation, cloning and subsetting
+* Relatively efficient create, clone and subset operations
 * Very efficient columnar access (eg, sum down a column), and access
   to the underlying column arrays.
 * Data access by row/col offset or label.  Labels are reasonably well
@@ -27,10 +43,10 @@ It supports:
 
 It does not currently support:
 * Views on matrix subset - for currently known access patterns,
-  it is more effiicent to copy on subsetting, optimizing for access
+  it is more efficient to copy on subsetting, optimizing for access
   speed over memory use.
 * JS iterators - they are too slow.  Use explicit iteration over
-  offest or labels.
+  offset or labels.
 
 Important assumptions embedded in the API:
 * Columns are implicitly categorical if they are a JS Array and numeric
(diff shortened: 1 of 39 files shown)


client/src/util/dataframe/dataframe.ts
```diff
@@ -141,17 +163,13 @@ class Dataframe {
     Object.freeze(this);
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types --- FIXME: disabled temporarily on migrate to TS.
+  /** @internal */
```
tihuan, line 166: Ayyy @internal 🔥 !!

client/src/util/dataframe/dataframe.ts
```diff
@@ -242,14 +265,12 @@ class Dataframe {
     };
 
     /* test for row label inclusion in column */
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-    const has = function has(rlabel: any) {
-      const offset = getRowByOffset(rlabel);
-      return offset >= 0 && offset < length;
+    const has = function has(rlabel: LabelType) {
+      const offset = getRowOffset(rlabel);
+      return offset !== undefined && offset >= 0 && offset < length;
```
tihuan, line 270: super nit: do we need to check `offset !== undefined` if `getRowOffset()` always returns a number?

client/src/util/dataframe/dataframe.ts
```diff
@@ -683,23 +709,26 @@ class Dataframe {
...
       colIndex = this.colIndex.isubset(colOffsets);
     }
 
     return this.__subset(rowIndex, colIndex);
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  isubsetMask(rowMask: any, colMask = null, withRowIndex = null) {
+  isubsetMask(
+    rowMask: Uint8Array | boolean[] | null,
```
tihuan, line 728: Nice!! Types are so helpful here 🤩 👏

client/src/util/dataframe/labelIndex.ts
```diff
@@ -53,77 +87,79 @@ class IdentityInt32Index {
...
   }
 
-  // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types, @typescript-eslint/no-explicit-any -- - FIXME: disabled temporarily on migrate to TS.
-  getLabel(i: any) {
+  getLabel(offset: number): number | undefined {
     // offset to label
-    return Number.isInteger(i) && i >= 0 && i < this.maxOffset ? i : undefined;
+    return Number.isInteger(offset) && offset >= 0 && offset < this.maxOffset
+      ? offset
+      : undefined;
```
tihuan, line 101: does it make sense to return `-1` instead of `undefined` here to have consistency with `getOffset()` above ?

client/src/util/dataframe/labelIndex.ts
```diff
@@ -214,82 +237,70 @@ class DenseInt32Index {
...
+      const offset = index[lblIdx];
+      return offset;
     };
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-    this.getLabels = function getLabels(arr: any) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-      return arr.map((i: any) => this.getLabel(i));
+    this.getLabel = function getLabel(offset: number) {
+      return Number.isInteger(offset) ? labels[offset] : undefined;
```
tihuan, line 252: same question here for `undefined` vs. `-1`

client/src/util/dataframe/labelIndex.ts
```diff
@@ -383,48 +386,27 @@ class KeyIndex {
...
+      if (offset === undefined) return -1;
+      return offset;
     };
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-    this.getLabels = function getLabels(arr: any) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any --- FIXME: disabled temporarily on migrate to TS.
-      return arr.map((i: any) => this.getLabel(i));
+    this.getLabel = function getLabel(offset: number) {
+      return Number.isInteger(offset) ? rindex[offset] : undefined;
```
tihuan, line 397: same question here for `undefined` vs. `-1`

client/src/util/dataframe/types.ts
```diff
@@ -0,0 +1,145 @@
...
+  indexOf: (value: DataframeValue) => LabelType | undefined;
+
+  /**
+   * Return the value at the given offset, or undefined if not present.
+   */
+  iget: (offset: OffsetType) => DataframeValue | undefined;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any --- a legitimate use of any.
+export type AnyFunction<T = unknown> = (...args: any[]) => T;
```
tihuan, line 142: just curious if `unknown[]` would work here instead of `any[]`?

https://blog.logrocket.com/when-to-use-never-and-unknown-in-typescript-5e4d6c5799ad/

tihuan (approved):
LGTM!! Thanks for the massive PR, Bruce 🤩 👏 !! This looks amazing!

Just a few nits I could find 😆 Thanks again!

client/src/util/dataframe/dataframe.ts
bkmartinjr: good catch - that is a BUG!!!

client/src/util/dataframe/labelIndex.ts
bkmartinjr: No, it doesn't.  -1 is a _legal_ label.  

Remember:
* offsets are numeric indices.  `-1` is "no such offset"
*  labels are any string|bool|number.    `undefined` is "no such label"

This is documented in the comments for `getOffset()` and `getLabel()`

client/src/util/dataframe/labelIndex.ts
bkmartinjr: same answer as above.  See the abstract method definition.

client/src/util/dataframe/labelIndex.ts
bkmartinjr: same answer as above. See the abstract method definition.

client/src/util/dataframe/types.ts
bkmartinjr: No, it won't work as you can't call the memoized function without a type error.  See, for example, how TS defines the signatures to Function.call, Function.bind, etc.

What _would_ be better is to improve the type pass through for the memoize() function, which is the only user of this type.  I'll fix that for both memoize() and callOnceLazy().

Eg,
```
export function memoize<
  T extends (...args: any[]) => any = (...args: any[]) => any
>(
  fn: T,
  hashFn: (...args: Parameters<T>) => string,
  maxResultsCached = -1
): (...args: Parameters<T>) => ReturnType<T> {
    ...
}
```

Side note: fixing this pointed out a small typing bug!  Yay.

bkmartinjr: ps. if you haven't already done so, it is worth studying the type signatures for stuff like Function.bind, Function.apply, etc.

tihuan: Ahh didn't think of checking out those signatures! Thanks for the tip 👍 

Using `Parameters<T>` SGTM!

client/src/util/dataframe/labelIndex.ts
tihuan: ooh got it! Thanks for the explanation 🎉 Makes sense to me now 😄 !

client/src/util/dataframe/dataframe.ts
tihuan: Hurray! Do we need code change here? It seems to still be `offset !== undefined && offset >= 0 && offset < ` 😆

seve (approved):

client/src/util/dataframe/dataframe.ts
bkmartinjr: Only because the comment refers to the outdated code.    Take a look at the most recent commit to the PR, not the obsolete commit :-)

MillenniumFalconMechanic (approved):
LGTM, thanks Bruce!

PR #2406, opened 2021-08-23
X float16 support

### LINKED ISSUE #2379: [BUG] Cellxpose returns 400 Bas request for some requests if data is 16-bit
**Describe the bug**

Try saving test file after converting to fp16:
```
adata.X = adata.X.astype('float16')
```

Server normally starts when given this data (but prints some warnigns about conversion from 64bits).

However when you select gene or run differential expression, 400 appears. 

Version affected: 0.17.0, within docker

In my case cellxgene first downloads file from the cloud, so 16bit format for storage is very reasonable - almost halves the space.

bkmartinjr:
Add support for numpy float16 in the X matrix.  This includes:
* handling float16 in the JSON encoder
* adding an alternative min/max calculation due to lack of numba support for half width floats

Fixes #2379

### DIFF
--- a/backend/common/compute/estimate_distribution.py
+++ b/backend/common/compute/estimate_distribution.py
@@ -1,3 +1,4 @@
+from typing import Tuple 
 import numba
 import concurrent.futures
 import numpy as np
@@ -6,7 +7,7 @@ from backend.common.constants import XApproximateDistribution
 
 
 @numba.njit(error_model="numpy", nogil=True)
-def min_max(arr: np.ndarray):
+def min_max_fast(arr: np.ndarray) -> Tuple[float, float]:
     """Return (min, max) values for the ndarray."""
 
     # initialize to first finite value in array.  Normally,
@@ -47,6 +48,18 @@ def min_max(arr: np.ndarray):
     return min_val, max_val
 
 
+def min_max_numpy(arr: np.ndarray) -> Tuple[float, float]:
+    return arr.min(), arr.max()
+
+
+def numba_has_support_for_scalar_type(arr: np.ndarray) -> bool:
+    if not hasattr(arr, "dtype"):
+        return False
+    if isinstance(arr.dtype, (numba.types.Float, numba.types.Integer, numba.types.Boolean)):
+        return True
+    return False
+
+
 def estimate_approximate_distribution(X) -> XApproximateDistribution:
     """
     Estimate the distribution (normal, count) of the X matrix.
@@ -71,6 +84,8 @@ def estimate_approximate_distribution(X) -> XApproximateDistribution:
         )
     else:
         raise TypeError(f"Unsupported matrix format: {str(type(X))}")
+
+    min_max = min_max_fast if numba_has_support_for_scalar_type(Xdata) else min_max_numpy
 
     CHUNKSIZE = 1 << 24
     if Xdata.size > CHUNKSIZE:
--- a/backend/common/utils/utils.py
+++ b/backend/common/utils/utils.py
@@ -65,7 +65,7 @@ def path_join(base, *urls):
     return btpl._replace(path=path).geturl()
 
 
-class Float32JSONEncoder(json.JSONEncoder):
+class FloatJSONEncoder(json.JSONEncoder):
     def __init__(self, *args, **kwargs):
         """
         NaN/Infinities are illegal in standard JSON.  Python extends JSON with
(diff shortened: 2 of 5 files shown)


seve (approved):
Readability ✅

backend/common/utils/utils.py
```diff
@@ -78,9 +78,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def default(self, obj):
-        if isinstance(obj, np.float32):
+        if np.issubdtype(obj, np.floating) and not isinstance(obj, np.float16):
```
blrnw3, line 81: I don't understand how this is changing the existing behavior, since you are still gating it to float32 (albeit adding in support for 64 and 128). How will float16s get encoded?

backend/server/data_anndata/anndata_adaptor.py
```diff
@@ -232,7 +232,7 @@ def _validate_data_types(self):
                 "Anndata data matrix is sparse, but not a CSC (columnar) matrix.  "
                 "Performance may be improved by using CSC."
             )
-        if self.data.X.dtype != "float32":
+        if self.data.X.dtype > np.dtype(np.float32):
```
blrnw3, line 235: curious, where does it get truncated?

backend/common/compute/estimate_distribution.py
```diff
@@ -47,6 +48,21 @@ def min_max(arr: np.ndarray):
     return min_val, max_val
 
 
+def min_max_numpy(arr: np.ndarray) -> Tuple[float, float]:
+    return arr.min(), arr.max()
+
+
+def numba_has_support_for_scalar_type(arr: np.ndarray) -> bool:
+    """Numba does not support half-floats or non-scalars."""
```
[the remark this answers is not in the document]
bkmartinjr: that seems reasonable.  I don't think I've ever seen a dataset with either type but we might as well do this once and for all :-)

backend/common/compute/estimate_distribution.py
```diff
@@ -72,6 +88,8 @@ def estimate_approximate_distribution(X) -> XApproximateDistribution:
     else:
         raise TypeError(f"Unsupported matrix format: {str(type(X))}")
 
+    min_max = min_max_fast if numba_has_support_for_scalar_type(Xdata) else min_max_numpy
```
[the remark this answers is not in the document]
bkmartinjr: AFAIK, only useful here.  I don't tend to promote code on spec, but if you know of a use case, LMK.

backend/server/data_anndata/anndata_adaptor.py
bkmartinjr: common/fbs/matrix.py

backend/common/utils/utils.py
bkmartinjr: oh my, this code has so many issues, I can't believe I actually committed it.   Will fix.

backend/common/utils/utils.py
```diff
@@ -65,7 +65,7 @@ def path_join(base, *urls):
     return btpl._replace(path=path).geturl()
 
 
-class Float32JSONEncoder(json.JSONEncoder):
+class FloatJSONEncoder(json.JSONEncoder):
```
blrnw3, line 68: Minor: I suppose it's not really just a float encoder if it handles ints :)

backend/common/utils/utils.py
```diff
@@ -78,9 +78,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def default(self, obj):
-        if isinstance(obj, np.float32):
+        if isinstance(obj, np.floating):
             return float(obj)
-        elif isinstance(obj, np.integer):
+        if isinstance(obj, np.integer):
```
blrnw3, line 83: Curious, how does it handle the conversion of the `ndarray`? Or are we just using native python lists?

blrnw3 (approved):
Minor Qs/comments. Looks good

backend/common/utils/utils.py
bkmartinjr: true, but the entire thing only exists to handle NaN/Inf (which is floats). Otherwise, we would just skip this...

backend/common/utils/utils.py
bkmartinjr: via slicing.  ie,   

> type(np.zeros(10, dtype=np.float32)[0]) == np.float32

we aren't using native lists, but ndarray

backend/common/utils/utils.py
```diff
@@ -65,7 +65,13 @@ def path_join(base, *urls):
     return btpl._replace(path=path).geturl()
 
 
-class FloatJSONEncoder(json.JSONEncoder):
+class StrictJSONEncoder(json.JSONEncoder):
+    """
+    Custom JSON encoder set-up performing two tasks:
+    1. Strict JSON conformance with non-finite flaots (NaN, +/-Inf) via allow_nan=False
```
blrnw3, line 71: typo: `flaots`

blrnw3 (approved):
🦾

PR #2419, opened 2021-09-07
refactor: drop non-session auth from frontend

seve:
## Changes
- Dropped all references to jwt based auth


MDunitz (approved):
lgtm 
Were you able to launch the desktop version and create annotations with this removed?
Also this looks good re auth in the front end but can you also update the backend auth code to remove the non session based auth classes and remove the factory pattern?

seve:
> Also this looks good re auth in the front end but can you also update the backend auth code to remove the non session based auth classes and remove the factory pattern?

This ticket was only for the frontend, I can jump onto the backend ticket too otherwise I was going to work on getting the dataset drawer off of desktop.  I forgot to link the issue, sorry!

tihuan (approved):
LGTM! Should we use this as feature branch for its BE counterpart to merge into, before we merge both into `main`? This way once we remove auth from BE, we can test it E2E to make sure it works!

seve:
>  Should we use this as feature branch for its BE counterpart to merge into, before we merge both into main? This way once we remove auth from BE, we can test it E2E to make sure it works!

@MDunitz What do you think?

MDunitz:
> > Should we use this as feature branch for its BE counterpart to merge into, before we merge both into main? This way once we remove auth from BE, we can test it E2E to make sure it works!
> 
> @MDunitz What do you think?

Does it currently work E2E? I think it should be safe to merge as is. I can work on a pr to remove the BE code today

PR #2428, opened 2021-09-14
fix: unit tests and reenable running

seve:
#### Reviewers
**Functional:**  @tihuan 

---

A bump in our jest version began running our unit tests in a node environment rather than a browser.  This PR adds SSR checks in various places in order to not make attempts to call window where there is none present.  Also re-enables unit tests on the push_tests workflow.


client/src/components/framework/toasters.js
```diff
@@ -1,31 +1,33 @@
 import { Position, Toaster, Intent } from "@blueprintjs/core";
 
 /** Singleton toaster instance. Create separate instances for different options. */
-
-const ToastTopCenter = Toaster.create({
-  className: "recipe-toaster",
-  position: Position.TOP,
-  maxToasts: 4,
-});
+let ToastTopCenter = null;
```
tihuan, line 4: is this still needed now we have JSDom? Thanks!

client/src/components/framework/toasters.js
```diff
@@ -1,31 +1,33 @@
...
+if (typeof document !== "undefined") {
```
tihuan, line 5: same here!

client/__tests__/util/annoMatrix/serverMocks/index.js
```diff
@@ -1,11 +1,4 @@
 export const baseDataURL = "https://a.fake.url/api/v0.2";
 
-window.CELLXGENE = {
```
tihuan, line 3: likely can be reverted too?

client/src/globals.js
```diff
@@ -102,7 +102,7 @@ const CXG_SERVER_PORT =
 
 let _API;
 
-if (window.CELLXGENE && window.CELLXGENE.API) {
+if (typeof window !== "undefined" && window.CELLXGENE && window.CELLXGENE.API) {
```
tihuan, line 105: same here!

client/src/components/framework/toasters.js
seve: I dropped JSDom on the unit tests since the tests themselves don't need any browser API's

tihuan (approved):
Ahh I see what you mean now 👍 That makes sense! Thank you!

Issue #2414, opened 2021-08-30
Renaming gene sets w/o a description allows user to enter a duplicate name


signechambers1:
Steps to reproduce:
1. Launch tabula muris dataset, release candidate 0.18.0
2. Create gene set named "testone" with genes Sfn and Lgals7, no description
3. Create gene set named "testtwo" with genes Krt5, Krt14, no description
4. Click "edit gene name and description"
5. Clear the description
6. Rename "testtwo" to "testone"

Actual Result:
"testtwo" renames to "test one" and the original "testone" is clobbered

Desired Result:
Error message appears showing "There is already a gene set with that name" & blue "Edit gene set name and description" button is disabled

Additional clean up requests:
1. Current error message in the Create gene set modal says "There is already a geneset with that name". Desired message: "There is already a gene set with that name"

2. If a user creates a gene set without a description, the hover over and rename modal shows the description populated with "description", ie hover over shows "testone: description". The desired result is that the description should be blank

PR #2429, opened 2021-09-15
fix: more accurate validation on geneset name

seve:
#### Reviewers
**Functional:** @tihuan @colinmegill 

---

## Changes

- modify
  -  rewrite parts of the validation code to be more accurate, also change some copy

https://user-images.githubusercontent.com/8716829/133507093-d63501bd-c2cb-4811-bf52-737de9c5fdc8.mov

#2414

### DIFF
--- a/client/src/components/annoDialog.js
+++ b/client/src/components/annoDialog.js
@@ -42,7 +42,7 @@ class AnnoDialog extends React.PureComponent {
               <p
                 style={{
                   marginTop: 7,
-                  visibility: validationError ? "visible" : "hidden",
+                  visibility: errorMessage !== "" ? "visible" : "hidden",
                   color: Colors.ORANGE3,
                 }}
               >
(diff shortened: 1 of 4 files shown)


client/src/components/annoDialog.js
```diff
@@ -42,7 +42,7 @@ class AnnoDialog extends React.PureComponent {
               <p
                 style={{
                   marginTop: 7,
-                  visibility: validationError ? "visible" : "hidden",
```
seve, line 45: validation error doesn't necessarily mean we'll be showing a message

signechambers1:
LGTM! Great work, thanks Seve!

client/src/components/annoDialog.js
tihuan: Maybe add this comment in the code just in case someone changes it back to `validationError` lol

tihuan (approved):
LGTM! Thanks for improving this, Seve 🎉 🥇

Issue #2410, opened 2021-08-26
Update GH compatibility workflow


bkmartinjr:
The compatibility test matrix in the GH workflow should be updated to add tests of:
* [x] Python 3.9 (see #2299)
~* [ ] Python 3.10 (at least RC1, but use a later version if available)~
* [x] MacOS

Please also update the compatibility matrix to substitute the `latest` version of `anndata` in the matrix instead of anndata `master`.

maesen:
TEAM COMMENTS:
* 3.9 tests should pass. We don't need to fix all the tests.
* Open question: are we promising 3.10 compatibility for Desktop 1.0? Do we need to add this to the compatibility to the workflow if we aren't making promises about it?

atolopko-czi:
At weekly kickoff squad meeting, We agreed _not_ to pin anndata (or any package for that matter).  I _think_ we also decided that we do not need to test the anndata `main` branch explicitly. @maniarathi Can you confirm?

maniarathi:
Confirmed! No pinning of anndata and we will verify both the minimum version of anndata (today that is 0.7.6) and the latest released version of anndata (today that also happens to be 0.7.6, but we should make the test automatically pull whatever is the latest).

PR #2431, opened 2021-09-17
#2410 update compatibility test matrix

atolopko-czi:
#### Reviewers
**Functional:** 
@MDunitz 
@danieljhegeman 

**Readability:** 
n/a (no source code changes)

---

## Changes

Github Actions Workflow updates compatibility matrix:
- Added MacOS Catalina and Big Sur to test compat matrix
- Added Python 3.9 to test compat matrix, but avoid running 3.9 for matrix jobs that do not have `tables` pypi build available for the given env
- Maintains running tests on both cellxgene main branch and latest pypi release.
- Add explicit matrix exclusions for matrix combinations that will never pass (see comments).
- Numerous refactorings to the workflow config to simplify matrix. Basically a rewrite.
- The anndata pkg is now tested at a pinned release and at latest release, but no longer using `master` branch version. To limit cross-product explosion of matrix jobs, the pinned anndata version is only tested on py3.8 and cellxgene latest release.
- Run unit and smoke tests in a single job, to improve speed, reduce workflow complexity and the number of jobs. Also fixes the redundant testing of unit tests. Within each job, the unit and smoke tests are run in separate steps for ease of troubleshooting. 
- Fixed termination of backend server to allow both smoke tests to run within a single job (both attempt to use 5005 port, sequentially, but first server was not being terminated).
- ~Run `npm ci` setup to address jest failures. This was a shot-in-the-dark fix, and I'm not 100% confident about it.~ [I removed the existing `make ci` by accident, and then rediscovered it's necessity 🤦 ]
- Replaced `continue-on-error: true` with `fail-fast: false`, which allows all matrix jobs to run independently, while also ensuring the that entire workflow is flagged as failed if any matrix job fails

N.B.: **Failures are still occurring!** All failures occur for `smoke-test-annotations` and they appear to be intermittent. I recommend we fix this as a [separate story](https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/2433), since the essential work for updating the compatibility testing matrix is otherwise complete.

## Validation
Passing run (`smoke-test-annotations` disabled): https://github.com/chanzuckerberg/cellxgene/runs/3654109451?check_suite_focus=true


.github/workflows/compatibility_tests.yml
```diff
@@ -1,11 +1,19 @@
...
+#    - cron: '0 8 7 * 2'
+#  push:
+#    branches:
+#      - main
+
+# TEMPORARY
+# TODO: Remove before merge to main
 on:
-  schedule:
-    - cron: '0 8 7 * 2'
```
atolopko-czi, line 5: Above will be updated after PR approval, before merging.

.github/workflows/compatibility_tests.yml
```diff
@@ -1,11 +1,19 @@
...
   push:
     branches:
-      - main
+      - atolopko/2410-update-compatibility-test-matrix
```
atolopko-czi, line 16: Above will be updated after PR approval, before merging.all of t

.github/workflows/compatibility_tests.yml
```diff
@@ -23,85 +31,76 @@ jobs:
       - name: Build docker image
         run: docker build .
 
-  cellxgene-main-with-python-and-anndata-versions:
```
atolopko-czi, line 26: this file is more easily reviewed by _not_ looking at the diff; it's basically a rewrite

scripts/start_server_and_test
```diff
@@ -14,12 +14,14 @@ START_SERVER_SCRIPT="$1"
 PORT="$2"
 TEST_SCRIPT="$3"
 
-await_port --await-free "$PORT"
+await_port --await-free --timeout 120 "$PORT"
```
atolopko-czi, line 17: address transient failures due to server not shutting down quickly enough prior to trying to spin it up again for next set of smoke test

scripts/start_server_and_test
```diff
@@ -14,12 +14,14 @@ START_SERVER_SCRIPT="$1"
...
 
 eval "$START_SERVER_SCRIPT &"
 SERVER_PID=$!
 
 function finish {
+  SERVER_PID=`lsof -i -P -n | grep LISTEN | grep ${PORT} | awk '{ print $2 }'`
+  echo Killing server PID=${SERVER_PID}
```
atolopko-czi, line 24: the pid of the `eval` was not (always?) the web server pid, so killing that pid didn't work.

scripts/start_server_and_test
```diff
@@ -28,3 +30,4 @@ trap finish EXIT
 await_port "$PORT"
 
 eval "$TEST_SCRIPT"
+
```
atolopko-czi, line 33: because the world needs more newlines

scripts/start_server_and_test
danieljhegeman: ah, I had just added a hackier solution to this by moving to `kill -9`. This is much prettier. (the more pipes the better, right? :trollface: ).

It only just dawned on me a couple hours ago that this was the problem; I didn't realize the server process could be orphaned from python... 😞 ... and eventually my machine would clean up the process, sometimes after 30s, sometimes after 5 min. 🤷‍♂️

danieljhegeman: but this was annoying 😅

scripts/start_server_and_test
```diff
@@ -14,12 +14,13 @@ START_SERVER_SCRIPT="$1"
...
-await_port --await-free "$PORT"
+await_port --await-free --timeout 120 "$PORT"
 
 eval "$START_SERVER_SCRIPT &"
-SERVER_PID=$!
 
 function finish {
+  SERVER_PID=`lsof -i -P -n | grep LISTEN | grep ${PORT} | awk '{ print $2 }'`
+  echo Killing server PID=${SERVER_PID}
   kill $SERVER_PID || true
```
danieljhegeman, line 24: @atolopko-czi for my own edification, the point of `|| true` here?

scripts/start_server_and_test
danieljhegeman: This is to prevent it from just spinning endlessly?

Does this mean your solution below doesn't reliably kill it every time?

danieljhegeman (approved):
Thanks for addressing the cleanup of the test server as a side issue

scripts/start_server_and_test
atolopko-czi: Assuming it allows the entire script to return success stats code even if the kill fails, though that seems undesirable.  I left as is, since I wasn't entirely sure myself, assuming there was greater wisdom in its creation.

scripts/start_server_and_test
atolopko-czi: I didn't investigate too deeply, but it seemed like killing the parent process (the `make` process, in this case) kills the child process (the web server) when running on MacOS, but not Ubuntu. Or maybe it never worked on either...

scripts/start_server_and_test
atolopko-czi: The kill works but it still takes time, since it's SIGTERM, not a SIGKILL. So sometimes 60 sec was not enough for the next web server _start_ command, which waits for the port to be free.

scripts/start_server_and_test
MDunitz: 😢

MDunitz (approved):
lgtm

scripts/start_server_and_test
danieljhegeman: it definitely failed on MacOS half the time and was successful half the time...

PR #2432, opened 2021-09-17
Reverse gene list order

### LINKED ISSUE #2420: Quick gene search: The latest gene added should appear at the top of the list
Right now, when you add multiple genes in the Quick Gene Search bar, the latest gene goes to the bottom of the list. We would like the latest gene to show up at the top of the list based on user feedback and for consistency with the gene sets section (the latest gene set the user creates shows up at the top of the gene sets section).

colinmegill:
Fixes #2420


bkmartinjr:
@signechambers1 - quick question:  why is this a UI improvement?  It is still an arbitrary order (although it is better upon first addition).

I would have expected that the quick gene add would create an alphabetically sorted list of genes, or something that would make it easier to scan for a given gene.

No action necessary - just a comment!

bkmartinjr (approved):
LGTM

signechambers1:
@bkmartinjr thanks for the feedback! we actually got feedback from users that they wanted the most recent result on top so it was easier to find once they created it. We didn't hear requests for other sorting methods but if we get additional feedback or find through usability testing (coming up in Q4) that another default sort order is more intuitive we can revisit. For now, this PR at least treats the Quick Gene Search and Gene Sets sections consistently.

seve (approved):
LGTM. Quick question for @signechambers1:

Do we have any guarantees about the UX of the gene set csv? This fix introduces a divergence between the CSV and display within cellxene is no longer one and the same, csv is 1,2,3,4,...,N where cxg is N,.....,3,2,1.

signechambers1:
@seve #2420 only addresses the quick gene search, the section at the top of the right side bar. the individual genes added there aren't persisted so i dont think it affects the genesets CSV at all. Am I misunderstanding you?

Just checked out this branch and double checked that the gene sets CSV matches the UI so I don't see any issues there and the quick gene search is working as expected for me!

PR #2436, opened 2021-09-17
Upgrade min versions of dependencies [zh2426]

blrnw3:
https://app.zenhub.com/workspaces/single-cell-5e2a191dad828d52cc78b028/issues/chanzuckerberg/cellxgene/2426

* Removed restriction on Flask 2.0 since Flask-Restful now supports it
* Removed jinja requirement since Flask 2.0 enforces it (we were not using it, just in there to avoid a vuln)
* Add arg in setup.py to~enforce py3.6+

# Testing
Tested flask 2.0 on py3.7 and py3.9


bkmartinjr:
Can you comment why we are bumping to latest?  Our historical policy has been to only bump when we have a concrete need for a later version (eg, security vuln, API / feature enhancement, etc), as there is a premium on remaining flexlbiel wrt old dependencies (as long as they function correctly).

The PR description is not clear if we are bumping because "we can" or if because "we need a feature" (etc).  Would appreciate clarification.

blrnw3:
Not sure if there is any current pressing need to bump to the latest. My assumption from the ticket was that we are doing this to perhaps enable future feature development and avoid running into security issues/bugs, but I'll admit I don't have experience with best practices of minimum version maintenance on open python packages, and Google was not very helpful!

The current set of minimum versions (before my changes) suggest that some of them were simply the latest packages at the time they were added, rather than being thoughtfully chosen to be the oldest non-buggy, feature-acceptable versions, so would certainly be useful to understand best practices here 😄 .

bkmartinjr:
> would certainly be useful to understand best practices here

Ultimately this is a PM policy question.  Suggest pulling a quick mtg together to nail it down. I'm happy to participate or sit it out, as useful.  I agree with you that the policy is (perhaps both) partially communicated and inconsistently applied!

blrnw3:
@bkmartinjr updated to just have the flask and jinja changes as discussed

bkmartinjr (approved):
LGTM.  Approved (contingent on merging with latest from main)

Issue #2170, opened 2021-04-20
Update `husky` package (+ frontend package bump)


bkmartinjr:
Please update the `husky` package to the latest/greatest version. Please also bump up other frontend packages to the latest/greatest if addressing a security vulnerability or addresses some sort of instability issue only.

~We have webpack major bump. Try to center the package updates around that. Ask @seve for help :)~

MDunitz:
Alternative approach
- remove husky
- port over pre commit linting from data portal

PR #2442, opened 2021-09-21
chore: update frontend packages

seve:
#2170


tihuan (approved):
LGTM! Thanks for upgrading this!

PR #2453, opened 2021-09-29
Create CNAME file to point docs to the new website

ebezzi:
#### Reviewers
**Functional:** 
@atolopko-czi 

---

## Changes
- Add a CNAME file that will redirect the current github pages (hosted at https://chanzuckerberg.github.io/cellxgene/) to the new website.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+docs.cellxgene.cziscience.com


atolopko-czi (approved):
Let's give this a shot, since it's not clear _how_ this works, but the Interwebs claim it will. 😄 

If it works, we should document what the heck this is doing, though I'm not sure where, unless a comment is allowed in this file.  Maybe we add an `Online Documentation` section somewhere, since I also wanted to doc the CNAME we added for the new gitbooks site.

PR #2483, opened 2022-02-09
Fix float16 support [#2379]

blrnw3:
Convert to float32 on startup unless backed, in which case error. 
scipy does not support complex slicing from float16 data so this is the easiest fix for now.

#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add
- remove
- modify

### DIFF
--- a/server/data_anndata/anndata_adaptor.py
+++ b/server/data_anndata/anndata_adaptor.py
@@ -236,6 +236,14 @@ class AnndataAdaptor(DataAdaptor):
             warnings.warn(
                 f"Anndata data matrix is in {self.data.X.dtype} format not float32. " f"Precision may be truncated."
             )
+        if self.data.X.dtype < np.float32:
+            if self.data.isbacked:
+                raise DatasetAccessError(f"Data matrix in {self.data.X.dtype} format is not supported in backed mode."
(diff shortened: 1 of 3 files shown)


server/data_anndata/anndata_adaptor.py
```diff
@@ -236,6 +236,14 @@ def _validate_data_types(self):
             warnings.warn(
                 f"Anndata data matrix is in {self.data.X.dtype} format not float32. " f"Precision may be truncated."
             )
+        if self.data.X.dtype < np.float32:
+            if self.data.isbacked:
+                raise DatasetAccessError(f"Data matrix in {self.data.X.dtype} format is not supported in backed mode."
+                                         " Please reload without --backed, or convert matrix to float32")
+            warnings.warn(
+                f"Anndata data matrix is in unsupported {self.data.X.dtype} format so will be recast to float32"
```
bkmartinjr, line 244: Nit: suggest grammatical improvement to message eg,
>  Anndata data matrix is in unsupported {self.data.X.dtype} format -- will be cast to float32.

bkmartinjr (approved):
LGTM.  One optional minor comment on an message.

PR #2501, opened 2022-03-10
add capability to store annotations on s3/gcs

arogozhnikov:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes
- add support for cellxgene annotations to be on s3.

After this patch, cellxgene-gateway can edit and store annotations on s3 out of the box.

### DIFF
--- a/server/common/annotations/local_file_csv.py
+++ b/server/common/annotations/local_file_csv.py
@@ -95,13 +95,14 @@ class AnnotationsLocalFile(Annotations):
 
             fname = self._get_celllabels_filename(data_adaptor)
             self._backup(fname)
+            locator = DataLocator(fname)
             if not df.empty:
-                with open(fname, "w", newline="") as f:
+                with locator.open("w") as f:
                     if header is not None:
                         f.write(header)
                     df.to_csv(f)
             else:
-                open(fname, "w").close()
+                locator.open("w").close()
 
             # update the cache
             self.last_label_fname = fname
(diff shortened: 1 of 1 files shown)


arogozhnikov:
can I get this reviewed @bkmartinjr ? Thanks!

blrnw3 (commented):
If we want this class to fully support s3-type paths (good idea), I think we'll need to change a few more things, basically anywhere `os.path` is used, e.g. in the `_backup` method. Your existing changes look good, so you should just be able to make similar changes in the rest of the methods.
Also I would advise changing the class name to exclude the word `local`.
Thanks

arogozhnikov:
@blrnw3 done, please review and merge

blrnw3:
Thanks @arogozhnikov . Changes lgtm. I do think we should change the name of the class and try to update our docs (if they exist?) but I'll defer to our project lead @seve on that point. 
Finally we'll need to fully test these changes before merging in. I'm not able to do that so will defer once more to @seve , thanks.

blrnw3:
Thanks @arogozhnikov . Changes lgtm. I do think we should change the name of the class and try to update our docs (if they exist?) but I'll defer to our project lead @seve on that point. 
Finally we'll need to fully test these changes before merging in. I'm not able to do that so will defer once more to @seve , thanks.

arogozhnikov:
Hi @seve can you please review and merge?

maniarathi:
Hi @arogozhnikov -- @seve is out this week, but will be back next week if that timeline is alright with you?

bkmartinjr (changes requested):
I like the modifications to use DataLocator - great enhancement.

Unfortunately, this change has broken the backup function. Could you resolve that and hit me up for another review?  I left some notes about what was causing the problem.

arogozhnikov:
thanks for pushing it to production-ready state @bkmartinjr

PR #2524, opened 2022-05-27
fix: remove unnecessary postgres req

MDunitz:
#### Reviewers
**Functional:** 
- @maniarathi 
**Readability:** 

---

## Changes
- remove unnecessary postgres dependency


maniarathi (approved):
LGTM. This dependency appears to have been added in [this PR](https://github.com/chanzuckerberg/cellxgene/pull/1619/files) when we added the DB for supporting annotations in the hosted version of cellxgene. That code is now split out into its own repo and there is no longer any additional need to support that DB in this version of cellxgene.

PR #2536, opened 2022-07-08
use fsspec.download for S3 artifacts + additional logging

ebezzi:
#### Reviewers
**Functional:** 
@bkmartinjr 

**Readability:** 

---

## Changes
- Replaces the old method to download a remote h5ad with `fsspec.download`, which should use buffering instead of loading the whole file in memoy.
- When the S3 download function fails, no additional logs were displayed. Now they are (in verbose mode).

### DIFF
--- a/server/common/utils/data_locator.py
+++ b/server/common/utils/data_locator.py
@@ -113,10 +113,9 @@ class DataLocator:
         # do our best to create a file with the same.
         ext = os.path.splitext(self.path)
         suffix = None if ext[1] == "" else ext[1]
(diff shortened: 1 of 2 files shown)


bkmartinjr (approved):
changes look OK and run fine on my laptop. Any idea why the unit & smoke tests fail?

ebezzi:
> changes look OK and run fine on my laptop. Any idea why the unit & smoke tests fail?

I believe those fail even in main (and in other recent PRs as well). I will look into them separately.

Issue #2566, opened 2022-09-08
Miscellaneous cli annotate fixes


atolopko-czi:
- [x]  `cellxgene --help`: `annotation` --> `annotate`
- [x] Required variables should not be options
- [x] Clarify `"“This option is mutually exclusive with --output-h5ad-file.”` in help
- [x] Error msg option names should use dashes not underscores: `--update_h5ad_file or --output_h5ad_file must be specified`
- [x] Use "gene names" not "gene identifiers" in help text for `--gene-column-name`
- [x] Message for annotations location is incorrect for --update-h5ad-file: `The output was Wrote annotations to None.`

PR #2568, opened 2022-09-09
feat: annotate command improvements

atolopko-czi:
#2566 

#### Reviewers
**Functional:** 
@seve 

**Readability:** 
@MaximilianLombardo: please review `cellxgene annotate --help` and review changes from UX perspective.

---

## Changes
* Replace --input-h5ad-file with a positional argument, for consistency with other CLI commands
* Replace --update-h5ad-file with --overwrite, for consistency with `prepare` command.
* Fix/clarify various help descriptions
* Fix final output message when input file is overwritten

The revised help output:
```
Usage: annotate.py <options> <path to H5AD input file>

  Add predicted annotations to an H5AD file. Run `cellxgene annotate --help`
  for more information.

Options:
  -c, --annotation-prefix TEXT    An optional prefix used to form the names
                                  of: 1) new `obs` annotation columns that
                                  will store the predicted annotation values
                                  and confidence scores, 2) `obsm` embeddings
                                  (reference and umap embedding), and 3) `uns`
                                  metadata for the prediction operation
                                  [default: cxg]
  --classifier TEXT               For cell type annotation, the classifier
                                  level to use. The classifier is model-
                                  dependent, so refer to documentation for the
                                  specified model for valid values.
  -l, --counts-layer TEXT         If specified, raw counts will be read from
                                  the AnnData layer of the specified name. If
                                  unspecified, raw counts will be read from
                                  `X` matrix, unless 'raw.X' exists, in which
                                  case that will be used.
  -g, --gene-column-name TEXT     The name of the `var` column that contains
                                  gene names. The values in this column will
                                  be used to match genes between the query and
                                  reference datasets. If not specified, the
                                  gene names are expected to exist in
                                  `var.index`.
  -h, --help                      Show this message and exit.
  --mlflow-env-manager [virtualenv|conda|local]
                                  Annotation model prediction will be
                                  installed and executed in the specified type
                                  of environment. MacOS users on Apple Silicon
                                  (arm64, M1, M2, etc.) are recommended to use
                                  'conda' to avoid Python package installation
                                  errors. If 'conda' is specified then
                                  cellxgene must also have been installed
                                  within a conda environment
  --model-cache-dir TEXT          Local directory used to store model files
                                  that are retrieved from a remote location.
                                  Model files will be read from this directory
                                  first, if they exist, to avoid repeating
                                  large downloads.
  -m, --model-url TEXT            The URL of the model used to prediction
                                  annotated labels. May be a local filesystem
                                  directory or S3 path (s3://)  [required]
  --organism [Homo sapiens|Mus musculus]
                                  For cell type annotation, the organism of
                                  the dataset. Used to normalize gene names to
                                  HGLC conventions when an annotation model
                                  has been trained using data from different
                                  organism.
  -o, --output-h5ad-file <filename>
                                  The output H5AD file that will contain the
                                  generated annotation values. If this option
                                  is not provided, the input file will be
                                  overwritten to include the new annotations;
                                  in this case you must specify --overwrite.
  --overwrite                     Allow overwriting of the specified H5AD
                                  output file, if it exists. For safety, you
                                  must specify this flag if the specified
                                  output file already exists or if the
                                  --output-h5ad-file option is not provided.
  -n, --run-name TEXT             An optional run name that will be used as a
                                  suffix to form the names of new `obs`
                                  annotation columns that will store the
                                  predicted annotation values and confidence
                                  scores. This can be used to allow multiple
                                  annotation predictions to be run on a single
                                  AnnData object.
  --use-gpu / --no-use-gpu        Whether to use a GPU for annotation
                                  operations (highly recommended, if
                                  available).
  --use-model-cache / --no-use-model-cache
```

### DIFF
--- a/server/cli/annotate.py
+++ b/server/cli/annotate.py
@@ -49,7 +49,6 @@ def annotate_args(func):
 )
 @click.option(
     "-o",
-    # "--output",
     "--output-h5ad-file",
     default="",
     help="The output H5AD file that will contain the generated annotation values.",
@@ -60,7 +59,7 @@ def annotate_args(func):
     default=False,
     is_flag=True,
     help="Overwrite the H5AD input file, adding the predicted annotation values. For safety, you must specify this "
-    "flag if the --output option is not specified.",
+    "flag if the --output-h5ad-file option is not specified.",
     show_default=True,
 )
 @click.option(
--- a/server/cli/annotate.py
+++ b/server/cli/annotate.py
@@ -4,7 +4,7 @@ import os.path
 import shlex
 import shutil
 import subprocess
-import sys
+from os.path import isfile
 from subprocess import STDOUT, PIPE
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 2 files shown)


seve (approved):

server/cli/annotate.py
```diff
@@ -27,23 +27,42 @@ def wrapper(*args, **kwargs):
...
+    nargs=1,
+    metavar="<path to H5AD input file>",
     required=True,
-    type=str,
-    help="The input H5AD file containing the missing annotations.",
 )
 @click.option(
     "-m",
     "--model-url",
+    # Making this a required "option", rather than an "argument", since we support automatic model selection in the
```
MaximilianLombardo, line 42: good call!

server/cli/annotate.py
```diff
@@ -53,8 +72,8 @@ def wrapper(*args, **kwargs):
 @click.option(
     "-g",
     "--gene-column-name",
-    help="The name of the `var` column that contains gene identifiers. The values in this column will be used to match "
-    "genes between the query and reference datasets. If not specified, the gene identifiers are expected to exist "
+    help="The name of the `var` column that contains gene names. The values in this column will be used to match "
```
MaximilianLombardo, line 75: maybe say ''gene symbols'' instead of ''gene names'' here?

MaximilianLombardo (approved):
Looks good!

server/cli/annotate.py
atolopko-czi: settled on keeping `gene names`, since that's in fact what we get from `feature_name` in our schema-compliant h5ad files that are used to train our models (at least for the lung model).

PR #2574, opened 2022-09-21
fix: fix static file serving in webpack

seve:


atolopko-czi (approved):
Looks like a reasonable fix! 🚀

PR #2622, opened 2023-07-19
chore(deps): change python-igraph to igraph

szhorvat:
Please see https://github.com/igraph/python-igraph/issues/699 for details


szhorvat:
I do not have the permission to re-open this.

PR #2685, opened 2024-08-23
chore: Fix compatibility tests

tihuan:
## Changes
1. Dropped support for python 3.8 and 3.9
2. Added support for python 3.10 and 3.11
3. Updated python packages
4. Updated code and fixed unit tests 
5. Pinned Node ~16
6. Updated deps for `pupetteer` and `jest-pupetteer` 

-----
As for Python 3.12 - it requires updating flatbuffers in order to work which is a breaking change and doesn't work with the current dep tree - it will also require migration


kaloster:
all green now - verified to install and run locally

Screenshot 2024-09-03 at 12 13 53 PM

.github/workflows/compatibility_tests.yml
```diff
@@ -6,6 +6,11 @@ on:
   push:
     branches:
       - main
+  # DEBUG
+  # DEBUG
+  # DEBUG
+  pull_request:
+    branches: "*"
```
tihuan, line 13: ah we still need to remove this thank you!

client/.nvmrc
```diff
@@ -0,0 +1 @@
+16.20.0
```
tihuan, line 1: sweet!

tihuan (commented):
Just one last change thank you!!

tihuan (commented):
LGTM! Thanks so much for fixing this, @kaloster 🙏 ❤️ Just one last question regarding whether we should test "3.12" as well thanks again!

tihuan (commented):
Just one comment left. Otherwise LGTM!! Thanks for delivering this miracle ✨ 🚀 ✨

tihuan:
V COOL!! LGTM 🎉🎉🎉 Thank you thank you!!

tihuan:
@kaloster I think you need to LGTM this PR to merge lol Sorry ahahah

kaloster (approved):

PR #2691, opened 2024-09-09
fix: FE deps critical

kaloster:
## Changes
- upgraded @babe/core and deps to address security alerts


tihuan (approved):
Amazing thank you so much!!

PR #2743, opened 2026-01-13
fix: improve anndata error msg

rwbaber:
#### Reviewers
**Functional:** 

**Readability:** 

---

## Changes

* **Modify** `server/data_anndata/anndata_adaptor.py` to improve error handling during `.h5ad` file loading.
* **Add** a specific check for `anndata` forward-compatibility errors (e.g., `IORegistryError`) within the `_load_data` exception block.
* **Modify** the generic "File not found or is inaccessible" error message to provide actionable feedback when a version mismatch is detected.

---

## Context & Motivation

Currently, when `cellxgene` attempts to load an `.h5ad` file created with a newer version of `anndata` than the one installed in the server environment, it triggers an `anndata._io.specs.registry.IORegistryError`. This frequently occurs with files created in `anndata` **v0.12.0+**, which utilizes a new `null` encoding for `None` types (which is found in `uns['log1p']` resulting from scanpy's `sc.pp.log1p()` function).

In these cases, the `AnndataAdaptor` catches the generic exception and reports that the file is "not found or is inaccessible." This is misleading, as the file exists and permissions are correct; the issue is an environment mismatch.

This change catches errors containing `"No read method registered"` and `"IOSpec"`, providing the user with the actual internal error and a suggestion to upgrade their environment.

## Testing

1. **Created a "Broken" Environment:** Python 3.10 with `anndata==0.11.4`.
2. **Generated a Test File:** Used `anndata==0.12.2` to save a file containing `adata.uns['log1p'] = {'base': None}` from running `sc.pp.log1p(adata)`.
3. **Verified the Fix:**
* **Before:** `cellxgene launch` reported "File not found or is inaccessible."
* **After:** `cellxgene launch` correctly identified the version mismatch and suggested `pip install --upgrade anndata`.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+# The MIT License (MIT)
 
-Copyright (c) 2017-2023 Chan Zuckerberg Initiative
+Copyright (c) 2017-2026 Chan Zuckerberg Initiative
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,3 @@
+# Reporting Security Issues
+
+If you believe you have found a security issue, please responsibly disclose by contacting us at [security@chanzuckerberg.com](mailto:security@chanzuckerberg.com).


tihuan:
Merged the PR! Thanks so much again, @rwbaber 🎉 ❤️

rwbaber:
Cool, thanks @tihuan!

Before seeing that you already merged this PR, I noticed the unit tests were failing due to a scipy incompatibility with float16 files (strict validation). Since this causes the general adaptor tests to crash during setup, and therefore the unit-test PR check to fail every time, I submitted a small follow-up PR to fix it: #2746 

It seems redundant given that test_anndata_adaptor_errors.py already covers this case, but if I missed a specific reason for keeping it in the success-path suite, feel free to close it!

PR #2767, opened 2026-03-09
fix: Fix opacity of centroid labels

LiudengZhang:
## Summary

- Fix typo in prop name: `opactity` → `opacity` when passing the opacity value to the `Label` component (line 153)
- Fix extraneous brackets in style object: `opacity: { opacity }` → `opacity` shorthand property (line 208). The original code wraps `opacity` in object literal braces `{}` inside the style object, producing an object value instead of the numeric opacity value.

## Test plan

- [ ] Verify centroid labels render with correct opacity when categories are selected/deselected
- [ ] Verify deselected centroid labels appear semi-transparent rather than invisible


tihuan:
Great catch! Thanks so much for the fix 😄 🙏

tihuan (approved):

PR #2794, opened 2026-07-29
Fix gene coloring 400 with pandas 3.0 string dtype var index

### LINKED ISSUE #2793: [BUG] Gene coloring returns HTTP 400 with pandas 3.0 string dtype var index
## Summary

Coloring by (fetching) a gene fails with **HTTP 400** for AnnData files whose `var` index (gene names) uses a **pandas string dtype** rather than `object`. This is the default for text columns under **pandas 3.0**, and under pandas 2.x with `pd.set_option("future.infer_string", True)`. The dataset loads fine; only gene expression fetch breaks:

```
GET /api/v0.2/data/var?var:name_0=<gene>  ->  400
```

## Minimal reproduction

```python
import anndata, numpy as np, pandas as pd

# pandas 3.0 makes this the default; on pandas 2.x opt in explicitly:
pd.set_option("future.infer_string", True)

adata = anndata.AnnData(
    X=np.random.default_rng(0).random((10, 2000), dtype=np.float32),
    var=pd.DataFrame(index=[f"gene{i}" for i in range(2000)]),
)
print(adata.var.index.dtype)   # 'str' (pandas 3.0) instead of object
adata.write_h5ad("repro.h5ad")
```

```
cellxgene launch repro.h5ad
```

In the UI, color by any gene (e.g. `gene0`). The request
`GET /api/v0.2/data/var?var:name_0=gene0` returns **400**.

(The dataset has 2000 genes only so the failure is deterministic — see below.)

## Root cause

`DataAdaptor._annotation_filter_to_mask` decides between an exact-value filter and a numeric min/max filter based on the column dtype name:

https://github.com/chanzuckerberg/cellxgene/blob/76a39fcf92da57c9e4fd59831ad805a3b007da8c/server/data_common/data_adaptor.py#L183

```python
if anno_data.dtype.name in ["boolean", "category", "object"]:
    values = v.get("values", [])
    key_idx = np.in1d(anno_data, values)      # exact-match value filter
    mask = np.logical_and(mask, key_idx)
else:
    min_ = v.get("min", None)                  # numeric range filter
    max_ = v.get("max", None)
    ...
```

A pandas string dtype (`"string"`, `"string[pyarrow]"`, or the pandas 3.0 NumPy-backed `"str"`) is **not** in the allow-list, so a gene-name value filter falls into the numeric branch. With no `min`/`max` present, that branch is a no-op and the mask stays **all-`True`** — i.e. *every* var column is selected.

`data_frame_to_fbs_matrix` then rejects the request because the selected column count exceeds `column_request_max` (default 32):

```python
if self.server_config.exceeds_limit("column_request_max", num_columns):
    raise ExceedsLimitError("Requested dataframe columns exceed column request limit")
```

which the REST layer maps to **400**. (Any dataset with more than `column_request_max` genes reproduces it; the repro uses 2000 to make it unconditional.)

The same bug affects obs categorical value filters stored as a string dtype.

## Environment

- Triggered by text columns using a pandas string dtype instead of `object` — the default in **pandas 3.0**, or pandas 2.x with `future.infer_string`.
- Confirmed against `server/data_common/data_adaptor.py` on `main` (76a39fc).

## Suggested fix

Route all pandas string dtypes through the value-equality branch, e.g.:

```python
if anno_data.dtype.name in ["boolean", "category", "object"] or pd.api.types.is_string_dtype(anno_data):
```

I'll open a PR with this fix and regression tests.

colganwi:
## Problem

Fixes #2793.

Coloring by a gene returns **HTTP 400** for AnnData files whose `var` index (gene names) uses a pandas **string dtype** instead of `object` — the default for text columns under **pandas 3.0**, and under pandas 2.x with `future.infer_string`:

```
GET /api/v0.2/data/var?var:name_0=<gene>  ->  400
```

## Root cause

`DataAdaptor._annotation_filter_to_mask` chooses between an exact-value filter and a numeric min/max filter by dtype name:

```python
if anno_data.dtype.name in ["boolean", "category", "object"]:
    key_idx = np.in1d(anno_data, values)   # value filter
    ...
else:
    # numeric min/max — a no-op when only "values" is provided
```

A pandas string dtype (`"string"`, `"string[pyarrow]"`, or the pandas 3.0 NumPy-backed `"str"`) is not in the allow-list, so a gene-name value filter falls into the numeric branch and does nothing. The mask stays all-`True`, so *every* var column is selected; `data_frame_to_fbs_matrix` then exceeds `column_request_max` and raises `ExceedsLimitError` → 400. The same applies to obs categorical value filters stored as a string dtype.

## Fix

Route all pandas string dtypes through the value-equality branch via `pandas.api.types.is_string_dtype`:

```python
if anno_data.dtype.name in ["boolean", "category", "object"] or pd.api.types.is_string_dtype(anno_data):
```

## Tests

Adds `StringDtypeFilterMaskTest` covering both `StringDtype` (`"string"`) and the pandas 3.0 `"str"` dtype (via `future.infer_string`), asserting the value filter selects exactly the matching row.

## Reproduction

```python
import anndata, numpy as np, pandas as pd
pd.set_option("future.infer_string", True)  # pandas 3.0 default
adata = anndata.AnnData(
    X=np.random.default_rng(0).random((10, 2000), dtype=np.float32),
    var=pd.DataFrame(index=[f"gene{i}" for i in range(2000)]),
)
adata.write_h5ad("repro.h5ad")
# cellxgene launch repro.h5ad -> color by a gene -> 400 (before this fix)
```

### DIFF
--- a/server/data_common/data_adaptor.py
+++ b/server/data_common/data_adaptor.py
@@ -180,7 +180,13 @@ class DataAdaptor(metaclass=ABCMeta):
             elif axis == Axis.OBS:
                 anno_data = self.query_obs_array(name)
 
-            if anno_data.dtype.name in ["boolean", "category", "object"]:
+            # Discrete/label columns are filtered by exact value. This must
+            # include the pandas string dtypes that modern anndata writes for
+            # gene names and obs labels: StringDtype ("string"/"string[pyarrow]")
+            # and, under pandas 3.0 (or pandas 2.x with future.infer_string), the
+            # NumPy-backed "str" dtype that replaces object for text. Otherwise
+            # the value filter is silently skipped and the whole axis is selected.
+            if anno_data.dtype.name in ["boolean", "category", "object"] or pd.api.types.is_string_dtype(anno_data):
                 values = v.get("values", [])
                 key_idx = np.in1d(anno_data, values)
                 mask = np.logical_and(mask, key_idx)
--- a/test/unit/data_anndata/test_anndata_adaptor.py
+++ b/test/unit/data_anndata/test_anndata_adaptor.py
@@ -8,6 +8,7 @@ import pytest
 from parameterized import parameterized_class
 
 import test.decode_fbs as decode_fbs
+from server.common.constants import Axis
 from server.common.utils.data_locator import DataLocator
 from server.common.errors import FilterError
 from server.data_anndata.anndata_adaptor import AnndataAdaptor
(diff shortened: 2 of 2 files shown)


colganwi:
Commenting to keep this PR open since the issue has not been fixed

